PR #14, opened 2019-08-20
Create releases on git tags

mdelapenya:
We prefer generating the build artefacts on each git tag, not on each merge to master

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -86,7 +86,7 @@ pipeline {
         }
         stage('Release') {
           options { skipDefaultCheckout() }
-          when { branch 'master' }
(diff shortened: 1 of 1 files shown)


.ci/Jenkinsfile
```diff
@@ -86,7 +86,7 @@ pipeline {
         }
         stage('Release') {
           options { skipDefaultCheckout() }
-          when { branch 'master' }
+          when { tag "v*" }
```
v1v: if we wanna ensure a particular semver then we might need to be more explicit,

mdelapenya: Yes, let's merge and improve the regex later

Thanks!

PR #18, opened 2019-08-22
fix: pre-commit requirements

v1v:
## highlight
- Let's install go as it's required to be in place based on the enabled hooks.
- Let's install all the required dependencies to trigger the pre-commit.

## Tasks
- [ ] Remove test changes


mdelapenya:
Are we going to remove the testing-purpose commits?

cli/config/config.go
```diff
@@ -316,6 +316,7 @@ func checkConfigFile(workspace string) {
 		}).Fatal("Cannot create workdir for 'op'")
 	}
 
+	log.Info("Creating %s with default values in %s.", fileName, workspace)
```
mdelapenya: Isn't this line the same as in L327?

v1v: This is only for testing purposes, but when I did the merge for some reason I did a wrong merge

PR #20, opened 2019-08-26
First functional tests against a running stack

mdelapenya:
## What is this PR doing?
This PR improves the existing functional tests, which verified that an output file was created, replacing them with a more elaborated ones: we will run an elastic stack (elasticsearch + kibana) so that we are able to perform assertions against the Elasticsearch APIs.

We run `metricbeat setup` in an async manner in the running metricbeat container, so that the kibana index patterns are properly created.

To build the ES queries, we are going to use the official [Elasticsearch Go Client](https://github.com/elastic/go-elasticsearch), so that we are able to easily query the APIs in a standard manner.

The query that we are going to build will use `service.version` and `event.module` fields to discover the documents relative to each module (apache, redis, mysql). Within the query result, we will verify that there aren't error documents inside the Elasticsearch's response.

## Why is this PR important?
The functional tests that we are proposing here will set up the foundations when writing functional tests for Metricbeat's integrations, with a more correct approach than the existing one, based in output files.

## Considerations
We have to take care of metricbeat's index names, because if we just use the one generated by metricbeat, then we will have to build the index name in code, which could not be totally accurate. IMHO we should define index name at metricbeat configuration file, so that it's a constant for the tests.


cli/cmd/run.go
```diff
@@ -87,7 +90,20 @@ func buildRunStackCommand(key string, stack config.Stack) *cobra.Command {
...
+
+				if k == "kibana" {
+					s.SetWaitFor(wait.ForLog("http server running"))
+				}
+
+				if k == "elasticsearch" {
+					serviceManager.Run(s)
+				} else {
+					servicesToRun[k] = s
+				}
```
jsoriano: Why adding special cases for kibana and elasticsearch?

cli/cmd/run.go
```diff
@@ -87,7 +90,20 @@ func buildRunStackCommand(key string, stack config.Stack) *cobra.Command {
 					"stack": stack.Name,
 				}
 				s := serviceManager.BuildFromConfig(srv)
-				serviceManager.Run(s)
+
+				if k == "kibana" {
+					s.SetWaitFor(wait.ForLog("http server running"))
```
jsoriano: In metricbeat tests the healthcheck checks that kibana is green in `/api/status`.

cli/docker/docker.go
```diff
@@ -26,6 +26,34 @@ func ConnectContainerToDevNetwork(containerID string, aliases ...string) error {
 		})
 }
 
+// ExecCommandIntoContainer executes a command, as a user, into a container
+func ExecCommandIntoContainer(containerName string, user string, cmd []string) error {
+	dockerClient := getDockerClient()
+
+	response, err := dockerClient.ContainerExecCreate(
+		context.Background(), containerName, types.ExecConfig{
```
jsoriano: Add the context as parameter of this function, or use `context.TODO()`.

cli/docker/docker.go
```diff
@@ -26,6 +26,34 @@ func ConnectContainerToDevNetwork(containerID string, aliases ...string) error {
...
+			Detach:       true,
+			Cmd:          cmd,
+		})
+
+	if err != nil {
+		return err
+	}
+
+	err = dockerClient.ContainerExecStart(
+		context.Background(), response.ID, types.ExecStartCheck{
```
jsoriano: I'd reuse the same context as the one used in the previous call.

cli/docker/docker.go
```diff
@@ -26,6 +26,34 @@ func ConnectContainerToDevNetwork(containerID string, aliases ...string) error {
...
+			Detach: true,
```
jsoriano: What does `Detach` mean here? Is the command executed without waiting for finalization?

metricbeat-tests/Makefile
```diff
@@ -11,5 +16,13 @@ install-godog:
 	go get github.com/DATA-DOG/godog/cmd/godog
```
jsoriano: Not related to this PR, but commands installed during CI should be vendorized, or fixed somehow to a version, to avoid failures by unexpected changes.

metricbeat-tests/configurations/apache.yml
```diff
@@ -16,6 +16,12 @@ metricbeat.modules:
...
\ No newline at end of file
+output.elasticsearch:
+  hosts: ["http://elasticsearch:9200"]
+  username: "elastic"
+  password: "p4ssw0rd"
+
+setup.kibana:
+  host: "http://kibana:5601"
+  username: "elastic"
+  password: "p4ssw0rd"
```
jsoriano: I think it'd be better to have this configuration as part of the metricbeat execution and not as part of the test files. These settings can be also set as execution flags, as `-E output.elasticsearch.hosts=http://elasticsearch:9200 -E output.elasticsearch.username=elastic ...`
This way we could also consider different scenarios with the same configuration files.

Kibana options shouldn't be needed for ingestion. We might avoid any test with kibana by now, I think it would only make sense if we test dashboards.

metricbeat-tests/elasticsearch.go
```diff
@@ -0,0 +1,122 @@
...
+// at configuration level. Then we will inspect the running container to get its port bindings
+// and from them, get the one related to the Elasticsearch port (9200). As it is bound to a
+// random port at localhost, we will build the URL with the bound port at localhost.
+func getElasticsearchClient(stackName string) *es.Client {
+	elasticsearchCfg, _ := config.GetServiceConfig("elasticsearch")
+	elasticsearchCfg.Name = elasticsearchCfg.Name + "-" + stackName
+
+	elasticsearchSrv := serviceManager.BuildFromConfig(elasticsearchCfg)
+
+	esJSON, _ := docker.InspectContainer(elasticsearchSrv.GetContainerName())
```
jsoriano: Handle error.

metricbeat-tests/elasticsearch.go
```diff
@@ -0,0 +1,122 @@
...
+
+	ports := esJSON.NetworkSettings.Ports
+	binding := ports[nat.Port("9200/tcp")]
+
+	esClient, _ := es.NewClient(
```
jsoriano: Handle error.

metricbeat-tests/elasticsearch.go
```diff
@@ -0,0 +1,122 @@
...
+		return result, err
+	}
+
+	result.Result = r
+
+	log.WithFields(log.Fields{
+		"status": res.Status(),
+		"hits":   int(r["hits"].(map[string]interface{})["total"].(map[string]interface{})["value"].(float64)),
+		"took":   int(r["took"].(float64)),
+	}).Info("Response information")
```
jsoriano: Debug?

metricbeat-tests/elasticsearch.go
```diff
@@ -0,0 +1,122 @@
...
+
+		return result, err
+	}
+
+	log.WithFields(log.Fields{
+		"query": fmt.Sprintf("%s", query),
+	}).Debug("Elasticsearch query")
+
+	res, err := esClient.Search(
+		esClient.Search.WithContext(context.Background()),
```
jsoriano: No need to add a context here if it is going to be an empty context.
```suggestion
```

metricbeat-tests/metricbeat.go
```diff
@@ -53,7 +46,19 @@ func RunMetricbeatService(version string, monitoredService services.Service) (se
...
 		"service":           serviceName,
+		"serviceVersion":    monitoredService.GetVersion(),
 	}).Info("Metricbeat is running configured for the service")
 
+	log.WithFields(log.Fields{
+		"metricbeatVersion": version,
+		"service":           serviceName,
+	}).Debug("Installing Kibana dashboards")
+	docker.ExecCommandIntoContainer(
+		service.GetContainerName(), "root", []string{"metricbeat", "setup"})
```
jsoriano: To avoid requiring Kibana for these tests, setup only indexes, that I think is all that is needed by now.

```suggestion
		service.GetContainerName(), "root", []string{"metricbeat", "setup", "--index-management"})
```

metricbeat-tests/runner_test.go
```diff
@@ -54,78 +58,85 @@ func TestMain(m *testing.M) {
...
-
-	files, err := filepath.Glob(pattern)
-	if err != nil {
-		log.WithFields(log.Fields{
-			"error": err,
-			"path":  pattern,
-		}).Fatal("Cannot retrieve outputs :(")
+// assertHitsArePresent returns an error if no hits are present
+func assertHitsArePresent(hits map[string]interface{}, q ElasticsearchQuery) error {
+	hitsCount := len(hits["hits"].(map[string]interface{})["hits"].([]interface{}))
```
jsoriano: Each one of these type conversions can panic if they don't exist or are not of the expected type, you may want to convert item by item:
```suggestion
      var hits []interface{}
      if hitsObject, ok := hits["hits"].(map[string]interface{}); ok {
          hits, _ = hitsObject["hits"].([]interface{})
      }
      hitsCount = len(hints)
```

Or use something like [mapstructure](https://godoc.org/github.com/mitchellh/mapstructure):
```suggestion
      var searchResponse struct {
          Hints struct {
              Hints []interface{}
          }
      }
      mapstructure.Decode(hits, &searchResponse)
      hitsCount := len(searchResponse.Hints.Hints)
```

metricbeat-tests/runner_test.go
```diff
@@ -54,78 +58,85 @@ func TestMain(m *testing.M) {
...
-	for {
-		exists := fileExists(filePath)
-		if exists {
-			foundChannel <- true
-			close(foundChannel)
-			break
+// assertHitsDoNotContainErrors returns an error if any of the returned entries contains
+// an "error.message" field in the "_source" document
+func assertHitsDoNotContainErrors(hits map[string]interface{}, q ElasticsearchQuery) error {
+	for _, hit := range hits["hits"].(map[string]interface{})["hits"].([]interface{}) {
```
jsoriano: Same comment here about unchecked type conversions.

metricbeat-tests/runner_test.go
```diff
@@ -54,78 +58,85 @@ func TestMain(m *testing.M) {
...
-		log.WithFields(log.Fields{
-			"file":    fileName,
-			"times":   loops,
-			"seconds": (loops * seconds),
-		}).Debug("Waiting for the file to be present")
+func thereAreNoErrorsInTheIndex(metricbeatVersion string) error {
+	esIndexName := strings.ReplaceAll(metricbeatVersion, "-SNAPSHOT", "")
+	now := time.Now()
+
+	formattedDate := strings.ReplaceAll(now.Format("2006-01-02"), "-", ".")
```
jsoriano: This can introduce flakiness on midnight, what do you think about using wildcards and querying `metricbeat-*` instead of building a full index name?

metricbeat-tests/runner_test.go
```diff
@@ -54,78 +58,85 @@ func TestMain(m *testing.M) {
...
+
+	// TODO: this index name is hardcoded after checking the index name on Kibana
+	// I would need help setting up the index name from metricbeat configuration
```
jsoriano: I think you could use wildcards. If you want to use custom indexes to isolate test executions or something you can modify the index preffix with [`setup.ilm.rollover_alias`](https://www.elastic.co/guide/en/beats/metricbeat/7.3/ilm.html#setup-ilm-rollover_alias-option).

.ci/functionalTests.groovy
```diff
@@ -98,6 +100,7 @@ pipeline {
               junit(allowEmptyResults: true, keepLongStdio: true, testResults: "${BASE_DIR}/outputs/junit-*.xml")
               archiveArtifacts allowEmptyArchive: true, artifacts: "${BASE_DIR}/outputs/junit-*"
               githubCheckNotify(currentBuild.currentResult == 'SUCCESS' ? 'SUCCESS' : 'FAILURE')
+              sh script: 'make -C metricbeat-tests shutdown-elastic-stack', label: 'Shutdown runtime dependencies'
```
v1v: I just realized if the make uses the `go` command, then it might be required to create a wrapper script to prepare the context for the given `${GO_VERSION}`. and replace the call for something like:
`.ci/scripts/shutdown-stack.sh "${GO_VERSION}"`

metricbeat-tests/configurations/redis.yml
```diff
@@ -5,8 +5,14 @@ metricbeat.modules:
...
   # Redis hosts
-  hosts: ["http://${HOST}:6379"]
+  hosts: ["http://${MONITORED_HOST}:6379"]
 
-output.file:
-  path: "/tmp"
-  filename: "${FILE_NAME}.metrics"
+output.elasticsearch:
+  hosts: ["http://elasticsearch:9200"]
+  username: "elastic"
```
v1v: Will it be possible to setup those values as variables somehow? Just in case in the future, we would like to use nonephemeral instances but we need to secure them with `vault` or something like that. I don't see this is a blocker but something to think about if it could be required in the future.

cli/cmd/run.go
mdelapenya: Testcontainers comes with a wait.ForHTTP("/api/status") that we could leverage instead of using the wait.ForLog("foo").

cli/cmd/run.go
mdelapenya: To implement the startup order, and following the BigPicture diagram (https://docs.google.com/drawings/d/1MJ-haufU06JnUHYxYDwP-EH-xBpyMDrpDn2UEdS2WgA/edit) we must do those checks.

If you look at the picture, and create a tree-graph, you'll get that elasticsearch is the root node, so the first one to be started. Next in the line would be the APM server...

cli/docker/docker.go
mdelapenya: You mean `ExecCommandIntoContainer` receiving a context, right?

cli/docker/docker.go
mdelapenya: It tells Docker client to run the container with `--detach` flag (as in _docker run -d ..._), so it will run the container and will continue its execution

metricbeat-tests/Makefile
mdelapenya: Agreed

metricbeat-tests/configurations/apache.yml
mdelapenya: Yeah, first idea was to configure the entire metricbeat with env vars, which would be more flexible. If possible, I'd go with that direcction

metricbeat-tests/runner_test.go
mdelapenya: The idea is to replace this `map[string]interface{}` with a fluent approach to create queries. I used this as in the Elasticsearch Go SDK

metricbeat-tests/runner_test.go
mdelapenya: Yes, I mentioned in the PR, I don't like this hardcoded string either :)

metricbeat-tests/configurations/redis.yml
mdelapenya: Yeah, @jsoriano proposed using env vars to completely replace configuration files, which was the initial plan

cli/cmd/run.go
mdelapenya: Applied

cli/docker/docker.go
mdelapenya: Applied

cli/docker/docker.go
mdelapenya: Applied

metricbeat-tests/elasticsearch.go
mdelapenya: Applied

metricbeat-tests/metricbeat.go
mdelapenya: Applied

metricbeat-tests/elasticsearch.go
mdelapenya: Applied

.ci/functionalTests.groovy
jsoriano: [gvm](https://github.com/andrewkroh/gvm/releases/tag/v0.2.1) can help to get specific Go versions and configure the environment variables for it.

cli/cmd/run.go
jsoriano: It'd be good to have some kind of dependencies mechanism to avoid these special cases (like `depends_on` in docker compose).

cli/cmd/run.go
jsoriano: Ok, take into account that ideally we should wait till `/api/status` says that the state is green. Is it possible to extend `wait.ForHTTP()` to check the content of the response?

cli/docker/docker.go
jsoriano: Ok, so we are using this exec to start the container.

metricbeat-tests/runner_test.go
jsoriano: Using wildcards is an option to remove the ugly parts :slightly_smiling_face:

.ci/functionalTests.groovy
mdelapenya: No, it does not use Go directly: it calls the go binary instead

cli/cmd/run.go
mdelapenya: Yeah, we do not have it in place now. It'd be something to work on, or wait for the potential move to compose...

cli/docker/docker.go
mdelapenya: Wait, my previous comment was misleading, sorry.

The `ContainerExecStart` method is to launch a command inside a running container, so the `Detach` attribute is to run the command in the background (Please read this: https://docs.docker.com/engine/reference/commandline/exec/#options)

jsoriano: Oh ok, but then don't you want to wait or the finalization of the command to see if it worked correctly?

metricbeat-tests/configurations/redis.yml
mdelapenya: We finally moved to the following approach: specific module setups will go in the YAML file, but the shared ones (elasticsearch output and kibana setup) will be defined as flags for the metricbeat instance

metricbeat-tests/runner_test.go
mdelapenya: Applied

metricbeat-tests/runner_test.go
mdelapenya: Applied

metricbeat-tests/runner_test.go
mdelapenya: Will work on a DSL to simplify its usage

cli/docker/docker.go
mdelapenya: We have removed it, as it's not needed anymore

PR #26, opened 2019-09-02
(#22) Create a simple DSL for querying elasticsearch

mdelapenya:
It simplifies the creation of the queries used in this POC.

Spoiler: it just covers this use case: a query > bool > must > match query !

### DIFF
--- a/metricbeat-tests/runner_test.go
+++ b/metricbeat-tests/runner_test.go
@@ -171,28 +171,24 @@ func thereAreNoErrorsInTheIndex(index string) error {
 	// no collitions between different test cases should appear
 	esIndexName += "-*"
 
-	esQuery := map[string]interface{}{
-		"query": map[string]interface{}{
-			"bool": map[string]interface{}{
-				"must": []map[string]interface{}{
-					{
-						"match": map[string]interface{}{
-							"event.module": query.EventModule,
-						},
-					},
-					{
-						"match": map[string]interface{}{
-							"service.version": query.ServiceVersion,
-						},
-					},
-				},
-			},
-		},
-	}
+	must := mustQuery{}.WithMatches(
+		matchQuery{}.WithMatchEntry(
+			matchEntry{
+				"event.module": query.EventModule,
+			}),
+		matchQuery{}.WithMatchEntry(
+			matchEntry{
+				"service.version": query.ServiceVersion,
+			}),
+	)
+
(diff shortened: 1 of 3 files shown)


metricbeat-tests/runner_test.go
```diff
@@ -171,28 +171,24 @@ func thereAreNoErrorsInTheIndex(index string) error {
...
-					{
-						"match": map[string]interface{}{
-							"service.version": query.ServiceVersion,
-						},
-					},
-				},
-			},
-		},
-	}
+	must := mustQuery{}.WithMatches(
```
kuisathaverat: for me is more complex because it does not match with the Elasticsearch query language, it means I will learn a new query language.

mdelapenya: Yeah, it was my purpose to evaluate that experience

Let's check out other's feedback, but it's very likely that I will close this one after your suggestion

Thanks!

mdelapenya: Thinking out loud about the learning process: a fluent DSL in a compiled language has the support of the compiler, so having a regular IDE will autocomplete methods, so you just follow the available methods.

We could improve the DSL so make it simpler... Let me send a new commit 🤔

jsoriano: One of the main purpouses for a DSL is to provide a language more suited for the specific use case. Even if it requires to learn a new language, learning it can worth it if offers a good abstraction.

In this project we probably don't need a direct match with the whole ES query language, as here we probably are only going to make a very limited set of queries, with a limited set of conditions, probably only querying for documents for an specific dataset, version and/or where certain fields are present or not.
We are probably never going to use `should` for example, so does it worth to expose the `must`?

We could do it as a higher-level API with an IDE-friendly pattern, so for example the could here could be like:
```
s := newSearch()
s.Module(query.EventModule)
s.ServiceVersion(query.ServiceVersion)
result, err := s.Do(stackName, esIndexName)
```
`newSearch()` would return a map for a query with an empty `must` clause, `Module` and `ServiceVersion` would add (or replace) `match` conditions for `event.module` and `service.version`. Other methods could be added that would add other conditions to the query.
 
In summary, if we add a DSL, I'd make it simple, and limited to our use cases :slightly_smiling_face:

mdelapenya:
Let's close this one, and get back to it if necessary

Thanks!

PR #34, opened 2019-09-04
fix pre-commit hooks in the CI

v1v:
## Highlights
- gometalinter is deprecated
- go-build/go-unit-test should be done within the pipeline itself rather than within the pre-commit.
- go-vet causes some issues when loading some packages. Let's remove it from now


.ci/Jenkinsfile
```diff
@@ -61,6 +60,7 @@ pipeline {
                   unstash 'source'
                   dir(BASE_DIR){
                     sh script: """.ci/scripts/install-dependencies.sh "${GO_VERSION}" """, label: 'Install dependencies'
+                    sh 'ls -ltrah && pwd && env | sort'
```
mdelapenya: Do we want `env` to be executed on CI?

cli/config/config.go
```diff
@@ -10,6 +10,10 @@ import (
 	"reflect"
 	"strings"
 
+
```
mdelapenya: I guess this is for CI to complain 😄

.ci/Jenkinsfile
v1v: debugging yet as it works locally but not in the CI

```
14:39:51  hookid: golangci-lint
14:39:51  
14:39:51  level=error msg="Running error: context loading failed: no go files to analyze"
```

cli/config/config.go
v1v: exactly

.ci/Jenkinsfile
v1v:
```suggestion
```

PR #38, opened 2019-09-06
(#37) Honor Godog's life cycle hooks (Before/After Scenario)

### LINKED ISSUE #37: Honor Godog life cycle hooks
The `BeforeScenario` and `AfterScenario` hooks are not properly defined.

After reading https://semaphoreci.com/community/tutorials/how-to-use-godog-for-behavior-driven-development-in-go, I realised that each featureContext function must define its own hook

mdelapenya:
## What does this PR do?
This PR moves the Before/After Scenario hooks to each test file (Redis, Apache, MySQL) so that each own controls the life cycle of the containers they start.

We are adding the ability to each Service to define a CleanUp function, that will be triggered after a successful termination of the container.

On the other hand, as a refactoring task, we are moving assertions to its own Go file, so that they can become an assertions library to be reused across tests in a simpler manner, cleaning up the test runner.

## Why is it important?
We want each test to be idempotent, which means starting the containers they need Before each scenario, and destroying them After each scenario. That why we can ensure we do not put dirty data into the storage, so that each test creates and destroys the data it produces.

## Related Issues
Closes #37


metricbeat-tests/mysql_test.go
```diff
@@ -2,21 +2,33 @@ package main
...
 	s.Step(`^there are no errors in the "([^"]*)" index$`, thereAreNoErrorsInTheIndex)
+
+	s.BeforeScenario(func(interface{}) {
+		log.Debug("Before scenario...")
+		mysqlTestSuite = MetricbeatTestSuite{}
+	})
+	s.AfterScenario(func(interface{}, error) {
+		log.Debug("After scenario...")
+		mysqlTestSuite.CleanUp()
+	})
```
jsoriano: Same code is being added to all feature contexts, I think this should be part of some helper.

metricbeat-tests/mysql_test.go
```diff
@@ -2,21 +2,33 @@ package main
...
 
 func metricbeatIsInstalledAndConfiguredForMySQLModule(metricbeatVersion string) error {
-	s, err := RunMetricbeatService(metricbeatVersion, mysqlService)
+	mysqlService := mysqlTestSuite.Service
 
-	metricbeatService = s
+	s, err := RunMetricbeatService(metricbeatVersion, mysqlService)
+	if err == nil {
+		mysqlTestSuite.Metricbeat = s
+	}
```
jsoriano: Same here, this is added for all services, this should probably be part of some helper.

metricbeat-tests/mysql_test.go
```diff
@@ -27,9 +39,14 @@ func metricbeatIsInstalledAndConfiguredForMySQLModule(metricbeatVersion string)
...
-	mysqlService = serviceManager.Build("mysql", mysqlVersion, false)
+	mysqlService := serviceManager.Build("mysql", mysqlVersion, true)
 
 	mysqlService.SetNetworkAlias("mysql_" + mysqlVersion + "-metricbeat_" + metricbeatVersion)
 
-	return serviceManager.Run(mysqlService)
+	err := serviceManager.Run(mysqlService)
+	if err == nil {
+		mysqlTestSuite.Service = mysqlService
+	}
```
jsoriano: And also here, same code added for all services.

metricbeat-tests/metricbeat.go
```diff
@@ -47,10 +50,16 @@ func RunMetricbeatService(version string, monitoredService services.Service) (se
...
 	service.SetEnv(env)
+	service.SetAsDaemon(true)
 	service.SetLabels(labels)
 
-	container, err := service.Run()
-	if err != nil || container == nil {
+	fn := func(ctx context.Context) error {
+		return deleteIndex(ctx, "metricbeat", indexName)
+	}
+	service.SetCleanUp(fn)
```
jsoriano: I think that it would be even more important to delete the indexes before running the scenario to ensure that they run in a clean environment.

mdelapenya:
Hey @jsoriano I addresses your suggestions in a refactor of the test cases:
- removed implementations for apache, redis and mysql, moving functions to the structure for the test suite.
- keep the initialisation of godog's context and steps
- refactored regex in steps so that it captures service type too, making them reusable

metricbeat-tests/apache_test.go
```diff
@@ -1,35 +0,0 @@
...
-	"github.com/elastic/metricbeat-tests-poc/cli/services"
-)
-
-var apacheService services.Service
-
-func ApacheFeatureContext(s *godog.Suite) {
-	s.Step(`^Apache "([^"]*)" is running for metricbeat "([^"]*)"$`, apacheIsRunningForMetricbeat)
-	s.Step(`^metricbeat "([^"]*)" is installed and configured for Apache module$`, metricbeatIsInstalledAndConfiguredForApacheModule)
-	s.Step(`^there are no errors in the "([^"]*)" index$`, thereAreNoErrorsInTheIndex)
-}
```
jsoriano: Great to see this generalized. If we need to add some specific step for a service, I guess that we can still create another context with its own steps? Or we'd need to add the steps to the general `FeatureContext`.

mdelapenya: This is something on my radar, so we must investigate it

PR #39, opened 2019-09-09
ci(jenkins): golangci-lint

v1v:
## Highlights
- Support `golangci-lint` per changed file
- https://github.com/dnephin/pre-commit-golang/blob/4742d87b795071fd68d133f51e408c12518a9477/.pre-commit-hooks.yaml#L40-L46 doesn't support but apply for all the files independently.


v1v:
Squash and merge please :)

PR #48, opened 2019-09-24
(#44) Better management when cleaning up test resources

### LINKED ISSUE #44: Use a unique name for the index per test case
We want to avoid collisions between test cases

mdelapenya:
## What does this PR do?
The first thing we do in this PR is to consume the latest released version of the provisioner CLI (rc4).

Besides that, now it's not needed to write an index name at feature level, because it can be calculated internally by the test framework. To achieve that, we have moved code around, simplifying the codebase, enhancing the metricbeat test suite object, which will be responsible of running the metricbeat service.

We are also re-adding the deletion of the test index created by metricbeat, being deleted after tests run.

As a final point, we fix here the deletion of a running service, which was not done in a proper manner: docker-compose handles the deletion of a service in two phases: first it needs to stop the service and then it requires removing it.

## Why is it important?
The most important topic here, apart from the bugfix when removing services, is the improvements done for the sake of developer experience. We want developers to contribute tests to metricbeat in an easy manner, being the test framework the responsible for most of the things.

So then, a developer contributing a test will:
1. create a feature file, or updating an existing one, adding a Cucumber tag (`i.e. @redis`) representing the service to be run.
2. check if the test framework already provides the functionality, which nowadays covers starting/removing services, starting a pre-configured metricbeat service for it, and querying elasticsearch looking for two things: documents in the test index, and the absence of errors in that index.
3. if not existing, add it to the test framework in the form of a reusable Golang function.

## Related issues
Closes #44


metricbeat-tests/runner_test.go
```diff
@@ -56,23 +64,78 @@ func (mts *MetricbeatTestSuite) CleanUp() error {
...
 	return err
 }
 
+// As we are using an index per scenario outline, with an index name formed by metricbeat-version1-module-version2,
+// and because of the ILM is configured on metricbeat side, then we can use an asterisk for the index name:
+// each scenario outline will be namespaced, so no collitions between different test cases should appear
+func (mts *MetricbeatTestSuite) getIndexName() string {
+	mVersion := strings.ReplaceAll(mts.Version, "-SNAPSHOT", "")
+
+	index := fmt.Sprintf("metricbeat-%s-%s-%s", mVersion, mts.ServiceName, mts.ServiceVersion)
```
jsoriano: There would still be the potential problem if we have two scenarios using the same versions? Could we create here a random index name and store it in `mts` to solve this?

metricbeat-tests/features/apache.feature
```diff
@@ -4,7 +4,7 @@ Feature: As a Metricbeat developer I want to check that the Apache module works
 Scenario Outline: Check module is sending metrics to Elasticsearch
   Given Apache "<apache_version>" is running for metricbeat "<metricbeat_version>"
     And metricbeat "<metricbeat_version>" is installed and configured for Apache module
-  Then there are no errors in the "metricbeat-<metricbeat_version>-apache-<apache_version>" index
+  Then there are no errors in the index
```
jsoriano: :heart:

cli/services/manager.go
```diff
@@ -49,10 +49,35 @@ func (sm *DockerServiceManager) RemoveServicesFromCompose(stack string, composeN
...
+		if err != nil {
+			log.WithFields(log.Fields{
+				"command":  command,
+				"services": composeNames,
+				"stack":    stack,
+			}).Error("Could not stop services")
+			return err
+		}
+
+		command = []string{"rm", "-fv"}
```
jsoriano: Could the `--stop` flag of `rm` replace the previous kill?

metricbeat-tests/runner_test.go
mdelapenya: Yeah, I can send something on top of it

cli/services/manager.go
mdelapenya: Applied!

PR #49, opened 2019-09-26
(#44) random index name

mdelapenya:
## What does this PR do?
This PR adds the ability to use pseudo-randomised index name for a test suite, storing the index at test suite level.

## Why is it important?
It will allow to execute each test suite in a separated index, so there would be no document collisions from test suite to test suite.

### DIFF
--- a/metricbeat-tests/runner_test.go
+++ b/metricbeat-tests/runner_test.go
@@ -48,9 +48,7 @@ type MetricbeatTestSuite struct {
 // As we are using an index per scenario outline, with an index name formed by metricbeat-version1-module-version2,
 // and because of the ILM is configured on metricbeat side, then we can use an asterisk for the index name:
 // each scenario outline will be namespaced, so no collitions between different test cases should appear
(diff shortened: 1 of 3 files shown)


metricbeat-tests/runner_test.go
```diff
@@ -68,23 +84,11 @@ func (mts *MetricbeatTestSuite) CleanUp() error {
...
-
-	return index
-}
-
 func (mts *MetricbeatTestSuite) installedAndConfiguredForModule(version string, serviceType string) error {
 	serviceType = strings.ToLower(serviceType)
 
-	mts.Version = version
+	// at this point we have everything to define the index name
+	mts.setIndexName(version)
```
jsoriano: It looks a bit counterintuitive to set also the version in `setIndexName`. What do you think about explicitly set the version as before with `mts.Version = version`, and instead of having a set method for the index name, just have a `IndexName() string` method that generates and stores the index name for following calls.

PR #51, opened 2019-09-27
Refactor search retry to use iterative approach

jsoriano:

PR #54, opened 2019-09-30
Recover period of time collecting events with Metricbeat

jsoriano:
In #51 the method used to search with retries was refactored and the
wait to collect events from Metricbeat was removed. Recover this logic
but only in the test that requires it, leaving the method to
search with retries dedicated only to this.


metricbeat-tests/runner_test.go
```diff
@@ -24,18 +24,21 @@ var opt = godog.Options{Output: colors.Colored(os.Stdout)}
...
 // It can be overriden by OP_METRICBEAT_FETCH_TIMEOUT env var
 var queryMetricbeatFetchTimeout = 20
 
-// queryRetryTimeout Number of seconds between elasticsearch retry queries.
+// queryRetryTimeout is the number of seconds between elasticsearch retry queries.
 // It can be overriden by OP_RETRY_TIMEOUT env var
 var queryRetryTimeout = 3
 
+// runtimeDuration is the time Metricbeat is running and generating events during the test.
+var runtimeDuration = 20 * time.Second
```
mdelapenya: What about using https://github.com/elastic/metricbeat-tests-poc/pull/54/files#diff-2ee40ac993e969261cc73d4f5aece55aR33?

metricbeat-tests/runner_test.go
```diff
@@ -180,7 +183,13 @@ func (mts *MetricbeatTestSuite) thereAreNoErrorsInTheIndex() error {
...
 	stackName := "metricbeat"
 
-	result, err := retrySearch(stackName, mts.IndexName, esQuery, queryMaxAttempts)
+	_, err := retrySearch(stackName, mts.IndexName, esQuery, queryMaxAttempts)
+	if err != nil {
+		return err
+	}
+
+	time.Sleep(runtimeDuration)
+	result, err := retrySearch(stackName, mts.IndexName, esQuery, 3)
```
mdelapenya: Should we use a constant value as max attempts here?

metricbeat-tests/runner_test.go
jsoriano: Yes :man_facepalming:

metricbeat-tests/runner_test.go
jsoriano: Retry removed, this should work at this point.

mdelapenya:
Can we also move the warn log inside the IF condition in the search method?

```golang
log.WithFields(log.Fields{
	"attempt":       attempts,
	"errorCause":    err.Error(),
	"index":         indexName,
	"query":         esQuery,
	"retryAttempts": queryMaxAttempts,
	"retryTimeout":  queryRetryTimeout,
}).Warnf("Waiting %d seconds for the index to be ready", queryRetryTimeout)
if attempts > 1 {
	time.Sleep(time.Duration(queryRetryTimeout) * time.Second)
}
```

jsoriano:
@mdelapenya done! I have also added an info message when waiting the 20 seconds for events.

mdelapenya:
> @mdelapenya done! I have also added an info message when waiting the 20 seconds for events.

Info or Warn? 🤔

jsoriano:
> > @mdelapenya done! I have also added an info message when waiting the 20 seconds for events.
> 
> Info or Warn? thinking

Info for the message before waiting the 20 seconds for messages, I would not consider this a warn, maybe this could be debug.
And warn for the message in the search retries.

mdelapenya:
You convinced me!

The info is not while trying to connect, it's just waiting for metricbeat to fetch data. I would use a Debug instead.

One thing that I'm considering is using the feature file to specify the fetch timeout, something like:

```cucumber
Then there are no errors in the index after "<fetch_timeout>" seconds
Examples:
| apache_version | metricbeat_version | fetch_timeout |
| 2.2  | 7.3.0 | 60 |
```

This way each test suite could be setting its own needs, like the vSphere, which takes longer to connect.

jsoriano:
Ok, changed to debug.

> One thing that I'm considering is using the feature file to specify the fetch timeout, something like:
> 
> ```
> Then there are no errors in the index after "<fetch_timeout>" seconds
> Examples:
> | apache_version | metricbeat_version | fetch_timeout |
> | 2.2  | 7.3.0 | 60 |
> ```
> 
> This way each test suite could be setting its own needs, like the vSphere, which takes longer to connect.

+1, this can be different per integration, but to be done in a future PR.

metricbeat-tests/runner_test.go
```diff
@@ -248,15 +260,15 @@ func retrySearch(stackName string, indexName string, esQuery map[string]interfac
...
-		}).Warnf("Waiting %d seconds for the index to be ready", queryRetryTimeout)
 		if attempts > 1 {
+			log.WithFields(log.Fields{
+				"attempt":       attempts,
+				"errorCause":    err.Error(),
+				"index":         indexName,
+				"query":         esQuery,
+				"retryAttempts": queryMaxAttempts,
+				"retryTimeout":  queryRetryTimeout,
+			}).Warnf("Waiting %d seconds for the index to be ready", queryRetryTimeout)
```
jsoriano: @mdelapenya wdyt about changing this to debug too, this is something that is always going to happen, I don't think it worths to be logged at the warn level.

mdelapenya: Agreed

jsoriano: Done

PR #55, opened 2019-10-03
(#41) Test for default configuration

### LINKED ISSUE #41: Add test for included default configuration
Beats packages include generated configurations, we should add some test to check that this configuration works to avoid regressions like this one https://github.com/elastic/beats/issues/13705

A way to check it for metricbeat would be to start metricbeat with default configuration (the one in `metricbeat/metricbeat.yml` overriding the outputs with `-E`) and check that metricbeat starts and collects events from the `system` module.

mdelapenya:
## What does this PR do?
This PR adds a test case for the several configurations of metricbeat (default, reference and docker), ensuring that it stores system events in Elasticsearch and without errors.

For that reason we added another generic step, which could be reused in existing test cases: instead of simply indicating that they return no errors, it's better to make it explicit that they return events and no errors. Therefore we have separated both test verifications.

On the other hand, as this new test case does nothing to do with a monitored service, the index name generation has been changed to consider it. So now, for this particular case, the index name will contain metricbeat version only, plus a randomised string.

To get the default configuration for each version, we are downloading it from Github, so that each test case will use its own default configuration.

## Why is it important?
We need to ensure that metricbeat's default configuration is still valid even though no other module is configured.

## Related Issues
Closes #41 

## Author's checklist
- [x] Godog tests for metricbeat passing
- [x] Matrix configuration working
- [x] Download from github working
- [x] Existing tests in the same state (Apache and MySQL already generated error documents in the index, Redis is passing)


metricbeat-tests/runner_test.go
```diff
@@ -51,7 +50,12 @@ type MetricbeatTestSuite struct {
 func (mts *MetricbeatTestSuite) setIndexName() {
 	mVersion := strings.ReplaceAll(mts.Version, "-SNAPSHOT", "")
 
-	index := fmt.Sprintf("metricbeat-%s-%s-%s", mVersion, mts.ServiceName, mts.ServiceVersion)
+	var index string
+	if mts.ServiceName != "" {
+		index = fmt.Sprintf("metricbeat-%s-%s-%s", mVersion, mts.ServiceName, mts.ServiceVersion)
```
jsoriano: Actually there is no need to append the service if later on we are going to add a random suffix.

mdelapenya: Well, it's a matter of making index name more self-explanatory.

If index name generation gets more complicated after new use cases, let's simplify it

PR #56, opened 2019-10-04
(#41) Remove temporary configuration file

mdelapenya:


metricbeat-tests/runner_test.go
```diff
@@ -133,6 +143,8 @@ func (mts *MetricbeatTestSuite) installedUsingConfiguration(version string, conf
 		return err
 	}
 
+	mts.configurationFile = configurationFilePath
```
jsoriano: If we are storing the configuration file in an attribute of `mts` then we wouldn't need to pass it to `mts.runMetricbeatService()`.

mdelapenya: you are totally right!

mdelapenya: Mmm, thinking out loud, no. We just want to remove it when it's downloaded. Otherwise it will try to remove those in the workspace

jsoriano: It wouldn't change when the file is removed, right? I only mean to modify `runMetricbeatService` so it takes the configuration file from the `configurationFile` attribute.

mdelapenya: But if we set the mts.configurationFile at some point, then the CleanUp method will remove it

jsoriano: Yes, but this also happens now...

mdelapenya: It only removes the file for the `installedUsingConfiguration()` method, which is the only one setting that attribute.

If we use the attribute inside `runMetricbeatService()` then it must be set also here: `installedAndConfiguredForModule` (the method using default files). Therefore, it would be removed on cleanUp even for the second option

jsoriano: Ok, I get your point, you are right, sorry.

Maybe we should make explicit that setting `mts.configurationFile` is only intended for cleanup.

jsoriano: Or maybe we should create a temporary dir that gets removed on cleanup and place there all temporary files a test may need.

mdelapenya: I like the idea!

mdelapenya: Maybe in another PR?

jsoriano: Sure, you already had my approval :slightly_smiling_face:

mdelapenya: I added two commits: one making explicit the cleanUp of tmp files, and another one fixing a regression based by my previous pull

PR #58, opened 2019-10-08
57 read stack from environment

mdelapenya:
## What does this PR do?
It adds the ability to run a Elastic stack with a specific version, using a flag in the CLI side of this tool.

Besides that, we are simplifying the test framework, moving the version matrix for metricbeat and elastic stack to the automation tool (local or on CI). We will read the metricbeat version from an environment variable (`OP_METRICBEAT_VERSION`).

## Why is it important?
It will allow iterating faster, as the test framework will just run tests for one single version of the stack and metricbeat. It will be a responsibility of the automation tool to define the values used for metricbeat and the stack.

### DIFF
--- a/cli/cmd/run.go
+++ b/cli/cmd/run.go
@@ -31,6 +31,7 @@ func init() {
 	for k, stack := range config.AvailableStacks() {
 		stackSubcommand := buildRunStackCommand(k, stack)
 
+		stackSubcommand.Flags().StringVarP(&versionToRun, "stackVersion", "v", "latest", "Sets the stack version to run")
 		stackSubcommand.Flags().StringVarP(&servicesToRun, "withServices", "s", "", "Sets a list of comma-separated services to be depoyed alongside the stack")
 
 		runStackCmd.AddCommand(stackSubcommand)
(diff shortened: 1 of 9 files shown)


kuisathaverat:
Could be possible that a Beats version is not compatible with a stack? I mean, if I have the following test definition

```
@mysql
Feature: As a Metricbeat developer I want to check that the MySQL module works as expected

Scenario Outline: Check module is sending metrics to Elasticsearch without errors
  Given MySQL "<mysql_version>" is running for metricbeat
    And metricbeat is installed and configured for MySQL module
  Then there are "MySQL" events in the index
    And there are no errors in the index
Examples:
| mysql_version |
| 5.6           |
| 5.7           |
| 8.0           |
```

and Beats `5.6` is not compatible with Elastic Stack `7.5`, Could you launch the test for the Elastic stack `7.5`  by using an environment variable?

mdelapenya:
I think it's totally doable to run an elastic stack in 7.5 and beats in 5.x. The tests should be robust enough to fail. What we should predefine are the space of values when running from CI, which is the controlled environment.

kuisathaverat:
>I think it's totally doable to run an elastic stack in 7.5 and beats in 5.x.

I know, it was an example, it is possible that some version of beats does not work with some version of a service, it would be an exclusion from a test we have several in the integration test where old agent versions do not support newer versions of a framework. How do we manage those exceptions here?

mdelapenya:
Hey @jsoriano, what @kuisathaverat brings sound reasonable to me. Wdyt from beats development? Is this compatibility matrix a thing to consider from test framework or, on the other hand, it adds complexity to its maintenance?

jsoriano:
This would be something definitely nice to have because this will happen, for example when we want to add tests for a new module that is not supported in some released version.
But not sure what would be the best way to do it, for example in some multiversion cases we support now in tests, we handle the detection of supported features or versions from the test itself (not from the framework), it could be also the case here.

For example, could we add steps that skip the rest of tests if the condition is not met?
```
When metricbeat version is at least "7.5.0"
```

In any case I think that this could be done in a future change.

mdelapenya:
@kuisathaverat do we agree in this? We'd like to merge it to continue 🙏

PR #63, opened 2019-10-29
Fix paths

mdelapenya:
## What does this PR do?
It cleans up the GO environment variables in the pipeline, removing the GOROOT one.

Besides, it replaces the external dependency on CLI, as they live in the same repository

### DIFF
--- a/.ci/functionalTests.groovy
+++ b/.ci/functionalTests.groovy
@@ -99,8 +99,7 @@ pipeline {
       options { skipDefaultCheckout() }
       environment {
         GO111MODULE = 'on'
-        GOROOT = "${env.WORKSPACE}/.gimme/versions/go${env.GO_VERSION}.linux.amd64"
-        PATH = "${env.WORKSPACE}/${env.BASE_DIR}/bin:${env.GOROOT}/bin:${env.GOPATH}/bin:${HOME}/go/bin:${env.PATH}"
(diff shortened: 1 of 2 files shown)


.ci/functionalTests.groovy
```diff
@@ -99,7 +99,6 @@ pipeline {
       options { skipDefaultCheckout() }
       environment {
         GO111MODULE = 'on'
-        GOROOT = "${env.WORKSPACE}/.gimme/versions/go${env.GO_VERSION}.linux.amd64"
         PATH = "${env.WORKSPACE}/${env.BASE_DIR}/bin:${env.GOROOT}/bin:${env.GOPATH}/bin:${HOME}/go/bin:${env.PATH}"
```
jsoriano: `GOROOT` is being used here.

metricbeat-tests/go.mod
```diff
@@ -13,3 +13,5 @@ require (
 )
 
 replace github.com/testcontainers/testcontainers-go v0.0.8 => github.com/mdelapenya/testcontainers-go v0.0.9-compose
+
+replace github.com/elastic/metricbeat-tests-poc/cli v0.1.0-rc6 => ../cli
```
jsoriano: :heart: thanks!

.ci/functionalTests.groovy
mdelapenya: 🤦‍♂

metricbeat-tests/go.mod
mdelapenya: I did not do it before because I expected that path to be absolute. I realised it works when you screenshared the other day :)

PR #67, opened 2019-10-30
Small refactors

mdelapenya:
## What does this PR do?
It moves the metricbeat-specific test code to a separate Go file, leaving the test runner as slim as possible.

It adds a few `//nolint` comments to exclude false positives.

## Why is it important?
Boy Scout rule: _keep code a little bit healthier than it was_


metricbeat-tests/assertions.go
```diff
@@ -7,6 +7,7 @@ import (
 )
 
 // assertHitsArePresent returns an error if no hits are present
+//nolint:unused
```
jsoriano: Umm, these methods are used, why is this needed?

mdelapenya: I believe they are false positives in `golangci-lint`: https://apm-ci.elastic.co/blue/organizations/jenkins/beats%2Fpoc-metricbeat-tests-poc-mbp/detail/PR-67/1/pipeline

PR #76, opened 2019-11-13
(#68) Use when step properly

### LINKED ISSUE #68: Rephrase existing feature files to include a When
This `When` clause indicates to a casual reader the action that triggers the assertion performed in the `Then` one, so it's important to have one.

For the existing features, our current action (the implicit `When`) is to leave metricbeat running for 20 seconds (configurable). So we could rephrase the feature files like this:

Before:
```gherkin
  Given Apache "<apache_version>" is running for metricbeat
    And metricbeat is installed and configured for Apache module
  Then there are "Apache" events in the index
    And there are no errors in the index
```

After:
```gherkin
  Given Apache "<apache_version>" is running for metricbeat
    And metricbeat is installed and configured for Apache module
  When metricbeat is running for "20" seconds
  Then there are "Apache" events in the index
    And there are no errors in the index
```

Another way of achieving this could be using a document-based approach, polling elasticsearch until it has a certain number of documents into it:
```gherkin
  Given Apache "<apache_version>" is running for metricbeat
    And metricbeat is installed and configured for Apache module
//  When metricbeat has sent "20" documents
//  When the index contains at least "20" documents
  Then there are "Apache" events in the index
    And there are no errors in the index
```

## Related issues
- Relates https://github.com/elastic/metricbeat-tests-poc/issues/45

mdelapenya:
## What does this PR do?
We added two generic steps: the first for a _pure_ wait, and the second for running the service and wait a number of seconds to grab metrics from the service under test.

Besides that, we removed the environment variable that configured the fetch timeout, which was the same per test execution and caused to configure it at test time (in Jenkins or in command line).

We also used the Observability CI Docker images for MySQL and Apache, which were pushed yesterday as part of testing https://github.com/elastic/beats/pull/14355.

## Why is it important?
We allow each service to grab metrics during a defined number of seconds, different per service and defined at feature file level, instead of having all of them waiting a fixed number of seconds.

## Related issues
Closes #68


metricbeat-tests/features/apache.feature
```diff
@@ -4,9 +4,11 @@ Feature: As a Metricbeat developer I want to check that the Apache module works
 Scenario Outline: Check module is sending metrics to Elasticsearch without errors
   Given Apache "<apache_version>" is running for metricbeat
     And metricbeat is installed and configured for Apache module
+    And metricbeat waits "20" seconds for the service
```
exekias: Would it be possible to use containers health report instead of this?

metricbeat-tests/features/apache.feature
```diff
@@ -4,9 +4,11 @@ Feature: As a Metricbeat developer I want to check that the Apache module works
...
+  When metricbeat runs for "20" seconds
```
exekias: In the same sense, could we just run until we get documents (or a timeout)?

metricbeat-tests/features/apache.feature
mdelapenya: Adding a healthcheck at compose level could help I guess. On the other hand, testcontainers supports creating services with WaitFor strategies (https://github.com/testcontainers/testcontainers-go/tree/master/wait)

metricbeat-tests/features/apache.feature
mdelapenya: As explained at #68, there were two options: run a number of seconds, or run until a number of documents is present. We went the first path because it could be easier to control. But we could use the second approach if you consider a better one

exekias: Would like to hear @jsoriano thoughts on this one, but to me it sounds like number of documents with a good default (1) should be fine for most cases.

metricbeat-tests/features/apache.feature
exekias: My concern with both `when` clauses is the same: the more we push on arbitrary timings the more we will need to keep adjusting tests later. Also about how users could define the conditions for a healthy scenario, if we take them out of `docker-compose` that means we cannot reuse the scenario in current "unit" tests.

In general I would expect this to work without new parameters, but configuration on health checks.

jsoriano: Yeah, probably the `Given Apache "<apache_version>" is running for metricbeat` should be enough to ensure that Apache is running.

metricbeat-tests/features/apache.feature
jsoriano: I talked about this with Manu some days ago, I think that waiting for a number of documents, or for a number of seconds is quite equivalent in the case of Metricbeat, as the number of documents collected directly depends on the number of fetches executed, and this depends on the time Metricbeat is running.
Waiting for a number of documents is probably faster (depending on the polling period we  use to check that documents are there), but I also see the benefits on waiting to see that after some time everything looks good.
So I don't have a strong opinion here.
In any case, apart of checking that the documents are collected and they are correct, we need a timeout in case no document is collected.

metricbeat-tests/features/apache.feature
mdelapenya: I can confirm that after using the new observability-ci images (generated from Beats repo, which include HEALTCHECK's) the MySQL tests are passing, because the HEALTHCHECK is present in the Dockerfiles

metricbeat-tests/features/apache.feature
exekias: 20s per check can mean a lot of time in the long run (multiply that by tested versions and number of integrations). I could see the benefit of a "catch everything for a while" clause in some cases, but it sounds like waiting for a document is faster and less error prone.

This said, I'm ok with starting like this and creating an issue to improve things. But let's just not expose this setting to users in a `When` clause, we should not be fine tuning this wait tiems

mdelapenya: @exekias Let's merge this and improve that in the next iteration, the same we did in the past ;)

exekias: SGTM, could you please create an issue for that? :pray:

mdelapenya: https://github.com/elastic/metricbeat-tests-poc/issues/78

PR #77, opened 2019-11-13
Do not use the network when fetching Go dependencies

mdelapenya:
## What is this PR doing?
A dependency started to fail on CI and we enabled not using the network when fetching it.

From `go get` docs:

>The -u flag instructs get to use the network to update the named packages
and their dependencies. By default, get uses the network to check out
missing packages but does not use it to look for updates to existing packages

## Why is it important?
CI was not able to get dependencies, failing before any test execution

### DIFF
--- a/.ci/scripts/install-dependencies.sh
+++ b/.ci/scripts/install-dependencies.sh
@@ -17,10 +17,10 @@ GOPROXY_CONTEXT=${GOPROXY}
 GOPROXY=''
 
 # Install some other dependencies required for the pre-commit
-go get -v -u golang.org/x/lint/golint
-go get -v -u github.com/golangci/golangci-lint/cmd/golangci-lint@v1.18.0
-go get -v -u github.com/go-lintpack/lintpack/...
-go get -v -u github.com/go-critic/go-critic/...
+go get -v golang.org/x/lint/golint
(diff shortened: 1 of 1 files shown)


.ci/scripts/install-dependencies.sh
```diff
@@ -17,10 +17,10 @@ GOPROXY_CONTEXT=${GOPROXY}
...
 
 # Install some other dependencies required for the pre-commit
-go get -v -u golang.org/x/lint/golint
-go get -v -u github.com/golangci/golangci-lint/cmd/golangci-lint@v1.18.0
-go get -v -u github.com/go-lintpack/lintpack/...
-go get -v -u github.com/go-critic/go-critic/...
+go get -v golang.org/x/lint/golint
+go get -v github.com/golangci/golangci-lint/cmd/golangci-lint@v1.18.0
+go get -v github.com/go-lintpack/lintpack/...
+go get -v github.com/go-critic/go-critic/...
```
jsoriano: This can still connect to network if these dependencies are not already in the go path, are we copying them by other means?

mdelapenya: No, the flag will cause an update over network if needed. I checked that the flag causes the build to fail because of `go-critic`

PR #88, opened 2020-02-21
feat: Support removing certain blocks from Beats Integrations compose files

### LINKED ISSUE #87: Remove build context YAML section from fetched docker-compose files
They are causing many problems, as the first docker-compose file must have access to the location of each _meta directory in Beats.

## Related issues
- https://github.com/elastic/metricbeat-tests-poc/pull/84
- https://github.com/elastic/beats/pull/16311

mdelapenya:
## What does this PR do?
It adds a method to sanitize the compose files that come from the Beats repo, removing the `build` section under any service.

Besides that, we fixed an issue that affected the creation of parent directories when copying files, also overridding the files and directories in this process.

## Why is it important?
The Integrations compose files come with a `build` section, because it's used in there to build the images, but we do not want to use that feature, as it forces the docker-compose commands to have access to the the file system paths that are present in the `build.context` block.

We proposed those paths to be configurables (https://github.com/elastic/beats/pull/16311), but we realised it's better/easier to remove them for our internal purpose.

## Related issues
- Superseds https://github.com/elastic/beats/pull/16311
- Closes #87

### DIFF
--- a/cli/cmd/sync.go
+++ b/cli/cmd/sync.go
@@ -196,7 +196,6 @@ func sanitizeComposeFile(composeFilePath string) error {
 	if err != nil {
 		return err
 	}
-	io.WriteFile(d, composeFilePath)
 
-	return nil
+	return io.WriteFile(d, composeFilePath)
 }
--- a/cli/cmd/sync.go
+++ b/cli/cmd/sync.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"errors"
+	"fmt"
 	"os"
 	"path"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


cli/cmd/sync.go
```diff
@@ -113,5 +131,72 @@ func copyIntegrationsComposeFiles(beats git.Project, target string) {
...
+		}
+
+		c.Services[k] = output
+	}
+
+	d, err := yaml.Marshal(&c)
+	if err != nil {
+		return err
+	}
+	io.WriteFile(d, composeFilePath)
```
cachedout: Could this silently fail if the user doesn't have permissions to write to the file?

PR #89, opened 2020-02-27
fix: make sure parallel release stages have Go installed

### LINKED ISSUE #46: Fix packr builder image

mdelapenya:
## What does this PR do?
It installs Go in the each linux machine used for releasing each supported platform

## Why is it important?
If Go is not installed in the workers, then the Go build command will fail 🤷‍♂ 

See https://apm-ci.elastic.co/blue/organizations/jenkins/beats%2Fpoc-metricbeat-tests-poc-mbp/detail/v0.1.0-rc9/1/pipeline

## Related issues
- Closes #46

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -125,7 +125,7 @@ def generateStep(Map params = [:]){
         deleteDir()
         unstash 'build'
         dir("${BASE_DIR}") {
-          sh script: """.ci/scripts/install-go.sh "${GO_VERSION}" """, label: 'Install Go'
+          sh script: ".ci/scripts/install-go.sh '${GO_VERSION}' ", label: 'Install Go'
         }
         dir("${BASE_DIR}/cli") {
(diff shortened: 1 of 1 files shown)


.ci/Jenkinsfile
```diff
@@ -124,6 +124,9 @@ def generateStep(Map params = [:]){
       try {
         deleteDir()
         unstash 'build'
+        dir("${BASE_DIR}") {
+          sh script: """.ci/scripts/install-go.sh "${GO_VERSION}" """, label: 'Install Go'
```
kuisathaverat: `${var}` expressions are evaluated before to launch the command, so the following step if `GO_VERSION=1.12.3`

```suggestion
          sh script: ".ci/scripts/install-go.sh '${GO_VERSION}' ", label: 'Install Go'
```

it will run `.ci/scripts/install-go.sh '1.12.3' `, becasue the expresion was evaluated on Groovy time.

PR #90, opened 2020-02-28
fix: set proper Go environment on release

mdelapenya:
## What does this PR do?
It builds the CLI inside a Docker container from Makefile.

Besides that, we are removing Windows compilation because of:
>/go/pkg/mod/github.com/docker/docker@v0.7.3-0.20190506211059-b20a14b54661/pkg/system/filesys_windows.go:111:24: cannot use uintptr(unsafe.Pointer(&sd[0])) (type uintptr) as type *"golang.org/x/sys/windows".SECURITY_DESCRIPTOR in assignment

## Why is it important?
Using a Docker container for running tools avoid messing things up with the environment.

## Author's checklist
- [x] Tested with local environment
~~- [ ] Generation of windows binaries failing~~

Screenshot 2020-02-28 at 11 16 24

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -88,10 +88,6 @@ pipeline {
         }
         stage('Release') {
           options { skipDefaultCheckout() }
-          environment {
-            GOROOT = "${env.WORKSPACE}/.gimme/versions/go${env.GO_VERSION}.linux.amd64"
-            PATH = "${env.WORKSPACE}/${env.BASE_DIR}/bin:${env.PATH}:${env.GOROOT}/bin"
-          }
           when { tag "v*" }
           steps {
             deleteDir()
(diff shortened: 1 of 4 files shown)


mdelapenya:
@jsoriano, have you ever seen the above error when building the windows binaries?

.ci/Jenkinsfile
```diff
@@ -88,6 +88,10 @@ pipeline {
         }
         stage('Release') {
           options { skipDefaultCheckout() }
+          environment {
+            GOROOT = "${env.WORKSPACE}/.gimme/versions/go${env.GO_VERSION}.linux.amd64"
```
v1v: I'm a bit skeptical of declaring those env variables explicitly, could we use a docker container instead?

PR #97, opened 2020-03-06
ci: run e2e tests in the main pipeline

mdelapenya:
## What is this PR doing?
It adds a stage to run the e2e tests in parallel, so that a subset of the existing e2e tests are run in a separate worker each.

## Why is it important?
We want to verify that the e2e tests are still valid after changes in the test framework

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -165,7 +165,7 @@ def generateStep(Map params = [:]){
 def generateFunctionalTestStep(Map params = [:]){
   def feature = params.get('feature')
   return {
-    node('docker && linux && immutable') {
+    node('linux && immutable') {
       try {
         deleteDir()
         unstash 'build'
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -80,7 +80,7 @@ pipeline {
               }
               post {
                 always {
-                  junit(allowEmptyResults: true, keepLongStdio: true, testResults: "${BASE_DIR}/outputs/junit-*.xml")
+                  junit(allowEmptyResults: true, keepLongStdio: true, testResults: "${BASE_DIR}/outputs/TEST-unit.xml")
                 }
               }
             }
@@ -175,8 +175,8 @@ def generateFunctionalTestStep(Map params = [:]){
       } catch(e) {
         error(e.toString())
       } finally {
-        junit(allowEmptyResults: true, keepLongStdio: true, testResults: "${BASE_DIR}/outputs/junit-*.xml")
-          archiveArtifacts allowEmptyArchive: true, artifacts: "${BASE_DIR}/outputs/junit-*"
+        junit(allowEmptyResults: true, keepLongStdio: true, testResults: "${BASE_DIR}/outputs/TEST-${feature}.xml")
+          archiveArtifacts allowEmptyArchive: true, artifacts: "${BASE_DIR}/outputs/TEST-${feature}.xml"
       }
     }
   }
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -98,16 +98,18 @@ pipeline {
             GOPROXY = 'https://proxy.golang.org'
           }
           steps {
-            deleteDir()
-            unstash 'source'
-            dir("${BASE_DIR}") {
-              script {
-                def features = readYaml(file: '.ci/.e2e-tests.yaml')
-                def parallelTasks = [:]
-                features['FEATURES'].each { feature ->
-                    parallelTasks["${feature}"] = generateFunctionalTestStep(feature: "${feature}")
+            withGithubNotify(context: 'E2E Tests', tab: 'tests') {
+              deleteDir()
+              unstash 'source'
+              dir("${BASE_DIR}") {
+                script {
+                  def features = readYaml(file: '.ci/.e2e-tests.yaml')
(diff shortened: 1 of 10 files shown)


.ci/Jenkinsfile
```diff
@@ -86,6 +86,32 @@ pipeline {
...
           }
         }
+        stage('End-To-End Tests') {
+          parallel {
+            options { skipDefaultCheckout() }
+            environment {
+              GO111MODULE = 'on'
+              PATH = "${env.WORKSPACE}/${env.BASE_DIR}/bin:${env.GOPATH}/bin:${HOME}/go/bin:${env.PATH}"
+              GOPROXY = 'https://proxy.golang.org'
+              STACK_VERSION = "7.6.0"
```
mdelapenya: Hardcoded. As a follow-up we could define a place for runtime dependencies (like a properties file or .env file)

.ci/Jenkinsfile
```diff
@@ -137,3 +163,23 @@ def generateStep(Map params = [:]){
...
+def generateFunctionalTestStep(Map params = [:]){
+  def feature = params.get('feature')
+  return {
+    node('docker && linux && immutable') {
+      try {
+        deleteDir()
+        unstash 'build'
+        dir("${BASE_DIR}"){
+          sh script: """.ci/scripts/functional-test.sh "${GO_VERSION}" "${feature}" """, label: "Run functional tests for ${feature}"
+        }
```
mdelapenya: Copied from the functional-tests grooovy file

mdelapenya:
The CI is failing because of a git clone of Beats using the git protocol. I guess we must eval the ssh-agent to grant access, event though it's a public repo.

Any suggestions?

jsoriano:
It should be possible to clone the Beats repository without any credential.

What is the error git clone is reporting?

kuisathaverat:
see https://github.com/elastic/apm-integration-testing/blob/master/.ci/integrationTestEC.groovy#L39-L42 it is an example of how to make a git with the credentials we have

mdelapenya:
> It should be possible to clone the Beats repository without any credential.
> 
> What is the error git clone is reporting?

```
directory=/var/lib/jenkins/.op/git/beats url="https://github.com/elastic/beats"
[2020-03-06T17:30:16.515Z] time="2020-03-06T17:30:15Z" level=fatal msg="Cloning using keys from SSH agent failed"
```
This message is handled by the CLI in [here](https://github.com/elastic/metricbeat-tests-poc/blob/master/cli/internal/git.go#L177-L180), so I have enhanced it in last commit. BTW, I modified the git clone method to not handle SSH if the protocol is not Git (my bad)

mdelapenya:
Working!

BTW, @kuisathaverat in order to speed up the build we'd need to cache some images in the workers. It'd be a matter of just adding a `packer_cache.sh` script in the `.ci` dir, like in https://github.com/elastic/apm-agent-ruby/blob/master/.ci/packer_cache.sh. Do we need to do something on infra side too?

kuisathaverat:
the library script is fine to cache those images.

mdelapenya:
@v1v could you take a look to this line? https://apm-ci.elastic.co/blue/organizations/jenkins/beats%2Fpoc-metricbeat-tests-poc-mbp/detail/PR-97/6/pipeline/189#step-208-log-598

We are generating the XML file properly, but the `sed` command is not parsing last line properly, causing the XML file to be not well-formed: https://apm-ci.elastic.co/job/beats/job/poc-metricbeat-tests-poc-mbp/job/PR-97/6/artifact/src/github.com/elastic/metricbeat-tests-poc/outputs/TEST-apache.xml

I downloaded that file and there is no`</testsuites>` closing tag.

kuisathaverat:
I do not have context, but this `sed` command will replace white spaces or tabs on **one-line** and the will replace any occurrence of `>.*failed$` with `>`, the first filter that replaces **only one-line occurrences** I guess it is not correct, this will be the correct filter

`sed -e 's/^[ \t]*//g; s#>.*failed$#>#g' outputs/TEST-mysql`

mdelapenya:
> I do not have context, but this `sed` command will replace white spaces or tabs on **one-line** and the will replace any occurrence of `>.*failed$` with `>`, the first filter that replaces **only one-line occurrences** I guess it is not correct, this will be the correct filter
> 
> `sed -e 's/^[ \t]*//g; s#>.*failed$#>#g' outputs/TEST-mysql`

The last commit removes/hides `make`'s well-known “Entering/Leaving directory ...” messages, so I think it's fixed.

About caching docker images, we can deal with current set up and optimise if needed. My idea is to run only this three e2e tests, as a smoke test suite of the PoC

.ci/Jenkinsfile
```diff
@@ -86,6 +86,30 @@ pipeline {
...
           }
         }
+        stage('End-To-End Tests') {
+          options { skipDefaultCheckout() }
+          environment {
+            GO111MODULE = 'on'
+            PATH = "${env.WORKSPACE}/${env.BASE_DIR}/bin:${env.GOPATH}/bin:${HOME}/go/bin:${env.PATH}"
+            GOPROXY = 'https://proxy.golang.org'
+            STACK_VERSION = "7.6.0"
+            METRICBEAT_VERSION = "7.6.0"
```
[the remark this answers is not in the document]
mdelapenya: I like the idea of having them as input parameters! Thanks!

PR #100, opened 2020-03-12
Support for verification of the Helm charts using BDD approach

### LINKED ISSUE #99: Validate Helm charts
Let's use a BDD approach to validate the official Helm charts for elastic. Something like this:

```gherkin
@helm
@k8s
@metricbeat
Feature: The Helm chart is following product recommended configuration for Kubernetes

Scenario: The Metricbeat chart will create recommended K8S resources
  Given a cluster is running
  When the "metricbeat" Elastic's helm chart is installed
  Then a pod will be deployed on each node of the cluster by a DaemonSet
    And a "Deployment" will manage additional pods for metricsets querying internal services
    And a "kube-state-metrics" chart will retrieve specific Kubernetes metrics
    And a "ConfigMap" resource contains the "metricbeat.yml" content
    And a "ConfigMap" resource contains the "kube-state-metrics-metricbeat.yml" content
    And a "ServiceAccount" resource manages RBAC
    And a "ClusterRole" resource manages RBAC
    And a "ClusterRoleBinding" resource manages RBAC
```

mdelapenya:
## What is this PR doing?
It creates the first feature file and the implementation of the steps for verifying that the Elastic's Helm charts for Metricbeat are correct.

For that, each step in the feature file will represent each of the verification points required for the Helm chart.

The implementation of each step will leverage the local binaries for `helm` and `kubectl`, so each step will run different commands for these tools reading the output and checking those values.

The test cluster will be managed by [kind](https://kind.sigs.k8s.io/), and the binaries needed for it (kind, kubectl, helm) are installed by the shell script that runs the e2e tests.

## Why is it important?
We are going to manually verify the Helm charts, but adding this automated test suite will support the verification of the charts with any frequency: with a PR, on each release, on feature freeze, etc.

## How to test this PR?

1. Please check that you have `helm` and `kubectl` installed in your local machine.
1. From the `e2e` directory, run:
```shell
$ godog -t helm
```

## Related issues
- Closes #99


e2e/helm_charts_test.go
```diff
@@ -0,0 +1,166 @@
...
+
+func (ts *HelmChartTestSuite) elasticsHelmChartIsInstalled(chart string) error {
+	ts.install(chart)
+
+	return nil
+}
+
+func (ts *HelmChartTestSuite) podsManagedByDaemonSet() error {
+	args := []string{
+		"get", "daemonsets", "--namespace=default", "-l", "app=" + ts.Name + "-" + ts.Name, "-o", "jsonpath='{.items[0].metadata.labels.chart}'",
```
mdelapenya: @kuisathaverat is this enough to check that a DaemonSet exists for the chart?

kuisathaverat: This only ensures that the DaemonSet is deployed, but the DaemonSet pods can be in fail, so you have to check if the number of scheduled pods is the same as the pods ready(maybe is redundant) and the pods available.

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#daemonset-v1-apps


```
kubectl get -n kube-system daemonset metricbeat -o jsonpath='{.status.currentNumberScheduled}'
kubectl get -n kube-system daemonset metricbeat -o jsonpath='{.status.numberAvailable}'
kubectl get -n kube-system daemonset metricbeat -o jsonpath='{.status.numberReady}'
```

mdelapenya: Wouldn't it mean we are checking that k8s is doing its job? 🤔

.ci/scripts/install-test-dependencies.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+#
+# Install the dependencies using in tests.
+#
+
+# install kubectl
+curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt`/bin/linux/amd64/kubectl
+chmod +x ./kubectl
+mv ./kubectl /usr/local/bin/kubectl
```
mdelapenya: Can we use `sudo` in a worker? Otherwise this will fail

kuisathaverat: no, we can not, because of that we install everything in the Jenkins workspace job

e2e/helm_charts_test.go
kuisathaverat: no, you are checking the configuration settings of the pods are correct and can deploy

mdelapenya:
Sample execution:
Screenshot 2020-03-12 at 15 20 26

.ci/scripts/install-test-dependencies.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+#
+# Install the dependencies using in tests.
```
cachedout: Should we find a way to bail out of this script or return an error code if all the dependencies were not installed successfully?

cli/internal/git.go
```diff
@@ -141,14 +143,21 @@ func Clone(repositories ...Project) {
 func GetBranch(gitRepositoryDir string) string {
 	args := []string{"rev-parse", "--abbrev-ref", "HEAD"}
 
-	return Execute(gitRepositoryDir, "git", args[0:]...)
+	branch, err := shell.Execute(gitRepositoryDir, "git", args[0:]...)
+	if err != nil {
+		return ""
```
cachedout: Given this, should we add some checks to ensure that on L55 we don't have an empty string?

cli/internal/git.go
```diff
@@ -141,14 +143,21 @@ func Clone(repositories ...Project) {
...
 }
 
 // GetRemote returns the remote from a git repository
 func GetRemote(gitRepositoryDir string, gitDomain string) string {
 	args := []string{"remote", "get-url", "origin"}
 
-	remote := Execute(gitRepositoryDir, "git", args[0:]...)
+	remote, err := shell.Execute(gitRepositoryDir, "git", args[0:]...)
+	if err != nil {
+		return ""
```
cachedout: Same comment as about but with L56

kuisathaverat:
this will work only with Helm 2.x, the commands are a little different for Helm 3.x

mdelapenya:
> this will work only with Helm 2.x, the commands are a little different for Helm 3.x

I'm using Helm 3:
```
$ helm version
version.BuildInfo{Version:"v3.1.1", GitCommit:"afe70585407b420d0097d07b21c47dc511525ac8", GitTreeState:"clean", GoVersion:"go1.13.8"}
```

cli/internal/git.go
mdelapenya: Yes, this is something I added on purpose to be able to continue with the test side. Now that the feature file passes locally, I can dedicate more time for this

kuisathaverat:
>I'm using Helm 3:

ouch!!, in that case, is the other way around would not work on Helm 2.x

.ci/scripts/install-test-dependencies.sh
mdelapenya: In last commit I address this: instead of checking the tools when the Golang code is initialised, it's done as part of the BeforeScenario of the test suite. Then the test, if executed, will fail because the tools are not there.

In the other hand, we must install the tools in the workers, using Infra scripts. Wdyt?

mdelapenya:
> > I'm using Helm 3:
> 
> ouch!!, in that case, is the other way around would not work on Helm 2.x

Which versions should we support, latest, all LTS?

kuisathaverat:
>Which versions should we support, latest, all LTS?

latest should be fine (2.15.3?)

jmlrt:
We currently support [Helm 2.16.1](https://github.com/elastic/helm-charts/#helm-versions), latest Helm 2 version ([2.16.3](https://github.com/helm/helm/releases/tag/v2.16.3)) should be fine (I'll create a PR to bump it soon).

As for Helm 3 we currently don't support it but we are currently studying migration impacts and may add its support for 7.7.0 regarding the required workload.

cli/services/helm.go
```diff
@@ -0,0 +1,83 @@
...
+func HelmFactory(version string) (HelmManager, error) {
+	var helm HelmManager
+	if version == "3.1.1" {
+		return &helm3_1_1{}, nil
+	}
+
+	return helm, errors.New("Sorry, we don't support Helm v" + version + " version")
+}
+
+type helm3_1_1 struct {
```
mdelapenya: @kuisathaverat Creating a struct for Helm v2 we will support different command execution.

Please let me know if you can help me with the commands

kuisathaverat: sure

cli/services/helm.go
```diff
@@ -63,7 +76,83 @@ func (h *helm3_1_1) DeleteChart(chart string) error {
...
+	log.WithFields(log.Fields{
+		"output": output,
+		"chart":  chart,
+	}).Debug("Chart deleted")
+	return nil
+}
+
+func (h *helm2_16) Init() error {
+	args := []string{
+		"init", "--namespace", "default",
```
mdelapenya: This command is failing because the underlying image is not pulled from gcr.io:

```
$ kubectl get pods --namespace default
NAME                             READY   STATUS         RESTARTS   AGE
tiller-deploy-7dffd9c7c7-6s8vn   0/1     ErrImagePull   0          21s
```

cli/services/helm.go
```diff
@@ -63,7 +76,83 @@ func (h *helm3_1_1) DeleteChart(chart string) error {
 	return nil
 }
 
-func (h *helm3_1_1) InstallChart(name string, chart string, version string) error {
+func (h *helm3_1) Init() error {
+	return nil
```
mdelapenya: We do not need helm initialisation in Helm 3.1.x

e2e/helm_charts_test.go
```diff
@@ -31,17 +30,22 @@ func init() {
...
+	Name        string // the name of the chart
+	Version     string // the helm chart version for the test
 }
 
 func (ts *HelmChartTestSuite) aClusterIsRunning() error {
-	args := []string{"status"}
+	args := []string{"get", "clusters"}
 
-	output, err := shell.Execute(".", "minikube", args...)
+	output, err := shell.Execute(".", "kind", args...)
```
mdelapenya: Thanks @kuisathaverat for the suggestion about kind!
image

e2e/helm_charts_test.go
```diff
@@ -276,7 +274,7 @@ func HelmChartFeatureContext(s *godog.Suite) {
 		toolsAreInstalled()
 
 		testSuite.addElasticRepo()
-		testSuite.createCluster()
+		testSuite.createCluster("1.15.3")
```
mdelapenya: This version will be configurable in a follow-up PR (or commit)

cli/services/helm.go
```diff
@@ -139,22 +139,42 @@ func (h *helm2_16) DeleteChart(chart string) error {
...
+
+	args = []string{"create", "clusterrolebinding", "tiller", "--clusterrole=cluster-admin", "--serviceaccount=kube-system:tiller"}
+	output, err = shell.Execute(".", "kubectl", args...)
+	if err != nil {
+		log.Error("Could not create Tiller ClusterRoleBinding")
+		return err
 	}
+	log.WithFields(log.Fields{
+		"output": output,
+	}).Debug("Tiller ClusterRoleBinding created")
```
mdelapenya: See https://rancher.com/docs/rancher/v2.x/en/installation/options/helm2/helm-init/

cli/services/helm.go
```diff
@@ -139,22 +139,42 @@ func (h *helm2_16) DeleteChart(chart string) error {
...
 
-	_, err := helmExecute(args...)
+	args = []string{
+		"init", "--wait", "--service-account", "tiller",
```
mdelapenya: Waiting to make sure the steps have everything in place.

The service account is needed to find the Tiller, but just for Helm 2.x

.ci/scripts/install-test-dependencies.sh
cachedout: > In the other hand, we must install the tools in the workers, using Infra scripts. Wdyt?

This seems like a reasonable compromise to me. I like the idea of doing a pre-flight check in the `BeforeScenario`.

e2e/features/filebeat/helm.feature
```diff
@@ -0,0 +1,13 @@
+@helm
+@k8s
+@filebeat
+Feature: The Helm chart is following product recommended configuration for Kubernetes
+
+Scenario: The Filebeat chart will create recommended K8S resources
```
mdelapenya: The magic of BDD! Combining existing steps to add more tests!

cli/services/helm.go
mdelapenya: Solved using Kind instead of minikube

e2e/helm_charts_test.go
```diff
@@ -0,0 +1,311 @@
...
+		log.Fatalf("Could not add Elastic Helm repo. Aborting: %v", err)
+	}
+}
+
+func (ts *HelmChartTestSuite) aResourceContainsTheContent(resource string, content string) error {
+	lowerResource := strings.ToLower(resource)
+	escapedContent := strings.ReplaceAll(content, ".", `\.`)
+
+	args := []string{
+		"get", lowerResource + "s", ts.getResourceName(resource), "-o", `jsonpath="{.data['` + escapedContent + `']}"`,
```
jsoriano: From the feature file I thought that the condition was going to check that the content of the file was present in a resource.

I.e. that `And a "ConfigMap" resource contains the "metricbeat.yml" content` would check that there is a config map with the content of the `metricbeat.yml` file. But this is checking that there is a config map that contains the `metricbeat.yml` string, right?

mdelapenya: Yes, I went this way for now to simplify and not having to download the original `metricbeat.yml` file, but you are right: that would be more accurate: comparing the contents of the file with the result of the configMap.

jsoriano: Oh, but I think that is fine to check that some content is in some resource, only that the current text confused me as it was only used with files. Maybe it could be rephrased to something like `a "ConfigMap" resource contains the "metricbeat.yml" string`.

mdelapenya: wdyt about `a "ConfigMap" resource contains the "metricbeat.yml" key`? Because it's a key

jsoriano: Oh, even better, yes, thanks!

mdelapenya: Done!

e2e/helm_charts_test.go
```diff
@@ -311,6 +311,53 @@ func (ts *HelmChartTestSuite) resourceWillManageAdditionalPodsForMetricsets(reso
...
+	lowerResource := strings.ToLower(resource)
+	strategyKey := "strategy"
+	name := ts.Name + "-" + ts.Name + "-metrics"
+
+	if lowerResource == "daemonset" {
+		strategyKey = "updateStrategy"
+		name = ts.Name + "-" + ts.Name
+	}
+
+	output, err := kubectl.Run("get", lowerResource, name, "-o", `go-template='{{.spec.`+strategyKey+`.type}}'`)
```
mdelapenya: @kuisathaverat same here

e2e/helm_charts_test.go
```diff
@@ -311,6 +311,53 @@ func (ts *HelmChartTestSuite) resourceWillManageAdditionalPodsForMetricsets(reso
 	return nil
 }
 
+func (ts *HelmChartTestSuite) strategyCanBeUsedDuringUpdates(strategy string) error {
+	output, err := kubectl.Run("get", "ds/"+ts.Name+"-"+ts.Name, "-o", `go-template='{{.spec.updateStrategy.type}}'`)
```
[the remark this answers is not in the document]
kuisathaverat: the command is correct if the DaemonSet is in the default namespace, however, to match other commands we have I'd use the full qualified name of the resource.

```suggestion
	output, err := kubectl.Run("get", "daemonset", ts.Name+"-"+ts.Name, "-o", `go-template='{{.spec.updateStrategy.type}}'`)
```

e2e/helm_charts_test.go
```diff
@@ -0,0 +1,521 @@
...
+	return nil
+}
+
+func (ts *HelmChartTestSuite) strategyCanBeUsedDuringUpdates(strategy string) error {
+	output, err := kubectl.Run("get", "ds/"+ts.Name+"-"+ts.Name, "-o", `go-template='{{.spec.updateStrategy.type}}'`)
+	if err != nil {
+		return err
+	}
+
+	output = strings.ReplaceAll(output, "'", "")
```
kuisathaverat: the output does not have `'` characters

e2e/helm_charts_test.go
kuisathaverat: both are the same command, the only difference is the go-template

mdelapenya: I'm calling one method from the other one.

e2e/helm_charts_test.go
mdelapenya: Solved! The go-template I was using included extra quotes, which I removed

e2e/helm_charts_test.go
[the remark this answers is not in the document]
mdelapenya: Solved too

cli/internal/git.go
mdelapenya: I reviewed that part and noticed it's dead code that I've removed in last commit (ba1e5d5f166f2c260bd175268681e07b7aa99cef)

cli/internal/git.go
mdelapenya: Same here: I reviewed that part and noticed it's dead code that I've removed in last commit (ba1e5d5f166f2c260bd175268681e07b7aa99cef)

PR #104, opened 2020-04-09
Use "go test" as test runner

mdelapenya:
## What is this PR doing?
It uses `go test` instead of the godog binary, so we are removing its installation in the Makefile.

We also remove all gherkin tags supporting godog's filters, creating a configuration layer (a map) with the supported modules. This is needed because at this moment godog is executing all life cycle methods (befores and afters), even when filtering by tag.

We are replacing the usage of tags with the arguments passed to the `go test` command:

```shell
$ go test -v --godog.format pretty redis mysql apache
```

## Why is it important?
With this configuration we are able to define which FeatureContext functions are contributed to the test runner, so inspecting the arguments we will be able to check if it's a supported module. This way, only those contributors of interest will be added to the test suite, only executing the life cycle methods of interest.

## Related issues
I created a sample repo demonstrating this life cycle issue: http://github.com/mdelapenya/sample-godog

### DIFF
--- a/.ci/scripts/functional-test.sh
+++ b/.ci/scripts/functional-test.sh
@@ -52,6 +52,6 @@ if ! FEATURE=${FEATURE} FORMAT=junit STACK_VERSION=${STACK_VERSION} METRICBEAT_V
 fi
 
 ## Transform report to Junit by parsing the stdout generated previously
-sed -i -E "s/testing: warning: no tests to run//" ${REPORT}
+sed -i "s/testing: warning: no tests to run//" ${REPORT}
 sed -e 's/^[ \t]*//; s#>.*failed$#>#g' ${REPORT} | grep -E '^<.*>$' > ${REPORT}.xml
 exit $exit_status
--- a/.ci/scripts/functional-test.sh
+++ b/.ci/scripts/functional-test.sh
@@ -52,5 +52,6 @@ if ! FEATURE=${FEATURE} FORMAT=junit STACK_VERSION=${STACK_VERSION} METRICBEAT_V
 fi
 
 ## Transform report to Junit by parsing the stdout generated previously
+sed -i -E "s/testing: warning: no tests to run//" ${REPORT}
 sed -e 's/^[ \t]*//; s#>.*failed$#>#g' ${REPORT} | grep -E '^<.*>$' > ${REPORT}.xml
 exit $exit_status
--- a/.ci/scripts/functional-test.sh
+++ b/.ci/scripts/functional-test.sh
@@ -35,10 +35,8 @@ rm -rf outputs || true
 mkdir -p outputs
 
 ## Parse FEATURE if not ALL then enable the flags to be passed to the functional-test wrapper
-FLAG=''
 REPORT=''
 if [ "${FEATURE}" != "" ] && [ "${FEATURE}" != "all" ] ; then
-  FLAG='-t'
   REPORT=outputs/TEST-${FEATURE}
 else
   FEATURE=''
@@ -48,7 +46,7 @@ fi
 ## Generate test report even if make failed.
 set +e
 exit_status=0
-if ! FLAG=${FLAG} FEATURE=${FEATURE} FORMAT=junit STACK_VERSION=${STACK_VERSION} METRICBEAT_VERSION=${METRICBEAT_VERSION} make --no-print-directory -C e2e functional-test | tee ${REPORT}  ; then
+if ! FEATURE=${FEATURE} FORMAT=junit STACK_VERSION=${STACK_VERSION} METRICBEAT_VERSION=${METRICBEAT_VERSION} make --no-print-directory -C e2e functional-test | tee ${REPORT}  ; then
   echo 'ERROR: functional-test failed'
   exit_status=1
 fi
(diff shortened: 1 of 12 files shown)


mdelapenya:
The existing CI failure [https://apm-ci.elastic.co/blue/organizations/jenkins/beats%2Fpoc-metricbeat-tests-poc-mbp/detail/PR-104/1/pipeline/196] happens because `go test` is printing `testing: warning: no tests to run` after test execution and our junit parser is not able to generate the XML properly.

I read the Go test code, and this message happens only here:
- https://golang.org/src/testing/testing.go?#L1203
- https://golang.org/src/testing/testing.go?#L1259

@jsoriano any idea why it happens?

mdelapenya:
I saw the problem: the grep parser is removing the last line, which closes the testsuites element (`</testsuites>`), as this line is continued with the `testing: warning: no tests to run`

.ci/scripts/functional-test.sh
```diff
@@ -52,5 +52,6 @@ if ! FEATURE=${FEATURE} FORMAT=junit STACK_VERSION=${STACK_VERSION} METRICBEAT_V
 fi
 
 ## Transform report to Junit by parsing the stdout generated previously
+sed -i -E "s/testing: warning: no tests to run//" ${REPORT}
```
mdelapenya, line 55: Using sed to remove the `go test` string from the output file

mdelapenya:
With this approach, we are removing 45 minutes to each branch in the parallel tests, as the kind cluster is not provisioned if not needed.

e2e/metricbeat-poc.md
```diff
@@ -46,28 +46,35 @@ For this POC, we have chosen Godog over any other test framework because the tea
 
 All the Gherkin (Cucumber) specifications are written in `.feature` files.
 
-A good example could be [this one](./features/mysql.feature):
+A good example could be [this one](./features/metricbeat/mysql.feature):
 
 ```cucumber
 @mysql
```
kuisathaverat, line 52: Should this tag be removed?

mdelapenya: Good catch!

e2e/runner_test.go
```diff
@@ -3,15 +3,77 @@ package e2e
...
+func (c *contextMetadata) getFeaturePaths() []string {
+	paths := []string{}
+	for _, module := range c.modules {
+		paths = append(paths, path.Join("features", module, c.name+".feature"))
+	}
+
+	return paths
+}
+
+var supportedProducts = map[string]*contextMetadata{
```
kuisathaverat, line 32: this structure can grow a lot, and decouple test from categories, it is odd for me.

mdelapenya: Yeah, same for me :( I think this is a workaround until I find how to fix it in the upstream project (as explained here https://github.com/mdelapenya/sample-godog)

.ci/scripts/functional-test.sh
v1v: If running in OSX then `sed -i` won't work. I'm not sure what's the requirements for `-E` but worth to double check.

See [sed-diff](https://unix.stackexchange.com/questions/13711/differences-between-sed-on-mac-osx-and-other-standard-sed)

mdelapenya: Yeah, it's really ugly that the godog does not support  a native XML output, instead of using the console. I'm thinking about opening an issue about this

mdelapenya: Another thing: are we going to run the jUnit output in a different worker than Linux? I tested this command in my local machine (mac) and in the worker (https://apm-ci.elastic.co/job/beats/job/poc-metricbeat-tests-poc-mbp/view/change-requests/job/PR-104/3/) and it works.

v1v: I've just run the below command in my mac:

```
$ echo '/testing: warning: no tests to run' > /tmp/test.txt
$ cat test.txt 
testing: warning: no tests to run
$ sed -i -E "s/testing: warning: no tests to run//" /tmp/test.txt 
$ ls -l test.txt*
-rw-r--r--  1 vmartinez  wheel   2 13 Apr 19:51 test.txt
-rw-r--r--  1 vmartinez  wheel  35 13 Apr 19:51 test.txt-E
$ cat test.txt 

```

In OSX, the `-E` is not used as far as I see, and the `-i` uses the `-E` as the suffix for the backup file, not sure if that's intended.

If `-E` refers to the regex format, I guess `testing: warning: no tests to run` is not a regex but a fully match?

mdelapenya: I removed the `-E` flag in 8720619 for simplicity

mdelapenya:
We are going to merge this PR but knowing that it's a bit workaround that adds complexity and maintainability concerns: the map linking test modules and the FeatureContext functions could grow A LOT.

This is a design decision that we made to bypass [the issue that we currently see in Godog](https://github.com/cucumber/godog/issues/196), where all the FeatureContext functions are dynamically exported when running godog (it builds a test binary on the fly), so every test Suite containing life cycle hook (Befores and Afters) will contribute those methods to every execution. In our particular case, the Helm charts test suite starts a Kubernetes cluster using Kind, which is very expensive in time and resources, and we do not want to run that code for any other test suite, possibly causing a performance degradation in the build agent running any other test suite making it fail.

For this reason, after talking and agreeing with @jsoriano, we want to make progress on this project, but having this in mind for future reference.

Issue #107, opened 2020-04-10
Check services status when running them


mdelapenya:
I can think of improving the `runMetricbeat` method, to check/wait for liveness of the service, and if it's not ready, then throw an error, failing the test at that step.

_Originally posted by @mdelapenya in https://github.com/elastic/metricbeat-tests-poc/issues/103#issuecomment-611991489_

That would also apply to any other service run by the framework

mdelapenya:
Waiting for testcontainers to support this feature: https://github.com/testcontainers/testcontainers-go/pull/308. Once released we can use the new version of the library.

mdelapenya:
Release is out! https://github.com/testcontainers/testcontainers-go/releases/tag/v0.11.0

I think we can leverage it in the framework cc/ @adam-stokes

PR #108, opened 2020-04-10
Fix metricbeat tests

### LINKED ISSUE #103: metricbeat configuration tests are failing (not running in the CI)
```shell
--- Failed steps:

  Scenario Outline: Check metricbeat configuration is sending metrics to Elasticsearch without errors # features/metricbeat/metricbeat.feature:3
    Then there are "system" events in the index # features/metricbeat/metricbeat.feature:6
      Error: There aren't documents for - on Metricbeat index

  Scenario Outline: Check metricbeat.docker configuration is sending metrics to Elasticsearch without errors # features/metricbeat/metricbeat.feature:3
    Then there are "system" events in the index # features/metricbeat/metricbeat.feature:6
      Error: There aren't documents for - on Metricbeat index

  Scenario Outline: Check metricbeat.reference configuration is sending metrics to Elasticsearch without errors # features/metricbeat/metricbeat.feature:3
    Then there are "system" events in the index # features/metricbeat/metricbeat.feature:6
      Error: Could not send query to Elasticsearch in the specified time (15 seconds)
```

mdelapenya:
## What is this PR doing?
It cleans up a bit the code for running metricbeat, as we detected it was called twice for the metricbeat configuration tests. As part of this refactor, we enhanced a few logs in the process, so that we can discover what happens more easily.

The fix is done in the way that the testSuite struct recieved the attributes in a wrong manner, setting the "system" module but being overriden in the second run.

Finally we add the metricbeat to the parallel execution on CI, which was not there.

## Why is it important?
The CI was not running these tests for metricbeat, and we see it important to check the configuration files too.

## Related issues
- Closes #103
- Discovered while testing #107


mdelapenya:
The metricbeat tests will hace one failure: still checking it. I'll keep this PR in draft until I fix it.

The root cause for this failure is described here: https://github.com/elastic/metricbeat-tests-poc/issues/103#issuecomment-611991006

e2e/metricbeat_test.go
```diff
@@ -32,44 +29,62 @@ func init() {
...
+func (mts *MetricbeatTestSuite) getIndexName() string {
+	return mts.Query.IndexName
+}
+
+func (mts *MetricbeatTestSuite) setEventModule(eventModule string) {
+	mts.Query.EventModule = eventModule
 }
 
 // As we are using an index per scenario outline, with an index name formed by metricbeat-version1-module-version2,
+// or metricbeat-version1-module-variant-version2,
```
jsoriano, line 52: Ok to continue with variants as something special by now, but take into account that there may be more variables, for example we already have a "codename" in ceph tests.

mdelapenya: Let's find in a follow-up PR a better way of index naming based on the discovered variables.

mdelapenya:
The two tests are failing maybe because of the interaction with the kind cluster, which is run by the before method in the helm chart test suite. I'll mark this PR as ready once we merge #104

.ci/Jenkinsfile
```diff
@@ -25,6 +25,8 @@ pipeline {
   }
   parameters {
     choice(name: 'LOG_LEVEL', choices: ['INFO', 'DEBUG'], description: 'Log level to be used')
+    choice(name: 'QUERY_MAX_ATTEMPTS', choices: ['5', '10', '20'], description: 'Number of attempts to create the cnnection to Elasticsearch')
```
mdelapenya, line 28: Let's use Jenkins UI to provide different retry strategies for internal retries and timeouts.

If we find it's a timeout problem, because the build agent is overloaded for some reason, then we can use a backoff strategy in the internal retry algorithm, instead of using a fixed (although configurable) one: 5 retries, waiting 3 seconds between each.

cachedout, line 28:
```suggestion
    choice(name: 'QUERY_MAX_ATTEMPTS', choices: ['5', '10', '20'], description: 'Number of attempts to create the connection to Elasticsearch')
```

mdelapenya: Good catch!

e2e/utils_test.go
```diff
@@ -0,0 +1,25 @@
+package e2e
+
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDownloadFile(t *testing.T) {
```
mdelapenya, line 10: @jsoriano I think the error on CI comes from an issue reading the downloaded file. I've added this test for it, but did not have time to run it on CI. Could you please check it out in your morning? We can talk about it later on tomorrow.

Thanks in advance!

mdelapenya: BTW, I think it happened on your local machine too, could you please run this test locally?

mdelapenya:
After a lot of debugging, even keeping a CI agent, SSH'ing into it and running commands there, we finally fixed. The problem was in how docker-compose handles mounted files in Linux: the file must have 0666 permission (read for all users), otherwise it cannot be mounted. This was hard to find and thanks to @jsoriano we realised the root cause.

image

In the between, we:
- Removed kibana from the stack, reducing 1-2 mins to the build
- Added more logs
- Refactored small things
- Added retry-backoff strategy to wait for elasticsearch to be ready
- Added parameters to the pipeline to control de timeouts and retries
- Removed dependency with the `op` binary for preparing the runtime dependencies (the stack)
- Used life cycle hook methods (BeforeSuite/AfterSuite) to run/destroy runtime dependencies (the stack)
- Added container logs when running the pipeline in debug mode
- Used pipeline log level as metricbeat log level

image

PR #112, opened 2020-04-24
chore: add vsphere tests to the CI

### LINKED ISSUE #111: Add vsphere tests to the CI

mdelapenya:
## What does this PR do?
It add vsphere to the CI, using same variable name than in Beats repo

## Why is it important?
It increases the coverage for integrations modules

## Related issues
- Closes #111


cli/config/compose/services/vsphere/docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 version: '2.3'
 services:
   vsphere:
-    image: "nimmis/vcsim:${vsphereTag}"
+    image: "nimmis/vcsim:${VSPHERE_VERSION}"
```
mdelapenya, line 4: The variables coming from Beats repo are following this name convention, although this vsphere is not including a supported-versions.yml file. I.e. apache:
https://github.com/elastic/beats/blob/master/metricbeat/module/apache/_meta/supported-versions.yml#L1-L3

PR #113, opened 2020-04-24
chore: add skeleton for the stack-monitoring tests

### LINKED ISSUE #94: Port over one of the stack monitoring parity tests
Hi,

Currently we have stack monitoring parity tests running in the Elastic Stack Testing Framework (ESTF).  You can read more about these tests, particularly why they were created and how they work conceptually, over here: https://github.com/elastic/elastic-stack-testing/blob/master/playbooks/monitoring/README.md.

These parity tests are broken up by each stack product that shows up in the Stack Monitoring UI, viz. elasticsearch, logstash, kibana, and beats. This division is reflected in the sub-folders seen under https://github.com/elastic/elastic-stack-testing/tree/master/playbooks/monitoring.

It would be great to try and port over one of these product's parity tests over to the metricbeat-tests-poc framework, mainly to see if there are any performance or maintainability gains in the process. To that end, I'd suggest porting over the parity tests for beats or logstash, as these are the least complex ones.

mdelapenya:
## What is this PR doing?
It ports the existing Python parity tests for stack monitoring to Go

## Why is it important?
We want to add support for Stack Monitoring in this test framework.

## How to test these changes?
You could run: 

```shell
$ cd e2e
$ OP_LOG_LEVEL=INFO go test -v --godog.format junit parity-tests
```

or using VSCode, supporting debug capabilities:
- Add breakpoints in the functions you're interested
- Open runner_test.go file as the current selected file in the editor
- Open Debug panel
- Select "Parity Tests" in the Run selector

## Related issues
- Closes #94

## Logs

<details><summary>Expand to view the logs</summary>
<p>
<pre><code>
OP_LOG_LEVEL=DEBUG go test -v --godog.format pretty parity-tests
DEBU[0000] Validating required tools: [docker docker-compose] 
DEBU[0000] Binary is present                             binary=docker path=/usr/local/bin/docker
DEBU[0000] Binary is present                             binary=docker-compose path=/usr/local/bin/docker-compose
DEBU[0000] 'op' workdirs created.                        servicesPath=/Users/mdelapenya/.op/compose/services stacksPath=/Users/mdelapenya/.op/compose/stacks
DEBU[0000] Boxed file                                    path=services/apm-server/docker-compose.yml service=apm-server
DEBU[0000] Boxed file                                    path=services/elasticsearch/docker-compose.yml service=elasticsearch
DEBU[0000] Boxed file                                    path=services/kibana/docker-compose.yml service=kibana
DEBU[0000] Boxed file                                    path=services/metricbeat/docker-compose.yml service=metricbeat
DEBU[0000] Boxed file                                    path=services/opbeans-go/docker-compose.yml service=opbeans-go
DEBU[0000] Boxed file                                    path=services/opbeans-java/docker-compose.yml service=opbeans-java
DEBU[0000] Boxed file                                    path=services/vsphere/docker-compose.yml service=vsphere
DEBU[0000] Boxed file                                    path=stacks/metricbeat/docker-compose.yml service=metricbeat
DEBU[0000] Workspace file                                path=/Users/mdelapenya/.op/compose/services/apache/docker-compose.yml service=apache
DEBU[0000] Workspace file                                path=/Users/mdelapenya/.op/compose/services/ceph/docker-compose.yml service=ceph
DEBU[0000] Workspace file                                path=/Users/mdelapenya/.op/compose/services/dropwizard/docker-compose.yml service=dropwizard
DEBU[0000] Workspace file                                path=/Users/mdelapenya/.op/compose/services/envoyproxy/docker-compose.yml service=envoyproxy
DEBU[0000] Workspace file                                path=/Users/mdelapenya/.op/compose/services/etcd/docker-compose.yml service=etcd
DEBU[0000] Workspace file                                path=/Users/mdelapenya/.op/compose/services/haproxy/docker-compose.yml service=haproxy
DEBU[0000] Workspace file                                path=/Users/mdelapenya/.op/compose/services/kafka/docker-compose.yml service=kafka
DEBU[0000] Workspace file                                path=/Users/mdelapenya/.op/compose/services/metricbeat/docker-compose.yml service=metricbeat
DEBU[0000] Workspace file                                path=/Users/mdelapenya/.op/compose/services/mysql/docker-compose.yml service=mysql
DEBU[0000] Workspace file                                path=/Users/mdelapenya/.op/compose/services/nats/docker-compose.yml service=nats
DEBU[0000] Workspace file                                path=/Users/mdelapenya/.op/compose/services/redis/docker-compose.yml service=redis
DEBU[0000] Workspace file                                path=/Users/mdelapenya/.op/compose/services/uwsgi/docker-compose.yml service=uwsgi
DEBU[0000] Workspace file                                path=/Users/mdelapenya/.op/compose/services/vsphere/docker-compose.yml service=vsphere
DEBU[0000] Workspace file                                path=/Users/mdelapenya/.op/compose/stacks/metricbeat/docker-compose.yml service=metricbeat
INFO[0000] Feature Context found                         modules="[stack-monitoring]" paths="[features/stack-monitoring/parity-tests.feature]"
DEBU[0000] Before StackMonitoring Suite...              
DEBU[0000] Installing elasticsearch monitoring instance 
Feature: Parity Tests
DEBU[0000] Before StackMonitoring Scenario...           
DEBU[0000] Installing elasticsearch                     
DEBU[0000] Enabling legacy collection, sending metrics to the monitoring instance 
DEBU[0000] Running elasticsearch for X seconds (default: 30) to collect monitoring data internally and index it into the Monitoring index for elasticsearch 
DEBU[0000] Stopping elasticsearch                       
DEBU[0000] Downloading sample documents from elasticsearch's monitoring index to a test directory 
DEBU[0000] Disable legacy                               

  Scenario Outline: The <product> documents indexed by the legacy collection method are identical in structure to those indexed by Metricbeat collection # features/stack-monitoring/parity-tests.feature:3
    Given "<product>" sends metrics to Elasticsearch using the "legacy" collection monitoring method                                                     # stack_monitoring_test.go:15 -> *StackMonitoringTestSuite
    When "<product>" sends metrics to Elasticsearch using the "metricbeat" collection monitoring method                                                  # stack_monitoring_test.go:15 -> *StackMonitoringTestSuite
    Then the structure of the documents for the "legacy" and "metricbeat" collection are identical                                                       # stack_monitoring_test.go:50 -> *StackMonitoringTestSuite

    Examples:
      | product       |
      | elasticsearch |
DEBU[0000] After StackMonitoring Scenario...            
DEBU[0000] Before StackMonitoring Scenario...           
DEBU[0000] Installing kibana                            
DEBU[0000] Enabling legacy collection, sending metrics to the monitoring instance 
DEBU[0000] Running kibana for X seconds (default: 30) to collect monitoring data internally and index it into the Monitoring index for kibana 
DEBU[0000] Stopping kibana                              
DEBU[0000] Downloading sample documents from kibana's monitoring index to a test directory 
DEBU[0000] Disable legacy                               
      | kibana        |
DEBU[0000] After StackMonitoring Scenario...            
DEBU[0000] Before StackMonitoring Scenario...           
DEBU[0000] Installing beats                             
DEBU[0000] Enabling legacy collection, sending metrics to the monitoring instance 
DEBU[0000] Running beats for X seconds (default: 30) to collect monitoring data internally and index it into the Monitoring index for beats 
DEBU[0000] Stopping beats                               
DEBU[0000] Downloading sample documents from beats's monitoring index to a test directory 
DEBU[0000] Disable legacy                               
      | beats         |
DEBU[0000] After StackMonitoring Scenario...            
DEBU[0000] Before StackMonitoring Scenario...           
DEBU[0000] Installing logstash                          
DEBU[0000] Enabling legacy collection, sending metrics to the monitoring instance 
DEBU[0000] Running logstash for X seconds (default: 30) to collect monitoring data internally and index it into the Monitoring index for logstash 
DEBU[0000] Stopping logstash                            
DEBU[0000] Downloading sample documents from logstash's monitoring index to a test directory 
DEBU[0000] Disable legacy                               
      | logstash      |
DEBU[0000] After StackMonitoring Scenario...            
DEBU[0000] After StackMonitoring Suite...               
DEBU[0000] Destroying elasticsearch monitoring instance, including attached services 

4 scenarios (4 pending)
12 steps (4 pending, 8 skipped)
1.28132ms
testing: warning: no tests to run
PASS
ok      github.com/elastic/metricbeat-tests-poc/e2e     0.790s
</code></pre>
</p>
</details>


e2e/runner_test.go
```diff
@@ -66,6 +66,12 @@ var supportedProducts = map[string]*contextMetadata{
 		},
 		modules: []string{"metricbeat"},
 	},
+	"parity-tests": &contextMetadata{
```
mdelapenya, line 69: For newcomers, long story about this map here: https://github.com/elastic/metricbeat-tests-poc/pull/104

e2e/elasticsearch.go
```diff
@@ -71,10 +71,7 @@ func deleteIndex(ctx context.Context, stackName string, index string) error {
 // and from them, get the one related to the Elasticsearch port (9200). As it is bound to a
 // random port at localhost, we will build the URL with the bound port at localhost.
 //nolint:unused
-func getElasticsearchClient(stackName string) (*es.Client, error) {
-	elasticsearchCfg, _ := config.GetServiceConfig("elasticsearch")
-	elasticsearchCfg.Name = elasticsearchCfg.Name + "-" + stackName
-
+func getElasticsearchClient() (*es.Client, error) {
```
mdelapenya, line 74: This is convenient not only for this PR, but for the project

e2e/elasticsearch.go
```diff
@@ -93,11 +90,11 @@ func getElasticsearchClient(stackName string) (*es.Client, error) {
 
 // maxAttempts could be redefined in the OP_QUERY_MAX_ATTEMPTS environment variable
 //nolint:unused
-func retrySearch(stackName string, indexName string, esQuery map[string]interface{}, maxAttempts int, retryTimeout int) (searchResult, error) {
+func retrySearch(indexName string, esQuery map[string]interface{}, maxAttempts int, retryTimeout int) (searchResult, error) {
```
mdelapenya, line 93: Removing the stack from here is convenient for the project, not for this PR

e2e/go.mod
```diff
@@ -3,15 +3,16 @@ module github.com/elastic/metricbeat-tests-poc/e2e
 go 1.12
 
 require (
+	github.com/Jeffail/gabs/v2 v2.5.0
```
mdelapenya, line 6: Supports generating structures from dynamic JSON on the fly

e2e/go.mod
```diff
@@ -3,15 +3,16 @@ module github.com/elastic/metricbeat-tests-poc/e2e
...
 	github.com/cenkalti/backoff/v4 v4.0.2
 	github.com/cucumber/godog v0.9.0
 	github.com/cucumber/messages-go/v10 v10.0.3
 	github.com/elastic/go-elasticsearch/v8 v8.0.0-20190731061900-ea052088db25
 	github.com/elastic/metricbeat-tests-poc/cli v1.0.0-rc1
+	github.com/nsf/jsondiff v0.0.0-20190712045011-8443391ee9b6
```
mdelapenya, line 12: Supports diffing two JSON objects

e2e/stack_monitoring_test.go
```diff
@@ -0,0 +1,614 @@
...
+
+	log.Debugf("Enabling %s collection, sending %s metrics to the monitoring instance", collectionMethod, sm.Product)
+
+	productIndexID := sm.Product
+
+	switch {
+	case sm.Product == "elasticsearch":
+		sm.Port = strconv.Itoa(9201)
+		sm.handleSpecialCases = func(docType string, legacy *gabs.Container, metricbeat *gabs.Container) error {
+			return fmt.Errorf("Not supported yet")
```
mdelapenya, line 63: Placeholder for future replacement

e2e/stack_monitoring_test.go
```diff
@@ -0,0 +1,614 @@
...
+				legacyPipelines := legacy.Path(pipelinesPath)
+				metricbeatPipelines := metricbeat.Path(pipelinesPath)
+
+				legacyPipeline := legacyPipelines.Children()[0]
+				metricbeatPipeline := metricbeatPipelines.Children()[0]
+
+				legacyVertices := legacyPipeline.Path("vertices")
+				metricbeatVertices := metricbeatPipeline.Path("vertices")
+
+				// sort by vertex ID
```
mdelapenya, line 101: Need to deal with sorting this sub-objects

e2e/stack_monitoring_test.go
```diff
@@ -0,0 +1,614 @@
...
+			env["httpPort"] = "5066"
+			env["xpackMonitoring"] = "false"
+		} else {
+			env["xpackMonitoring"] = "true"
+		}
+	default:
+		return fmt.Errorf("Product %s not supported", product)
+	}
+
+	sm.IndexName = ".monitoring-" + productIndexID + "-7"
```
mdelapenya, line 155: Hey @ycombinator, when running kibana or logstash, I do not see the related legacy monitoring index: `.monitoring-logstash`, so the program cannot query ES for the documents and die (expected). Any guess?

e2e/stack_monitoring_test.go
```diff
@@ -0,0 +1,614 @@
...
+	allowedInsertionsInMetricbeatDocs := []string{
+		"service",
+		"@timestamp",
+		"agent",
+		"event",
+		"host",
+		"ecs",
+		"metricset",
+	}
+	allowedInsertionsInMetricbeatDocs = append(allowedInsertionsInMetricbeatDocs, sm.allowedInsertionsExtra...)
```
mdelapenya, line 425: Extras are populated when checking the product

e2e/stack_monitoring_test.go
```diff
@@ -0,0 +1,614 @@
...
+
+	allowedDeletionsFromMetricbeatDocs := []string{
+		"source_node",
+	}
+	allowedDeletionsFromMetricbeatDocs = append(allowedDeletionsFromMetricbeatDocs, sm.allowedDeletionsExtra...)
```
mdelapenya, line 430: Extras are populated when checking the product

e2e/stack_monitoring_test.go
```diff
@@ -0,0 +1,614 @@
...
+
+		lines := strings.Split(differences, "\n")
+		for _, line := range lines {
+			trimmedLine := strings.TrimSpace(line)
+
+			if !strings.HasPrefix(trimmedLine, `"ADDED":`) && !strings.HasPrefix(trimmedLine, `"REMOVED":`) {
+				continue
+			}
+
+			if strings.HasPrefix(trimmedLine, `"ADDED": `) {
```
mdelapenya, line 486: ADDED is added when diffing the JSON docs. See https://github.com/elastic/metricbeat-tests-poc/pull/113/files#diff-0e8159a7feeaf000a697c0e34b10e104R585

e2e/stack_monitoring_test.go
```diff
@@ -0,0 +1,614 @@
...
+				if !strings.HasSuffix(trimmedLine, ",") {
+					continue
+				} else if checkAllowedField(trimmedLine, allowedInsertionsInMetricbeatDocs) {
+					continue
+				} else if strings.HasSuffix(trimmedLine, "{} (object)},") {
```
mdelapenya, line 491: It's curious that I find this line when checking parity:

"ADDED": {{} (object)},

e2e/stack_monitoring_test.go
```diff
@@ -0,0 +1,614 @@
...
+			if strings.HasPrefix(trimmedLine, `"REMOVED": `) {
+				if !strings.HasSuffix(trimmedLine, ",") {
+					continue
+				} else if checkAllowedField(trimmedLine, allowedDeletionsFromMetricbeatDocs) {
+					continue
+				} else if strings.HasSuffix(trimmedLine, "{} (object)},") {
+					continue
+				}
+
+				unexpectedDeletions = append(unexpectedDeletions, trimmedLine)
```
mdelapenya, line 508: I'm adding here the entire line retrieved by jsonDiff

e2e/stack_monitoring_test.go
mdelapenya: You can take a look at the generated ouput for beats here:

https://apm-ci.elastic.co/blue/organizations/jenkins/beats%2Fpoc-metricbeat-tests-poc-mbp%2FPR-113/detail/PR-113/3/pipeline#step-297-log-554

ycombinator:
Hi @mdelapenya (and maybe also @jsoriano @andresrc) , quick question about this PR. I saw the recent email about closing the modules testing working group. Just want to double check that:
a. we want to continue the investment in the E2E testing prototype via this PR, and
b. if so, we are **also** committing to running the tests being implemented by this PR in CI, so we can deprecate the existing stack monitoring parity tests currently running in CI (e.g. https://internal-ci.elastic.co/job/elastic+estf-monitoring-snapshots+master/).

In other words, just re-implementing the _existing_ parity tests in the E2E testing prototype will be wasteful, IMO, if we don't also complete the picture by running them in CI just like the existing ones are. I just want to double check that this is the plan here before spending more time on this path.

mdelapenya:
> In other words, just re-implementing the _existing_ parity tests in the E2E testing prototype will be wasteful, IMO, if we don't also complete the picture by running them in CI just like the existing parity tests. I just want to double check that this is the plan here before spending more time on this path.

Yes, in terms of stack-monitoring, once we have something that we like, the idea is to put that code in the CI and run it with the same frequency as it runs now. As you know, this repo is only run whenever there are changes on it, but we can integrate it with any PR on beats, or logstash or elasticsearch if we want.

This PR is an example, https://github.com/elastic/beats/pull/13073 (very old in time, but not in behavior): we can monitor anything we need and run this project in consequence

cli/config/compose/services/elasticsearch/docker-compose.yml
```diff
@@ -6,10 +6,9 @@ services:
...
       - xpack.security=false
-      - xpack.monitoring.collection.enabled=true
+      - xpack.monitoring.collection.enabled=${xpackMonitoringCollection:-true}
       - ELASTIC_USERNAME=elastic
       - ELASTIC_PASSWORD=p4ssw0rd
     image: "docker.elastic.co/elasticsearch/elasticsearch:${elasticsearchTag}"
     ports:
-      - "9200:9200"
-      - "9300:9300"
+      - "${elasticsearchPort:-9200}:9200"
```
ycombinator, line 14: I can't comment on line 3 (because it's not part of the diff) but WDYT about calling this service `productionES`, to mirror the other Elasticsearch service, `monitoringES`?

ycombinator: BTW, "production ES cluster" and "monitoring ES cluster" are terms that are commonly used in the context of Stack Monitoring.

cli/config/compose/stacks/metricbeat/docker-compose.yml
```diff
@@ -11,4 +11,4 @@ services:
       - ELASTIC_PASSWORD=p4ssw0rd
     image: "docker.elastic.co/elasticsearch/elasticsearch:${stackVersion}"
     ports:
-      - "9200:9200"
+      - "${elasticsearchPort:-9200}:9200"
```
ycombinator, line 14: What is the difference between this `elasticsearch` service and the other `elasticsearch` one, defined in `cli/config/compose/services/elasticsearch/docker-compose.yml`?

cli/config/compose/stacks/stack-monitoring/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+      - bootstrap.memory_lock=true
+      - discovery.type=single-node
+      - ES_JAVA_OPTS=-Xms512m -Xmx512m
+      - xpack.security.enabled=false
+      - xpack.monitoring.collection.enabled=true
+      - ELASTIC_USERNAME=elastic
+      - ELASTIC_PASSWORD=p4ssw0rd
+    image: "docker.elastic.co/elasticsearch/elasticsearch:${stackVersion}"
+    ports:
+      - "${monitoringEsPort:-9200}:9200"
```
ycombinator, line 14: Should this file be under `cli/config/compose/services/` like the other services?

e2e/configurations/metricbeat/beat-xpack.yml
```diff
@@ -0,0 +1,8 @@
+metricbeat.modules:
+- module: beat
+  metricsets:
+    - stats
+    - state
```
ycombinator, line 5: BTW, starting with Metricbeat 7.8.0, we will no longer need to specify the `metricsets` when `xpack.enabled: true` is set: https://github.com/elastic/beats/pull/17609. Since this PR here is targeting `master`, I think we should remove `metricset`. When we backport to `7.7` and `6.8` (we will need these branches in this repo, BTW, similar to how https://github.com/elastic/elastic-stack-testing has branches corresponding to product repos), we can add them back.

e2e/runner_test.go
```diff
@@ -184,3 +192,39 @@ func parseFeatureFlags(flags []string) ([]string, []*contextMetadata) {
...
+			"minutes": minutesToBeHealthy,
+			"stack":   stackName,
+		}).Error("The Elasticsearch cluster could not get the healthy status")
+		panic("The Elasticsearch cluster could not get the healthy status")
+	}
+}
+
+// tearDownRuntimeDependencies destroys the runtime dependencies for a stack,
+// not failing the execution in the case it's not possible to destroy them
+func tearDownRuntimeDependencies(stackName string) {
```
ycombinator, line 222: Should we keep `stackName` here? Looks like it has been removed from other similar functions?

e2e/runner_test.go
```diff
@@ -184,3 +192,39 @@ func parseFeatureFlags(flags []string) ([]string, []*contextMetadata) {
 
 	return featurePaths, metadatas
 }
+
+// startRuntimeDependencies spins up the runtime dependencies for a stack, represented
+// by a docker-compose file, It will panic if they cannot be satisfied
+func startRuntimeDependencies(stackName string, env map[string]string, minutesToBeHealthy time.Duration) {
```
ycombinator, line 198: Should we keep `stackName` here? Looks like it has been removed from other similar functions?

e2e/metricbeat_test.go
```diff
@@ -143,39 +142,18 @@ func MetricbeatFeatureContext(s *godog.Suite) {
...
-
-		minutesToBeHealthy := 3 * time.Minute
-		healthy, err := waitForElasticsearch(minutesToBeHealthy, "metricbeat")
-		if !healthy {
-			log.WithFields(log.Fields{
-				"error":   err,
-				"minutes": minutesToBeHealthy,
-			}).Error("The Elasticsearch cluster could not get the healthy status")
-		}
+		startRuntimeDependencies("metricbeat", env, 3)
```
ycombinator, line 150: The third argument here is a `time.Duration`. As it stands, we're using a magic number here, `3`, for this argument. The nice thing about this being a `time.Duration` is that we can instead use `3 * time.Minute` here, which makes it less of a magic number and conveys more easily what this argument is.

e2e/stack_monitoring_test.go
```diff
@@ -0,0 +1,614 @@
...
+			}).Warn("The monitoring index was not deleted, but we are not failing the test case")
+		}
+	}
+	defer fn(context.Background())
+
+	return nil
+}
+
+// StackMonitoringFeatureContext adds steps to the Godog test suite
+func StackMonitoringFeatureContext(s *godog.Suite) {
```
ycombinator, line 365: Please correct me if I'm wrong, but this function here appears to be the entry point or exported API to run the stack monitoring parity tests. Could we move this to the top of this file to reflect this more clearly?

e2e/stack_monitoring_test.go
```diff
@@ -0,0 +1,614 @@
...
+
+		// look up configurations under workspace's configurations directory
+		dir, _ := os.Getwd()
+		env[sm.Product+"ConfigFile"] = path.Join(dir, "configurations", "parity-testing", sm.Product+".yml")
+
+		env["serviceName"] = sm.Product
+
+		if collectionMethod == "metricbeat" {
+			env["httpEnabled"] = "true"
+			env["httpPort"] = "5066"
```
ycombinator, line 146: I wonder if we should move these two under the beat case in the switch above. That's because these are only really valid when Metricbeat is monitoring a Beat. They are not valid when Metricbeat is monitoring other stack products like LS, etc.

e2e/stack_monitoring_test.go
ycombinator: In the case of Logstash, to configure "legacy" collection, you need to set `xpack.monitoring.enabled: true` in the Logstash node's configuration. Since we're running Logstash on Docker here, that can be done by setting the environment variable `XPACK_MONITORING_ENABLED=true` in Logstash's `docker-compose.yml`. 

For more info, see https://www.elastic.co/guide/en/logstash/current/docker-config.html#docker-env-config and https://www.elastic.co/guide/en/logstash/current/monitoring-internal-collection.html#monitoring-general-settings.

ycombinator: Kibana is similar to Logstash. See https://www.elastic.co/guide/en/kibana/current/docker.html#environment-variable-config.

e2e/stack_monitoring_test.go
```diff
@@ -0,0 +1,614 @@
...
+	if err != nil {
+		log.WithFields(log.Fields{
+			"error":   err,
+			"service": product,
+			"version": stackVersion,
+		}).Error("Could not stop the service.")
+		return err
+	}
+
+	log.Debugf("Downloading sample documents from %s's monitoring index to a test directory", product)
```
ycombinator, line 341: Omit "to a test directory"?

e2e/stack_monitoring_test.go
```diff
@@ -0,0 +1,614 @@
...
+	log.Debugf("Deleting monitoring index %s", sm.IndexName)
+	fn := func(ctx context.Context) {
+		err := deleteIndex(ctx, sm.IndexName)
+		if err != nil {
+			log.WithFields(log.Fields{
+				"index": sm.IndexName,
+			}).Warn("The monitoring index was not deleted, but we are not failing the test case")
+		}
+	}
+	defer fn(context.Background())
```
ycombinator, line 359: Suggest moving this cleanup closure to the top of this function, or at least before `runProduct` is called since that will end up creating the indices. That way, we'll do this cleanup no matter where we return from this function.

ycombinator: Also, you could probably collapse the the closure definition and deferred call into a single statement like so:

```
defer func(ctx context.Context) {
  ... 
}(context.Background())
```

e2e/stack_monitoring_test.go
```diff
@@ -0,0 +1,614 @@
...
+		return err
+	}
+
+	sm.collectionHits[collectionMethod] = hits
+	log.Debugf("Hits: %v", hits)
+
+	log.Debugf("Deleting monitoring index %s", sm.IndexName)
+	fn := func(ctx context.Context) {
+		err := deleteIndex(ctx, sm.IndexName)
+		if err != nil {
```
ycombinator, line 353: Nit: since `deleteIndex` just returns an error, you could collapse these two statements into:

```
if deleteIndex(...) != nil {
   ...
}
```

or, alternatively if you prefer having the `err` around:

```
if err := deleteIndex(...); err != nil {
```

e2e/stack_monitoring_test.go
```diff
@@ -0,0 +1,614 @@
...
+	legacyHits := legacyContainer.Path(hitsPath)
+	metricbeatHits := metricbeatContainer.Path(hitsPath)
+
+	legacyTypes, legacySources := checkSourceTypes(legacyHits)
+	metricbeatTypes, metricbeatSources := checkSourceTypes(metricbeatHits)
+
+	if len(legacyTypes) > len(metricbeatTypes) {
+		// returns an array as the result of removing all elements in 'b' from the 'a' array
+		// only used here
+		arrayDiff := func(a []string, b []string) []string {
```
ycombinator, line 442: Consider moving this out to a top-level function in this file so you're not re-defining it every time `checkParity` is called.

e2e/stack_monitoring_test.go
```diff
@@ -0,0 +1,614 @@
...
+			target := map[string]bool{}
```
ycombinator, line 443: Fun fact: `bool` takes more memory (1 byte) than an empty struct (0 bytes)! See https://dave.cheney.net/2014/03/25/the-empty-struct.

So when using a map for this sort of "existence-check" pattern, where we don't really care about the value of the map entries, consider using `map[string]struct{}`.

e2e/stack_monitoring_test.go
```diff
@@ -0,0 +1,614 @@
...
+	diffJSON, differences := jsondiff.Compare(a.Bytes(), b.Bytes(), &opt)
+	if diffJSON != jsondiff.FullMatch {
+		log.WithFields(log.Fields{
+			"diff":     differences,
+			"dest":     a.String(),
+			"diffType": diffJSON,
+			"source":   b.String(),
+		}).Debug("There are differences in the json")
+
+		fmt.Println(differences)
```
ycombinator, line 609: Should we move this to `log.Debug`?

e2e/stack_monitoring_test.go
```diff
@@ -0,0 +1,614 @@
...
+		unexpectedDeletions := []string{}
+
+		differences := jsonDiff(legacyDoc, metricbeatDoc)
+
+		lines := strings.Split(differences, "\n")
+		for _, line := range lines {
+			trimmedLine := strings.TrimSpace(line)
+
+			if !strings.HasPrefix(trimmedLine, `"ADDED":`) && !strings.HasPrefix(trimmedLine, `"REMOVED":`) {
+				continue
```
ycombinator, line 483: For each of these `continue` cases, it might be helpful to have a comment _why_ we're OK to continue in each case.

e2e/stack_monitoring_test.go
```diff
@@ -0,0 +1,614 @@
...
+
+				unexpectedDeletions = append(unexpectedDeletions, trimmedLine)
+				continue
+			}
+		}
+
+		if len(unexpectedInsertions) == 0 && len(unexpectedDeletions) == 0 {
+			log.WithFields(log.Fields{
+				"docType": docType,
+			}).Debugf("Expected parity between Metricbeat-indexed doc and legacy-indexed doc found.")
```
ycombinator, line 516: It's nice to know when the tests pass. How about we move this to INFO level and also mention the product name and `docType`?

e2e/stack_monitoring_test.go
```diff
@@ -0,0 +1,614 @@
...
+					"product":  sm.Product,
+				}).Warn(err.Error())
+				errors = append(errors, err)
+			}
+		}
+	}
+	log.Debugf("Found %d errors", len(errors))
+
+	if len(errors) > 0 {
+		return fmt.Errorf("")
```
ycombinator, line 547: I assume you intended to flesh this out? 🙂

e2e/stack_monitoring_test.go
```diff
@@ -0,0 +1,614 @@
...
+	}
+
+	return nil
+}
+
+func (sm *StackMonitoringTestSuite) runProduct(product string, collectionMethod string) error {
+	env := map[string]string{
+		"logLevel":       log.GetLevel().String(),
+		product + "Tag":  stackVersion, // all products follow stack version
+		product + "Port": sm.Port,      // we could run the service in another port
```
ycombinator, line 269: Are these two env vars being used anywhere?

e2e/stack_monitoring_test.go
```diff
@@ -0,0 +1,614 @@
...
+		"monitoringEsPort": "9200",         // monitoring elasticsearch port
+	}
+
+	log.Debugf("Enabling %s collection, sending %s metrics to the monitoring instance", collectionMethod, sm.Product)
+
+	productIndexID := sm.Product
+
+	switch {
+	case sm.Product == "elasticsearch":
+		sm.Port = strconv.Itoa(9201)
```
ycombinator, line 61: What is the rationale behind setting the port here (and same question for other products below as well)?

cli/config/compose/stacks/metricbeat/docker-compose.yml
mdelapenya: There are two abstractions:
- stacks, which is a preconfigured set of services that are the runtime dependencies for a specific project, like the metricbeat tests, or the stack-monitoring parity tests.
- services, which are isolated services that can be aggregated to running services or stacks, one by one.

The metricbeat stack only contains one runtime dependency, which is an ES instance, but it could involve a Kibana for UI verifications.

Another idea of stacks for stacks is to share configurations across teams

cli/config/compose/stacks/stack-monitoring/docker-compose.yml
mdelapenya: No, because it represents the dependent services for the stack-monitoring tests. As I mentioned above, this is a stack.

e2e/configurations/metricbeat/beat-xpack.yml
mdelapenya: We could pair the branches to stack branches too, and update files in consequence

e2e/runner_test.go
mdelapenya: Yes, we should keep it to be able to execute docker-compose down to the proper compose file.

We could think about having some state to simply use the compose we initially ran, maybe in a follow-up?

e2e/runner_test.go
mdelapenya: I think this cannot be removed, as we need to start the proper runtime dependencies for the test suite, in this case the stack-monitoring tests. As this method belongs to the test runner, it's an abstraction to be reused with different test suites (metricbeat, stack-mon...), so it needs to know which stack to run

e2e/metricbeat_test.go
mdelapenya: Love this suggestion! In fact I had in mind to add an env var to be able to modify that at CI time

e2e/stack_monitoring_test.go
mdelapenya: Agreed.

How I organise the code is: structs to the top, then public struct methods in alphabetical order, then private struct methods in alphabetical order, then other functions in alphabetical order

e2e/stack_monitoring_test.go
mdelapenya: I think they are under the `strings.HasSuffix(sm.Product, "beat"):` scope, right?

e2e/stack_monitoring_test.go
mdelapenya: Agreed, I initially thought that, although I added the function there inside the `if` branch to be created only if the length is different, which ideally will happen rarely, versus having it always there. I'll move it out

e2e/stack_monitoring_test.go
mdelapenya: Sure thing, it maybe was missed

e2e/stack_monitoring_test.go
mdelapenya: This was one of the initial things I did when started this PR, and thinking about it now, it makes no sense. I'll remove them until needed

mdelapenya: BTW, I realised that for the Elasticsearch test, the port it's needed, as there is an env variable (product + "Port": sm.Port) that binds the port to the docker host, and that port (9200) is already in use by the elasticsearch stack instance

mdelapenya:
Hey @ycombinator, thanks for you feedback, I've addressed it in the last commit.

What I'm concerned about is that for elasticsearch, kibana and logstash, the monitoring indices are not present:

```log
DEBU[0072] Elasticsearch query                           index=.monitoring-elasticsearch-7-2020.05.12 query="map[collapse:map[field:type] sort:map[timestamp:asc]]"
ERRO[0072] Could not send query to Elasticsearch in the specified time (15 seconds)  error="Could not send query to Elasticsearch in the specified time (15 seconds)" index=.monitoring-elasticsearch-7-2020.05.12 query="map[collapse:map[field:type] sort:map[timestamp:asc]]" retryAttempts=5 retryTimeout=3
DEBU[0072] Deleting monitoring index .monitoring-elasticsearch-7-2020.05.12 
DEBU[0072] Index deleted using Elasticsearch Go client   indexName=.monitoring-elasticsearch-7-2020.05.12 status="404 Not Found"

...

DEBU[0123] Elasticsearch query                           index=.monitoring-kibana-7-2020.05.12 query="map[collapse:map[field:type] sort:map[timestamp:asc]]"
ERRO[0123] Could not send query to Elasticsearch in the specified time (15 seconds)  error="Could not send query to Elasticsearch in the specified time (15 seconds)" index=.monitoring-kibana-7-2020.05.12 query="map[collapse:map[field:type] sort:map[timestamp:asc]]" retryAttempts=5 retryTimeout=3
DEBU[0123] Deleting monitoring index .monitoring-kibana-7-2020.05.12 
DEBU[0123] Index deleted using Elasticsearch Go client   indexName=.monitoring-kibana-7-2020.05.12 status="404 Not Found"

...

DEBU[0169] Elasticsearch query                           index=.monitoring-logstash-7-2020.05.12 query="map[collapse:map[field:type] sort:map[timestamp:asc]]"
ERRO[0169] Could not send query to Elasticsearch in the specified time (15 seconds)  error="Could not send query to Elasticsearch in the specified time (15 seconds)" index=.monitoring-logstash-7-2020.05.12 query="map[collapse:map[field:type] sort:map[timestamp:asc]]" retryAttempts=5 retryTimeout=3
DEBU[0169] Deleting monitoring index .monitoring-logstash-7-2020.05.12 
DEBU[0169] Index deleted using Elasticsearch Go client   indexName=.monitoring-logstash-7-2020.05.12 status="404 Not Found"
```

cli/config/compose/services/elasticsearch/docker-compose.yml
mdelapenya: I'm afraid that the service name is not configurable in runtime, so the `elasticsearch` name is hardcoded.

ycombinator: Gotcha, so just so I'm understanding this right: `elasticsearch` here refers to the "production" ES cluster, right?

cli/config/compose/stacks/metricbeat/docker-compose.yml
ycombinator: Ah, I didn't realize this. Thanks for explaining!

e2e/configurations/metricbeat/beat-xpack.yml
ycombinator: That would be great. Not something we need to do just yet, but just a thought for the future when we start seeing differences between branches (like the one I mentioned above).

cli/config/compose/services/elasticsearch/docker-compose.yml
mdelapenya: Yes: anything under `services` is reusable to be aggregated to a running compose on the fly, that's why they are generic

mdelapenya: On the other hand, I'm seeing a potential problem if somebody wants to add two services of the same kind in an already running compose: service names would collide 🤔

e2e/configurations/metricbeat/beat-xpack.yml
mdelapenya: Opened an issue here: https://github.com/elastic/e2e-testing/issues/119

e2e/stack_monitoring_test.go
mdelapenya: Added in latest commit

e2e/stack_monitoring_test.go
```diff
@@ -466,54 +477,47 @@ func checkParity(sm *StackMonitoringTestSuite, legacyContainer *gabs.Container,
...
 
-		differences := jsonDiff(legacyDoc, metricbeatDoc)
-
-		lines := strings.Split(differences, "\n")
-		for _, line := range lines {
-			trimmedLine := strings.TrimSpace(line)
+		// Flatten a JSON array or object into an object of key/value pairs for each
+		// field, where the key is the full path of the structured field in dot path
+		// notation matching the spec for the method Path.
+		flatLegacy, err := legacyDoc.Flatten()
```
mdelapenya, line 483: This is the magic I was looking for to check for existence!

e2e/stack_monitoring_test.go
mdelapenya: @ycombinator I did progress with kibana (tests passing for Kibana and Filebeat). But not sure what to dod here with the vertices. In my execution, MB vertices are null, and they are retrieved as unexpected deletions. Is this correct?

e2e/stack_monitoring_test.go
```diff
@@ -462,8 +462,15 @@ func (sm *StackMonitoringTestSuite) runProduct(product string, collectionMethod
 		return err
 	}
 
+	if sm.Product == "elasticsearch" {
```
mdelapenya, line 465: This is using Elasticsearch client waiting for the cluster to be healthy

e2e/testresources/elasticsearch-legacy-monitoring.json
```diff
@@ -0,0 +1,1344 @@
+{
```
mdelapenya, line 1: These JSON files are extracted from here:
- https://internal-ci.elastic.co/job/elastic+estf-monitoring-snapshots+7.7+multijob-beat/50/consoleText
- https://internal-ci.elastic.co/job/elastic+estf-monitoring-snapshots+7.7+multijob-elasticsearch/50/consoleText
- https://internal-ci.elastic.co/job/elastic+estf-monitoring-snapshots+7.7+multijob-kibana/50/consoleText

And they are very useful for debugging the helper functions to handle special cases.
- https://internal-ci.elastic.co/job/elastic+estf-monitoring-snapshots+7.7+multijob-logstash/50/consoleText

e2e/stack_monitoring_checks_test.go
```diff
@@ -0,0 +1,406 @@
+package e2e
```
mdelapenya, line 1: Running `go test -v ./...` from the e2e dir, or using VSCode to debug the tests helped me to debug and improve the helper functions

ycombinator:
Hi @mdelapenya, I pulled down this PR and ran it. I focussed just on the Elasticsearch parity tests, not the other products. I did this because these tests are failing for me (and I believe for you as well) with this error:

```
        Found more legacy-indexed document types than metricbeat-indexed document types. Document types indexed by legacy collection but not by Metricbeat collection: [index_stats indices_stats shards]
``` 

As you know, the setup is as follows:
- The monitoring ES cluster runs on `localhost:9200`.
- The production ES cluster runs on `localhost:9201`.

At the end of the legacy collection stage, we expect monitoring documents to end up in `.monitoring-es-*` indices on the monitoring cluster. I was examining these documents and I noticed something wrong. The documents have a `cluster_uuid` field which **should** have the cluster UUID of the **production** ES cluster. Instead they contained the cluster UUID of the **monitoring** ES cluster. 

You can check this as follows:
1. Grab the cluster UUID of the monitoring ES cluster: 
   ```
   curl -s 'http://localhost:9200/?filter_path=cluster_uuid' | grep cluster_uuid
   ```
2. Grab the cluster UUID of the production ES cluster:
   ```
   curl -s 'http://localhost:9201/?filter_path=cluster_uuid' | grep cluster_uuid
   ```
3. Check the value of the `cluster_uuid` field in various documents in `.monitoring-es-*` indices on the monitoring ES cluster:
   ```
   curl -s 'http://localhost:9200/.monitoring-es-7-*/_search?filter_path=**._source.cluster_uuid&size=1000&pretty' | grep cluster_uuid | sort -u
   ```

To me, this indicates that the monitoring ES cluster is currently monitoring itself via legacy collection, instead of the production cluster monitoring itself via legacy collection and shipping it's data to the monitoring ES cluster.  So something is going wrong in how we are setting up the monitoring ES cluster and the production ES cluster both for legacy collection.

ycombinator:
@mdelapenya Thinking about this some more, I think we are complicating the test implementation more than we need to. The issue that led to this PR is https://github.com/elastic/e2e-testing/issues/94. As I expressed in that issue, I really just want us to start by porting over **one** product's parity tests. Porting over, to me, also means keeping more or less the same steps as being done by the current parity tests in the new one here in the E2E testing framework. We can always make things more sophisticated later.

For instance, here are two examples of additional complexity we are taking on in this PR that I think we can avoid for now:

1. The current parity tests only spin up a single Elasticsearch cluster. It acts as the production and monitoring cluster both. The new ones in this PR are trying to spin up two Elasticsearch clusters, one to act as the production one and one to act as the monitoring one. Let's not tackle this complexity just yet. Let's simplify for now and just do what the existing parity tests are doing.

2. Another thing is that the current parity tests setup monitoring settings in Elasticsearch via API calls. In contrast, the new ones in this PR are trying to do this statically via environment variables passed to the docker containers. In theory this difference should not matter but I'd rather not introduce differences while we're porting over the tests. We can make these changes later if we believe they are necessary.

In short, I would really like the **initial version** of the new parity tests in the E2E framework to **just mimic exactly** the [steps](https://github.com/elastic/elastic-stack-testing/blob/master/playbooks/monitoring/elasticsearch/docs_parity.yml#L44) being done by the current parity tests. And let's just focus on **one product: Elasticsearch** for now as well.  Doing both these things will make debugging a whole lot easier IMO. Once we have a working PR, we can then move on making things more sophisticated in the future (like separating out the monitoring cluster from the production cluster).

mdelapenya:
> To me, this indicates that the monitoring ES cluster is currently monitoring itself via legacy collection, instead of the production cluster monitoring itself via legacy collection and shipping it's data to the monitoring ES cluster. So something is going wrong in how we are setting up the monitoring ES cluster and the production ES cluster both for legacy collection.

Ok, I'm going to check the configuration of both services, as this seems a reasonable issue for the observed documents. Thanks fr looking into this

mdelapenya:
> @mdelapenya Thinking about this some more, I think we are complicating the test implementation more than we need to. The issue that led to this PR is #94. As I expressed in that issue, I really just want us to start by porting over **one** product's parity tests. Porting over, to me, also means keeping more or less the same steps as being done by the current parity tests in the new one here in the E2E testing framework. We can always make things more sophisticated later.
> 
> For instance, here are two examples of additional complexity we are taking on in this PR that I think we can avoid for now:
> 
> 1. The current parity tests only spin up a single Elasticsearch cluster. It acts as the production and monitoring cluster both. The new ones in this PR are trying to spin up two Elasticsearch clusters, one to act as the production one and one to act as the monitoring one. Let's not tackle this complexity just yet. Let's simplify for now and just do what the existing parity tests are doing.

When reading about monitoring, specially here https://www.elastic.co/guide/en/elasticsearch/reference/7.6/monitoring-overview.html, I thought that having production and monitoring instances, although a bit more complex, would be better in terms of accessing the proper documents. And because one of the main benefits of this framework is to leverage Docker and Docker Compose, spinning up two different containers (i.e. monitoring and production instances) is super simple. That's why I decided to go with this implementation, as the framework gives it us it for free.

> 2. Another thing is that the current parity tests setup monitoring settings in Elasticsearch via API calls. In contrast, the new ones in this PR are trying to do this statically via environment variables passed to the docker containers. In theory this difference should not matter but I'd rather not introduce differences while we're also porting over the tests (fewer changes = fewer chances in something going wrong). We can make these changes later if we believe they are necessary.

I see here the cows Vs pets analogy: having a long-living instance that is modified in runtime to behave differently, versus creating and destroying two types of instances each time. Besides that, implementing the API calls to reconfigure the instances in runtime would have come with the addition of new code that, although relatively simple, is not present in the current codebase. Also the existing metricbeat tests are spinning up dependencies in this manner, which was on of the design principles when building this framework. Thanks to this design having two different containers is super easy to achieve because of Docker and Docker Compose, and that's the reason I'd prefer going this way.

> In short, I would really like the **initial version** of the new parity tests in the E2E framework to **just mimic exactly** the [steps](https://github.com/elastic/elastic-stack-testing/blob/master/playbooks/monitoring/elasticsearch/docs_parity.yml#L44) being done by the current parity tests. And let's just focus on **one product: Elasticsearch** for now as well. Doing both these things will make debugging a whole lot easier IMO. Once we have a working PR, we can then move on making things more sophisticated in the future (like separating out the monitoring cluster from the production cluster).

I can create a new PR removing everything not related to the Elasticsearch product, having in mind that the framework gives the execution of the other products for free (only the runtime configuration of each product and their `handle_special_cases` functions and unit tests are specific to other products). If that simplifies the review, adding the code for the other products (kibana, logstash, and filebeat) in a follow-up PR, then I'm OK with it. I simply added them here because the direct translation of the code from python to the framework was piece of cake, a quick win that I thought it wouldn't come with too much over-work. Nevertheless, I agree that we can add them in a follow-up PR.

e2e/testresources/kibana-legacy-monitoring.json
```diff
@@ -0,0 +1,566 @@
...
+                        "requests": {
+                            "disconnects": 0,
+                            "total": 0
+                        },
+                        "response_times": {
+                            "average": 0,
+                            "max": 0
+                        },
+                        "timestamp": "2020-05-21T07:59:15.934Z",
+                        "usage": {
```
mdelapenya, line 61: @ycombinator I see differences in the kibana parity checks for `kibana_stats`:
```
[2020-05-25T18:35:54.294Z] time="2020-05-25T18:35:49Z" level=warning msg="Metricbeat-indexed doc for type='kibana_stats' has unexpected insertion" docType=kibana_stats insertion=kibana_stats.usage.uptime.last_24_hours.hits.no_of_unique_observer_locations product=kibana

[2020-05-25T18:35:54.294Z] time="2020-05-25T18:35:49Z" level=warning msg="Metricbeat-indexed doc for type='kibana_stats' has unexpected insertion" docType=kibana_stats insertion=kibana_stats.usage.uptime.last_24_hours.hits.monitor_name_stats.max_length product=kibana

[2020-05-25T18:35:54.294Z] time="2020-05-25T18:35:49Z" level=warning msg="Metricbeat-indexed doc for type='kibana_stats' has unexpected insertion" docType=kibana_stats insertion=kibana_stats.usage.uptime.last_24_hours.hits.observer_location_name_stats.max_length product=kibana

[2020-05-25T18:35:54.294Z] time="2020-05-25T18:35:49Z" level=warning msg="Metricbeat-indexed doc for type='kibana_stats' has unexpected insertion" docType=kibana_stats insertion=kibana_stats.usage.uptime.last_24_hours.hits.no_of_unique_monitors product=kibana

[2020-05-25T18:35:54.294Z] time="2020-05-25T18:35:49Z" level=warning msg="Metricbeat-indexed doc for type='kibana_stats' has unexpected insertion" docType=kibana_stats insertion=kibana_stats.usage.uptime.last_24_hours.hits.observer_location_name_stats.avg_length product=kibana

[2020-05-25T18:35:54.294Z] time="2020-05-25T18:35:49Z" level=warning msg="Metricbeat-indexed doc for type='kibana_stats' has unexpected insertion" docType=kibana_stats insertion=kibana_stats.usage.uptime.last_24_hours.hits.monitor_name_stats.min_length product=kibana

[2020-05-25T18:35:54.295Z] time="2020-05-25T18:35:49Z" level=warning msg="Metricbeat-indexed doc for type='kibana_stats' has unexpected insertion" docType=kibana_stats insertion=kibana_stats.usage.uptime.last_24_hours.hits.observer_location_name_stats.min_length product=kibana

[2020-05-25T18:35:54.295Z] time="2020-05-25T18:35:49Z" level=warning msg="Metricbeat-indexed doc for type='kibana_stats' has unexpected insertion" docType=kibana_stats insertion=kibana_stats.usage.uptime.last_24_hours.hits.monitor_name_stats.avg_length product=kibana

[2020-05-25T18:35:54.295Z] time="2020-05-25T18:35:49Z" level=warning msg="Metricbeat-indexed doc for type='kibana_stats' has unexpected insertion" docType=kibana_stats insertion=kibana_stats.usage.uptime.last_24_hours.hits.autoRefreshEnabled product=kibana

[2020-05-25T18:35:54.295Z] time="2020-05-25T18:35:49Z" level=info msg="Expected parity between Metricbeat-indexed doc and legacy-indexed doc found." 
```
Please see here: https://apm-ci.elastic.co/blue/organizations/jenkins/stack%2Fe2e-testing-mbp/detail/PR-113/28/pipeline#step-296-log-519

ycombinator: >  let's just focus on one product: Elasticsearch for now

>> Nevertheless, I agree that we can add them in a follow-up PR.

Again, let's please remove Kibana and other products (except ES) from this PR. Let's tighten the scope of this PR so we can get the ES stack monitoring parity tests ported over, passing, and this PR merged. Then we can add the other products' tests in follow up PRs.

mdelapenya: I've removed any other product apart from Elasticsearch (I guess we will simply revert that commit in a very short time 😄)

Besides that, I found 5 new insertions and 2 deletions, which I added in a separate commit (https://github.com/elastic/e2e-testing/pull/113/commits/94ccffd43dcff69ab0b945a914463ee28407dfe2), so you can investigate where they come from.

Now the ES tests are passing 🎉 

I'm happy we did this, as I was able to refactor the code a bit and also add unit tests using the JSON files from the existing Ansible tests. Those unit tests are executing all functions handling special cases, creating a safety net for future refactors.

Please let me know if I can help with the review, and thanks for your guidance!

ycombinator:
Thanks @mdelapenya. As stated in #94, the goal of this PR was:

> mainly to see if there are any performance or maintainability gains in the process [of porting over the current parity tests to the E2E framework]. 

Now that this PR is only dealing with Elasticsearch parity tests, I was able to do an apples-to-apples comparison by running the parity tests as implemented by this PR and comparing the results with the existing parity tests.

### This PR

```
$ time OP_LOG_LEVEL=INFO go test -v --godog.format junit parity-tests
...
PASS
ok      github.com/elastic/e2e-testing/e2e      274.351s
```

### [Existing parity tests](https://github.com/elastic/elastic-stack-testing/tree/master/playbooks/monitoring#running-the-tests)

```
$ time AIT_STACK_PRODUCT=elasticsearch ./playbooks/monitoring/buildenv.sh
...
PLAY RECAP *********************************************************************
aithost                    : ok=157  changed=61   unreachable=0    failed=0    skipped=37   rescued=0    ignored=0

[2020-05-26 16:59:04] [INFO] Run script: /Users/shaunak/development/github/elastic-stack-testing/scripts/shell/vagrant_vm.sh
mkdir: /Users/shaunak/development/github/elastic-stack-testing/ait_workspace/7-7-1-6d314aa6_os: File exists
[2020-05-26 16:59:04] [INFO] Using Vagrantfile: /Users/shaunak/development/github/elastic-stack-testing/ait_workspace/7-7-1-6d314aa6_os/Vagrantfile
[2020-05-26 16:59:04] [WARNING] No vagrant vm action specified
[2020-05-26 16:59:04] [INFO] Deactivate python venv
AIT_STACK_PRODUCT=elasticsearch ./playbooks/monitoring/buildenv.sh  52.54s user 20.88s system 26% cpu 4:32.60 total
```

So as you can see the existing parity tests took 4 minutes and 32 seconds while the ones in this PR took 274s == 4 minutes and 33 seconds — almost exactly the same amount of time!

Given this, we can conclude that there doesn't seem to be any major performance benefit in porting over the existing parity tests to the E2E framework.

However, performance was only one of the goals stated in #94. The other one was maintainability. Given that the @elastic/stack-monitoring-ui team is currently maintaining these parity tests, I will defer to them on whether they think porting them over to the E2E framework (as illustrated in this PR) would be more maintainable for them in the long run over the current implementation.

mdelapenya:
> However, performance was only one of the goals stated in #94. The other one was maintainability. Given that the @elastic/stack-monitoring-ui team is currently maintaining these parity tests, I will defer to them on whether they think porting them over to the E2E framework (as illustrated in this PR) would be more maintainable for them in the long run over the current implementation.

About performance, I'd run benchmarks for both implementations to verify it in a controlled environment to avoid getting simplistic results.

As an example, in my local machine (Macbook 13'', 16GB RAM, 2,7 GHz Intel Core i7, SSD) it takes:
```
ok      github.com/elastic/e2e-testing/e2e      195.666s
OP_LOG_LEVEL=INFO go test -v --godog.format junit parity-tests  6.06s user 2.75s system 4% cpu 3:18.55 total
```
which is faster than the above numbers (curious about the CPU usage: 4% vs 26%).

As a side note, the tests for the 4 products (elasticsearch, kibana, logstash and filebeat) complete  fairly fast:
```
FAIL    github.com/elastic/e2e-testing/e2e      576.602s
OP_LOG_LEVEL=INFO go test -v --godog.format junit parity-tests  14.35s user 4.28s system 3% cpu 9:38.84 total
```

Talking about **maintainability**, I'm going to list the things I (and the software industry via ISO's) considered important when working in both projects, creating this framework and reading the current parity tests. Hope I don't get anybody bored here 😄 

### Testability
- This framework:
   - Unit testing possible using programming language primitives (go test)
   - Language compiler supporting type checking, and code style
   - JSON processing relying on a third-party library, plus a few helper methods
   - Test Life cycle hooks (befores, afters) provided by the test libraries, being possible to create and destroy test resources using them.
- Current parity tests:
   - Linting of YAML difficult (indentations could be hell)
   - Unit tests for Ansible possible (i.e. [molecule](https://molecule.readthedocs.io/en/latest/)), although not present
   - Native JSON support in Python is great

### Analyzability
- This framework:
   - Golang as sole language for control flow
   - Golang code is organised in one single package, with files per functionality
   - Configuration variables defined in code, being possible to override them in the shell (i.e. OP_ variables).
- Current parity tests:
   - Ansible for control flow
   - YAML for Ansible role and playbook definition as main language
   - Ansible layout includes roles, playbooks, dependant roles
   - Configuration variables defined at Ansible level in multiple places: group_vars, host_vars, each role, playbook, shell (i.e. AIT_ variables).

### Functionality
- This framework:
   - Suitability:
     - It uses a high level, general purpose programming language (Golang), which seems more appropriate to create assertions
- Current parity tests:
   - Suitability:
     - It uses a declarative language (YAML) and Ansible, which uses YAML to define the state of the software to be installed in a target machine. It's possible to create assertions in YAML files.

### Usability
- This framework:
   - Learnability:
     - Golang is relatively simple to learn and understand, so any developer coming from a backend language would find most of language capabilities present in the language.
- Current parity tests:
   - Learnability:
     - YAML as declarative syntax is easy to learn, although difficult to maintain

### Installability
- This framework needs:
   - Golang 1.13
   - Docker
   - Docker Compose
- Current parity tests needs:
   - Python 3.6 (it does not run locally on my Mac with Python 3.7)
   - Virtualenv
   - Ansible
   - Vagrant
   - VirtualBox
   - Bash (to run shell scripts)

ycombinator:
> About performance, I'd run benchmarks for both implementations to verify it in a controlled environment to avoid getting simplistic results.

Isn't that what I did? Can you explain why your environment should be considered more controlled than mine?

Also, your results prove that the new parity tests run faster in your environment than mine. But the question is about the relative performance between the existing and new parity tests, right?

In my comment I ran both implementations (existing and new) and posted both their results. However, so far you have only posted the results of the new parity tests in your comment. For an apples to apples comparison, shouldn't you also run the existing parity tests in your environment and post those results as well?

As for maintainability, thanks for listing out the various factors that go into it. We should indeed consider these in deciding which implementation is more maintainable long term. I have no horse in this race as I'm no longer the primary maintainer of these tests. So I'll reiterate what I said in my previous comment: it's up to the @elastic/stack-monitoring-ui team to weigh in on this aspect as they are primarily maintaining the parity tests.

I think it's worth remembering that we already have stack monitoring parity tests implemented and running in CI for well over a year at this point. They're stable and have been catching bugs for us already. We made an investment in them to get them to be stable and useful, which took a few months, and since then that investment has been paying off with very little upkeep. The framework they are built on is being actively maintained and is well supported. So there is no _inherent_ reason to abandon this investment now. 

Essentially, we've paid off the costs and are reaping the benefits. Whereas with a new implementation we're going to have to put in some work first before we start seeing the same level of stability and usefulness. For example, right now my involvement in maintaining the existing parity tests is pretty much zero. By comparison, I've had to be quite involved in helping debug the new implementation and I suspect I would be for at least some more time.

Basically, in my mind, there has to be enough of a benefit to move away from the existing implementation that's stable, useful, and well-supported; otherwise I'm not sure it's worth it.

chrisronline:
I wasn't really aware of this effort, but I'm having a hard time understanding why this is happening. I don't think we (the stack monitoring team) has any issues with these tests as is.

Is there any background I can read about this effort?

mdelapenya:
Sorry if I expressed that I ran the tests in a controlled environment, but my intention was to clarify that none of mine or yours is controlled 😊 I only posted the results as a pure demonstration of non being controlled. I can picture a bare-metal, dedicate machine running on CI for different runs to accomplish this performance task.

Besides that, I tried to run the existing ones without success, that's why I added the `it does not run locally on my Mac with Python 3.7` comment. I'd like to run them today, after installing all dependant tools and versions.

Please do not get me wrong, I'm not saying there is no value in the parity tests so we have to replace them. I totally value what they provide to detect changes between both JSON schemas. The one-month work in this PR is trying to address the initial request to check if this new dockerised, BDD-based framework was able to mimic what the parity tests do, and that's what we did: bring them to the framework, where we would be able to run it with the frequency that we need (with every PR or merge, daily, etc.)

TBH I was simply surprised that measuring performance in a few runs would determine the performance mark for both use cases, that's why I brought in the maintainability aspects to consider when taking a final decision.

mdelapenya:
I've not mentioned it, but any initiative from our team (Eng. Productivity) is trying to assist where needed: if we see no improvements in this new approach, then it's totally OK 😊 so closing this PR would be fine too

ycombinator:
> Sorry if I expressed that I ran the tests in a controlled environment, but my intention was to clarify that none of mine or yours is controlled 😊 I only posted the results as a pure demonstration of non being controlled. I can picture a bare-metal, dedicate machine running on CI for different runs to accomplish this performance task.

Ah, thanks for clarifying. I agree with you that a controlled environment is ideal for benchmarking. Given that we're more interested in the relative comparison between two implementations I think, as long as they run in the _same_ environment, it's good enough — not perfect, but good enough IMO.

BTW, it is for this exact reason that I avoided comparing the performance results of the [existing](https://internal-ci.elastic.co/job/elastic+estf-monitoring-snapshots+7.7+multijob-elasticsearch/56) and [new](https://apm-ci.elastic.co/blue/rest/organizations/jenkins/pipelines/stack/pipelines/e2e-testing-mbp/branches/PR-113/runs/33/nodes/209/log/?start=0) implementations in a CI environment — I would consider a CI environment to be even less controlled than our local machines. So I'm right there with you on how the environment being controlled vs. not can influence the numbers.

> I tried to run the existing ones without success, that's why I added the it does not run locally on my Mac with Python 3.7 comment. I'd like to run them today, after installing all dependant tools and versions.

👍 Please do. We have one comparative performance analysis done on my system. It would be great to have another one done on yours as well. 

> The one-month work in this PR is trying to address the initial request (#94) to check if this new dockerised, BDD-based framework was able to mimic what the parity tests do, and that's what we did.

Indeed, and I'm grateful for the work you've done in bringing this PR to where it is. Without it we couldn't be having the discussion we're having right now about performance and maintainability, which is what #94 was about. Quoting from it:

> It would be great to try and port over one of these product's parity tests over to the e2e-testing framework, mainly to see if there are any performance or maintainability gains in the process.

> TBH I was simply surprised that measuring performance in a few runs would determine the performance mark for both use cases, that's why I brought in the maintainability aspects to consider when taking a final decision.

This is a fair criticism. I will re-run both implementations locally multiple times (resetting state by removing all docker images and VMs+images and making sure the CPU usage drops to a low stable rate each time). I'll post the results of each run here. If you do the same in your environment and post the results, hopefully we can get a fairer picture of relative performance.

ycombinator:
> I wasn't really aware of this effort, but I'm having a hard time understanding why this is happening. I don't think we (the stack monitoring team) has any issues with these tests as is.

> Is there any background I can read about this effort?

@chrisronline This effort is an exploration of whether there might be benefits (performance and maintainability, specifically) to migrating the existing parity tests over to the E2E testing framework being developed by the Observability Engineering Productivity team. I requested this exploration in #94. My thinking was/is that if there are significant benefits in either or both areas (performance and/or maintainability), it would be worth moving the tests over. As the primary maintainer of the existing tests along with @igoristic, WDYT?

mdelapenya:
@ycombinator I have found some errors while preparing the environment, so I created this issue in the repo: https://github.com/elastic/elastic-stack-testing/issues/585

chrisronline:
It sounds like the ask here is for myself and @igoristic to pull this down and run the tests and get a sense of how they work and how we can debug them. We can do that and report back

ycombinator:
@chrisronline Yes, that would be an ideal way to get a sense of maintainability. Thanks!

mdelapenya:
Hey @ycombinator, I managed to run the parity test locally. These are the time results for the parity tests:

1st run:
```
AIT_STACK_PRODUCT=elasticsearch ./playbooks/monitoring/buildenv.sh  58.07s user 87.49s system 48% cpu 5:00.84 total
```
2nd run:
```
AIT_STACK_PRODUCT=elasticsearch ./playbooks/monitoring/buildenv.sh  55.85s user 84.51s system 48% cpu 4:49.31 total
```
3rd run:
```
AIT_STACK_PRODUCT=elasticsearch ./playbooks/monitoring/buildenv.sh  55.13s user 82.31s system 48% cpu 4:45.46 total
```

ycombinator:
Thanks @mdelapenya. Those are about the same numbers I saw for my single run for the existing parity tests. I still need to run them again multiple times to get more results; just haven't gotten to it.

I'm curious why the new tests (this PR) ran so much faster in your environment than mine when the existing tests took about the same time. Did you clear out all your docker images before you ran the new tests? I did that with `docker system prune -af` before I ran the new tests on my system.

mdelapenya:
I'm just running these tests removing elasticsearch and metricbeat images from the Docker host.

To speed things up on CI, those images can be cached in the CI workers, as we already do for other projects

mdelapenya:
New results! 🍞 

Removing docker images, so both elasticsearch and metricbeat are pulled during the process:

```
OP_LOG_LEVEL=INFO go test -v --godog.format pretty parity-tests  6.30s user 3.19s system 3% cpu 4:25.09 total
```
One of the benefits of Docker is this reusability of the Docker cache, so I would like to keep these results as valid, because it's how Docker works. If the existing framework is not caching resources between runs (it downloads a TAR file from the network) then maybe that could be a good candidate for improvement.

ycombinator:
> One of the benefits of Docker is this reusability of the Docker cache, so I would like to keep these results as valid, because it's how Docker works. If the existing framework is not caching resources between runs (it downloads a TAR file from the network) then maybe that could be a good candidate for improvement.

++ agreed!

I don't know too much about our CI environments so I might need some education here. Given that we use ephemeral workers for CI, does that impact cacheability? Specifically, it is possible to cache Docker images (or other artifacts) on each worker and have it preserved between two uses of the same worker?

mdelapenya:
Yes! We can bake packer images with some docker images already present in the host, under `/var/lib/docker`. As an example see https://github.com/elastic/beats/blob/master/.ci/packer_cache.sh

We also generate most of the Beats integrations Docker images daily, and push them to our own registry, to avoid building them on each CI build (See https://apm-ci.elastic.co/blue/organizations/jenkins/beats%2Fbeats-docker-images-pipeline/detail/beats-docker-images-pipeline/200/pipeline/57).

But this is not our case, as we consume already existing Docker images. Nevertheless, I could imaging creating images for a specific PR on metricbeat, pushing it to our registry, and testing running these tests with it. Everything automated, of course

ycombinator:
Gotcha, thanks. 

In that case I agree that there will be a performance gain from using the E2E testing framework for the parity tests (this PR), which use Docker images over the existing implementation. Even if we were to try and cache tar files for the existing implementation, that counts towards extra effort in that framework because it's not something we get "for free" already. Whereas it sounds like the analogous effort for caching Docker images on CI workers has already been made so we will get it "for free".

Concretely, we're looking at 195.666s with the E2E testing framework (taken from [this comment](https://github.com/elastic/e2e-testing/pull/113#issuecomment-634496437)) vs. 291.333s (taken as the average of 3 values from [this comment](https://github.com/elastic/e2e-testing/pull/113#issuecomment-634790265)). That's **a performance gain of 95s or 32%**, which is not insignificant IMO.

So _purely_ from a performance perspective, I'm now 👍  to migrate the tests over to the E2E framework. I do think we should wait to hear @chrisronline and @igoristic's thoughts on the maintainability perspective since they are the ones maintaining the existing tests and would have to maintain the new ones if we migrate.

mdelapenya:
I'm still curious about the CPU usage 🤔 but I guess it's a Docker Vs Vagrant thing. I did not expect such a difference!

Docker: 3%
Vagrant: 48%

liza-mae:
Nice performance analysis. I am planning to remove the use of Vagrant in CI. Ansible can be used directly, with docker or a cloud provider.

mdelapenya:
> Nice performance analysis. I am planning to remove the use of Vagrant in CI. Ansible can be used directly, with docker or a cloud provider.

We use Ansible as Cloud provisioner replacing Terraform, and it works pretty well!

mdelapenya:
@chrisronline @igoristic @ycombinator Please let us know if you have any questions about this PR 🙏  we are glad to help!

mdelapenya:
Hi 👋 

Would you mind if I close this one until a final decision is taken? You could reopen it at any time if/when needed.

Thanks in advance!

chrisronline:
Sounds good to me. We are a bit swamped and won't be able to look at this in the near future

mdelapenya:
> Sounds good to me. We are a bit swamped and won't be able to look at this in the near future

Never mind! We have the commits here ready to come to them whenever there is more bandwidth.

Thank you!

PR #123, opened 2020-05-28
Use a separate layout per test suite

### LINKED ISSUE #117: Move each subproject to a separate directory
To improve the maintainability of the project, we should move each FeatureContext to a separate project, but why?

The main reason why we followed the current approach is to bypass Godog's lifecycle when filtering by tags, which is not filtering feature contexts (as demonstrated here https://github.com/mdelapenya/sample-godog) we had to create a map with all our subprojects: metricbeat-apache, metricbeat-redis, helm-metricbeat, helm-apm-server, etc.  [This map](https://github.com/elastic/e2e-testing/blob/master/e2e/runner_test.go#L32-L75), which connects feature files and feature context providers, could grow a lot.

Having separate subprojects: metricbeat, helm, and stack-monitoring would allow to run them in isolation, loading each subproject code on each run.

For that we must:
- move common, shared code to a library
- create a directory per subproject, representing an autonomous test project consuming the commonalities provided by the shared test library. This test project will use go modules.
- modify automation to iterate through subprojects, instead of using the map and `go test`
- use `godog` binary again

cc/ @jsoriano

mdelapenya:
## What does this PR do?
It moves each test suite to a separate ddirectory, so that it's possible to run godog directly, without the abstraction we created to handle filtering different feature contexts.

We have to export many test helper methods to be consumed by the new test suites projects.

## Why is it important?
We do not want to maintain a map with every featureContext object, which is a custom abstraction to workaround the limitation of godog FeatureContext, which are discovered in runtime, independently of the filtered tags used when calling godog binary. Now each suite will be a separate project, so we won't see different FeatureContext's life cycle hooks (Befores and Afters) to be run when filtering by a Godog/Cucumber tag, which was the main reason for moving from godog to `go test`.

## Related issues
- Closes #117
- Affects #113 (It contains some commits to be merged in master in the case that PR is not accepted. If accepted, then it will contain conflicts)


.ci/.e2e-tests.yaml
```diff
@@ -1,8 +1,4 @@
 ---
-FEATURES:
-  - "apache"
+SUITES:
```
mdelapenya, line 2: Waiting fr the CI to see this in practice. I'd like to add a secondary parallel execution per feature:
```yaml
- SUITES:
  - "helm":
    - "apm-server"
    - "filebeat"
    - "metricbeat"
  - "metricbeat":
    - "apache"
    - "metricbeat"
    - "mysql"
    - "redis"
    - "vsphere"
```

.ci/scripts/functional-test.sh
```diff
@@ -40,12 +41,11 @@ fi
...
 set +e
 exit_status=0
-if ! FEATURE=${FEATURE} FORMAT=junit STACK_VERSION=${STACK_VERSION} METRICBEAT_VERSION=${METRICBEAT_VERSION} make --no-print-directory -C e2e functional-test | tee ${REPORT}  ; then
+if ! SUITE=${SUITE} FEATURE=${FEATURE} FORMAT=junit STACK_VERSION=${STACK_VERSION} METRICBEAT_VERSION=${METRICBEAT_VERSION} make --no-print-directory -C e2e functional-test | tee ${REPORT}  ; then
   echo 'ERROR: functional-test failed'
   exit_status=1
 fi
 
 ## Transform report to Junit by parsing the stdout generated previously
-sed -i "s/testing: warning: no tests to run//" ${REPORT}
```
mdelapenya, line 49: No more workarounds!

e2e/Makefile
```diff
@@ -1,3 +1,4 @@
+SUITE?=metricbeat
```
mdelapenya, line 1: Added suite to CD into the right suite directory

e2e/_suites/metricbeat/metricbeat_test.go
```diff
@@ -12,32 +12,51 @@ import (
...
 	"github.com/elastic/e2e-testing/cli/services"
+	"github.com/elastic/e2e-testing/e2e"
 	log "github.com/sirupsen/logrus"
 )
 
 // metricbeatVersion is the version of the metricbeat to use
 // It can be overriden by OP_METRICBEAT_VERSION env var
 var metricbeatVersion = "7.6.0"
 
+// queryMaxAttempts is the number of attempts to query elasticsearch before aborting
```
mdelapenya, line 23: Now the runner does not contain this shared info, which only affected Metricbeat

e2e/runner_test.go
```diff
@@ -4,7 +4,6 @@ import (
 	"flag"
```
mdelapenya, line 4: This class will be removed in 3, 2, 1...

dracarys

.ci/.e2e-tests.yaml
mdelapenya: I finally went with this approach: 
```yaml
---
SUITES:
  - suite: "helm"
    feature: "apm-server"
  - suite: "helm"
    feature: "filebeat"
  - suite: "helm"
    feature: "metricbeat"
  - suite: "metricbeat"
    feature: "apache"
  - suite: "metricbeat"
    feature: "metricbeat"
  - suite: "metricbeat"
    feature: "mysql"
  - suite: "metricbeat"
    feature: "redis"
  - suite: "metricbeat"
    feature: "vsphere"
```
which is more verbose, but only used on CI

e2e/_suites/helm/features/apm-server.feature
```diff
@@ -1,3 +1,4 @@
+@apm-server
```
mdelapenya, line 1: Bring back Cucumber tags, so we will simply filter test execution by tags. As an example, for the metricbeat test suite: `godog -t redis`

e2e/runner_test.go
```diff
@@ -29,93 +28,14 @@ func (c *contextMetadata) getFeaturePaths() []string {
...
-		},
-		modules: []string{"metricbeat"},
-	},
-	"vsphere": &contextMetadata{
-		contextFuncs: []func(s *godog.Suite){
-			MetricbeatFeatureContext,
-		},
-		modules: []string{"metricbeat"},
-	},
-}
```
jsoriano, line 75: :fire:

mdelapenya:
I'm going to merge this little one, and merge-with-commit #113. _Hello conflicts my old friend..._

Issue #124, opened 2020-06-03
Create a PoC for ingest management


mdelapenya:
After a demo of this framework, we'd like to create tests for Ingest Management, with the acknowledge of only supporting linux Docker containers.

We'll start with he current test script:

<table class="relative-table" style="width: 99.1251%;">
  <colgroup>
    <col style="width: 19.1519%;"/>
    <col style="width: 68.9753%;"/>
    <col style="width: 11.7314%;"/>
  </colgroup>
  <tbody>
    <tr>
      <th colspan="1">actual code / api / url / scripting</th>
      <th>test case description / action</th>
      <th>notes / pass fail</th>
    </tr>
    <tr>
      <td colspan="1">
        <br/>
      </td>
      <td>
        <p>enable Fleet user and create initial Kibana setup</p>
      </td>
      <td>
        <br/>
      </td>
    </tr>
    <tr>
      <td colspan="1">
        <br/>
      </td>
      <td>
        <p>set (if needed) the kibana /es global settings</p>
      </td>
      <td>skip in prototype - we can use Default config + enrollment key to start if we want</td>
    </tr>
    <tr>
      <td colspan="1">
        <br/>
      </td>
      <td>
        <p>add nginx package</p>
      </td>
      <td>skip in prototype</td>
    </tr>
    <tr>
      <td colspan="1">
        <br/>
      </td>
      <td>
        <p>add new configuration</p>
      </td>
      <td>skip in prototype</td>
    </tr>
    <tr>
      <td colspan="1">
        <br/>
      </td>
      <td>
        <p>add nginx data source to configuration, with some defaults</p>
      </td>
      <td>skip in prototype</td>
    </tr>
    <tr>
      <td colspan="1">
        <br/>
      </td>
      <td>
        <p>get new enrollment key for that config</p>
      </td>
      <td>skip in prototype</td>
    </tr>
    <tr>
      <td colspan="1">
        <br/>
      </td>
      <td>
        <p>get the whole enrollment string from Ingest</p>
      </td>
      <td>skip in prototype</td>
    </tr>
    <tr>
      <td colspan="1">
        <br/>
      </td>
      <td>
        <p>transfer Agent binary, unzip it on host and enroll it with noted command, if easy que</p>
      </td>
      <td>
        <br/>
      </td>
    </tr>
    <tr>
      <td colspan="1">
        <br/>
      </td>
      <td>
        <p>In Kibana (or ES) query against Fleet API for the new Agent</p>
      </td>
[...]

PR #126, opened 2020-06-03
feat: initial specs for ingest management

mdelapenya:
## What does this PR do?
It adds the initial specs for the Ingest management project.

## Why is it important?
We should start a discussion around them to make them perfect and totally understandable by anybody in the team: product owners, developers, testers, consumers, etc.

## Related issues
- Relates #124


e2e/_suites/ingest-manager/features/ingest-manager.feature
```diff
@@ -0,0 +1,46 @@
...
+  Then filebeat is started
+    And metricbeat is started
+    And endpoint is started
+
+Scenario: Stopping the agent stops backend processes
+  Given an agent is running in a host
+  When the agent is stopped in the host
+  Then filebeat is stopped
+    And metricbeat is stopped
+    And endpoint is stopped
```
mdelapenya, line 46: Exactly the same as above. We could also combine them into one:

```gherkin
   "my-lovely-process" is in the state "stopped"
```

So the implementation method would accept both.

e2e/_suites/ingest-manager/features/ingest-manager.feature
```diff
@@ -0,0 +1,46 @@
...
+  When the agent binary is installed in the target host
+  Then the dashboards for the agent are present in Elasticsearch
+    And the agent shows up in Kibana
+
+Scenario: Un-enrolling an agent
+  Given there is a "Fleet" user in Kibana
+    And the "Fleet" Kibana setup has been created
+    And the agent binary is installed in the target host
+  When the agent is un-enrolled from Kibana
+  Then no new data shows up in Elasticsearc locations using the enrollment token
```
mdelapenya, line 16: I added `using the enrollment token` to match an existing step below. Is this assumption correct?

e2e/_suites/ingest-manager/features/ingest-manager.feature
```diff
@@ -0,0 +1,46 @@
+@ingest
```
mdelapenya, line 1: It's possible to add annotations in the form of tags. We can then instrument the test runner to filter by tags, even combining multiple of them:

```shell
godog -t "ingest && unenroll"
```

e2e/_suites/ingest-manager/features/ingest-manager.feature
```diff
@@ -0,0 +1,46 @@
...
+Feature: Enable Fleet user and create initial Kibana setup
+
+Scenario: Enrolling an agent
+  Given there is a "Fleet" user in Kibana
+    And the "Fleet" Kibana setup has been created
+  When the agent binary is installed in the target host
+  Then the dashboards for the agent are present in Elasticsearch
```
mdelapenya, line 8: I'd like to know the exact data needed here: the ES query

e2e/_suites/ingest-manager/features/ingest-manager.feature
mdelapenya, line 16: What data is not present here? I'd be great to understand more about its nature to identify when it shows up and when not

e2e/_suites/ingest-manager/features/ingest-manager.feature
```diff
@@ -0,0 +1,46 @@
...
+  Then no new data shows up in Elasticsearc locations using the enrollment token
+
+Scenario: Enrolling, un-enrolling and re-enrolling an agent
+  Given there is a "Fleet" user in Kibana
+    And the "Fleet" Kibana setup has been created
+    And the agent binary is installed in the target host
+    And the agent is un-enrolled from Kibana
+  When the agent is re-enrolled from the host
+    And the agent runs from the host
+  Then the agent shows up in Kibana
```
mdelapenya, line 25: We will need here the exact thing to check: and API call, an XPATH element in the UI...

e2e/_suites/ingest-manager/features/ingest-manager.feature
```diff
@@ -0,0 +1,46 @@
+@ingest
+Feature: Enable Fleet user and create initial Kibana setup
+
+Scenario: Enrolling an agent
+  Given there is a "Fleet" user in Kibana
+    And the "Fleet" Kibana setup has been created
```
mdelapenya, line 6: I'd group this two, which seems to me as an always-together pair, into one step:

```gherkin
Given "Fleet" is configured in Kibana
```

e2e/_suites/ingest-manager/features/ingest-manager.feature
```diff
@@ -0,0 +1,46 @@
...
+  When the agent binary is installed in the target host
+  Then the dashboards for the agent are present in Elasticsearch
+    And the agent shows up in Kibana
```
mdelapenya, line 9: Is it possible to get this without checking the UI, maybe an API call? I'd like to avoid any U/DOMI interaction if possible

nchaulet: You can call `GET /api/ingest_manager/fleet/agents`

e2e/_suites/ingest-manager/features/ingest-manager.feature
nchaulet: here I think you should say using the `access token` when an agent enroll into fleet we exchange an enrollmont token for an access token (that is one per agent). 
One you invalidate an enrollment token, the agent already enrolled should continue to work, but you cannot enroll more agents with that enrollment token

nchaulet:
Putting this here, I think it could help you later when implementing the steps.
 https://github.com/elastic/kibana/blob/master/x-pack/test/api_integration/apis/fleet/agent_flow.ts

e2e/_suites/ingest-manager/features/ingest-manager.feature
mdelapenya: Thanks for the clarification Nicholas! Please look at L27:33 There is specific scenario for revoking the enrollment token for an agent. Is that what you mean?

mdelapenya: Mmm, reading your comment, I'd rephrase this second scenario (the one revoking the token) to this:

```gherkin
Scenario: Revoking the enrollment token for an agent
  Given there is a "Fleet" user in Kibana
    And the "Fleet" Kibana setup has been created
    And the agent binary is installed in the target host
    And the agent is un-enrolled from Kibana
  When the enrollment token is revoked
  Then no new data shows up in Elasticsearc locations using the enrollment token
    And the enrolled agent continues to work
```

And I'd create another use case:
```gherkin
Scenario: A revoked enrollment token cannot enroll more agents
  Given there is an enrollment token
  When the enrollment token is revoked
  Then it's not possible to use the token to enroll more agents
```

Does it make sense to you?

mdelapenya: BTW, we should clarify what `the enrolled agent continues to work` means: i.e. it sends data to elasticsearch, there is an endpoint we can query, a process is running in the host, etc.

mdelapenya: Combining above two scenarios into one:

```gherkin
Scenario: Revoking the enrollment token for an agent
  Given there is an agent enrolled with an enrollment token
  When the enrollment token is revoked
  Then it's not possible to use the token to enroll more agents
    And the enrolled agent continues to work
```

e2e/_suites/ingest-manager/features/ingest-manager.feature
```diff
@@ -0,0 +1,49 @@
...
+  Given an agent is deployed to Fleet
+  When the agent is stopped on the host
+  Then filebeat is stopped
+    And metricbeat is stopped
+
+@unenroll
+Scenario: Un-enrolling an agent
+  Given an agent is deployed to Fleet
+  When the agent is un-enrolled
+  Then the agent is not listed as online in Fleet
```
mdelapenya, line 29: Hey @EricDavisX, could we reuse here the step `the agent is listed in Fleet as "online"`?

Therefore we would have:

```
the agent is listed in Fleet as "online"
the agent is listed in Fleet as "offline"
```
which would be one single step. wdyt?

e2e/_suites/ingest-manager/features/ingest-manager.feature
```diff
@@ -0,0 +1,49 @@
...
+    And no new documents are inserted into Elasticsearch
+
+@reenroll
+Scenario: Re-enrolling an agent
+  Given an agent is enrolled
+    And the agent is un-enrolled
+    And the Agent is stopped on the host
```
mdelapenya, line 36: Is this one automatically inferred from un-enrolling the agent, or must be done as a separate action?

If the later, I would keep it as is (although from user's perspective it seems more -unproductive?- work)

e2e/_suites/ingest-manager/features/ingest-manager.feature
```diff
@@ -0,0 +1,49 @@
...
+  When the agent is re-enrolled on the host
+    And the agent is run on the host
+  Then the agent is listed in Fleet as online
+    And new documents are inserted into Elasticsearch
```
mdelapenya, line 40: I'd like to abstract this step to a more product-related level, as I see it very technical.

e2e/_suites/ingest-manager/features/ingest-manager.feature
```diff
@@ -0,0 +1,49 @@
...
+  When an agent is deployed to Fleet
+  Then the agent is listed in Fleet as online
+    And system package dashboards are listed in Fleet
+    And new documents are inserted into Elasticsearch
+
+@start-agent
+Scenario: Starting the agent starts backend processes
+  When an agent is deployed to Fleet
+  Then filebeat is started
+    And metricbeat is started
```
mdelapenya, line 16: wdyt about adding "... on the host", so we can reuse the existing steps?

e2e/_suites/ingest-manager/features/ingest-manager.feature
```diff
@@ -0,0 +1,49 @@
...
+  Then the agent is listed in Fleet as online
+    And new documents are inserted into Elasticsearch
+
+@revoke-token
+Scenario: Revoking the enrollment token for an agent
+  Given an agent is enrolled
+  When the enrollment token is revoked
+    And new documents are inserted into Elasticsearch
+  Then the agent is un-enrolled
+    And the agent is stopped on the host
```
mdelapenya, line 48: See https://github.com/elastic/e2e-testing/pull/126/files#r436508010

e2e/_suites/ingest-manager/features/ingest-manager.feature
```diff
@@ -0,0 +1,49 @@
...
+Scenario: Starting the agent starts backend processes
+  When an agent is deployed to Fleet
+  Then filebeat is started
+    And metricbeat is started
+
+@stop-agent
+Scenario: Stopping the agent stops backend processes
+  Given an agent is deployed to Fleet
+  When the agent is stopped on the host
+  Then filebeat is stopped
```
michalpristas, line 22: we would need probably more like `Then there are '2' metricbeat processes` as we will need to check monitoring and ingesting beats

EricDavisX:
This is so great - moving fast and getting better!  I agree with Michal we could enhance the 'stopped on the host' to indicate more accurate that the Agent (with defaults set) will start 2 of the Metricbeat and 2 of the Filebeat processes.  Sorry I forgot that nuance.  Still, for the first version we can leave it as implied and handle it in the implementation assertion (and correct it in a coming PR)

 - I say that so we can truly iterate on only what is most critical to getting a test in and running fast.

e2e/_suites/ingest-manager/features/ingest-manager.feature
michalpristas: what about something like 
```
index `xxx` is created 
And index `xxx` has more than 123 documents
```

mdelapenya: I like it! What if the number of documents is not there after an amount of time (minutes)?

EricDavisX:
quick comment on the step:
    And new documents are inserted into Elasticsearch

 @mdelapenya
mdelapenya 7 hours ago Author Member
I'd like to abstract this step to a more product-related level, as I see it very technical.

 @michalpristas
michalpristas 28 minutes ago Member
what about something like

index `xxx` is created 
And index `xxx` has more than 123 documents
 @mdelapenya
mdelapenya 24 minutes ago Author Member
I like it! What if the number of documents is not there after an amount of time (minutes)?

from @EricDavisX I don't mind any rework we want to do in elaborating this.  I would like to suggest we keep it really stable and simple however, and I don't know if a given # of documents over a given amount of time would be.  The Filebeat / Metricbeat info sent is based on host vm activity, right?  I suggest if we have control over the environment and agents then we should be able to wait seconds (not minutes) and confirm changes regarding what Docs the Agent are sending in.  Might be good to take this off line and discuss in a quick call if we have this and any other 'final' items before we can get further into the implementation.

mdelapenya:
Cool! Let's discuss about the specific implementation details in a follow-up iteration. Then I'd keep that step as `And there is data for Ingest Manager in the index`. Then it's an implementation detail

EricDavisX:
that sounds great to me.  thanks Manu

mdelapenya:
@EricDavisX @michalpristas I think we are in the right track! Please let me know if the requirements are ready to be merged, so I can continue with the implementation

Thanks!

e2e/_suites/ingest-manager/features/ingest-manager.feature
mdelapenya: I like this scenario, because it's very straight-forward and simple at the same time. I'd replace what we had. wdyt about rephrasing the `Given...` to `Given an agent is enrolled`? Or do we want to make it clear for this scenario that we need the fleet user and the existence of a valid token?

mdelapenya: This query is perfect! :)

mdelapenya: BTW, in what state would be the existing agent? Will it pause? will it continue to send data?

mdelapenya:
@EricDavisX @michalpristas merged!

I'm going to send a PR with the Go code scaffolding, so please feel free to contribute to it in the way you prefer

e2e/_suites/ingest-manager/features/ingest-manager.feature
EricDavisX: note, the above api_key relates to an ephemeral instance of Kibana and it was anonymized / changed anyways (it was not a valid key).

e2e/_suites/ingest-manager/features/ingest-manager.feature
michalpristas: yes exactly this should fail then

PR #132, opened 2020-06-08
[Ingest Manager] Add basic implementation for stand-alone mode scenarios

mdelapenya:
## What does this PR do?
This PR creates the basis scaffolding for the tests, starting the runtime dependencies in the form of a profile (see modified files under the /cli dir). This profile is a Docker Compose file with all runtime dependencies for ingest-manager. At this moment it runs elasticsearch, kibana and package-registry, but it would eventually run a base Debian/Ubuntu/Alpine machine representing a vanilla host to install the agent on it.
>The current setup for runtime dependencies has been extracted from @mtojek's work here:  https://github.com/elastic/integrations/tree/master/testing/environments 

We are implementing here the first step: creating and checking the setup for adding an agent in stand-alone mode.

## Why is it important?
It creates the minimal setup to run tests, including life cycle hooks to provision the state needed by this test suite, destroying it at the end of the test suite.

## How to test this
Apart of having Go installed in the system (https://golang.org/dl/) in its >1.13.x version, please install `godog`, which is the Cucumber test runner for Go (https://github.com/cucumber/godog/#install)

Then, from the ingest-manager tests run:

```shell
$ godog -t stand_alone_mode
```

It's possible to increase the log level of the test framework, which uses Logrus, with the OP_LOG_LEVEL=DEBUG environment variable (default: INFO).

The only scenario that is implemented (and passing) is:
```gherkin
@start-agent
Scenario: Starting the agent starts backend processes
  Given a stand-alone agent is deployed
  Then the "filebeat" process is "started" on the host
    And the "metricbeat" process is "started" on the host
```


e2e/_suites/ingest-manager/features/ingest_manager.feature
```diff
@@ -13,15 +13,15 @@ Scenario: Deploying an agent
...
+  Then the "filebeat" process is "started" on the host
+    And the "metricbeat" process is "started" on the host
 
 @stop-agent
 Scenario: Stopping the agent stops backend processes
   Given an agent is deployed to Fleet
-  When the agent is "stopped" on the host
-  Then "filebeat" is "stopped" on the host
-    And "metricbeat" is "stopped" on the host
+  When the "agent" process is "stopped" on the host
```
mdelapenya, line 22: With this subtle change we can reuse steps. What I'd like to know is if we can consider the agent a process

e2e/kibana.go
```diff
@@ -0,0 +1,54 @@
...
+
+// WaitForKibanaFromHostPort waits for kibana running in a host:port to be healthy, returning false
+// if kibana does not get healthy status in a defined number of minutes.
+func WaitForKibanaFromHostPort(host string, port int, maxTimeoutMinutes time.Duration) (bool, error) {
+	exp := getExponentialBackOff(maxTimeoutMinutes)
+
+	retryCount := 1
+
+	kibanaStatus := func() error {
+		kibanaEndpoint := fmt.Sprintf("http://%s:%d/status", host, port)
```
mdelapenya, line 26: @v1v There was a log entry to check this, but could not find it 😕

e2e/_suites/ingest-manager/ingest-manager_test.go
```diff
@@ -0,0 +1,208 @@
...
+	err = curl.Post(fleetSetupURL, payloadBytes)
+	if err != nil {
+		log.WithFields(log.Fields{
+			"error": err,
+			"url":   fleetSetupURL,
+		}).Error("Could not initialise Fleet")
+		return err
+	}
+
+	log.Debug("Ensuring Fleet was initialised")
```
mdelapenya, line 130: I'd like to check that Fleet was configured properly: is there a check (on Kibana API or elasticsearch) to create an assertion for this?

e2e/kibana.go
v1v: See https://github.com/elastic/kibana/pull/61803/files#diff-2766238c50e581d8e076ac51a70fe255R123-R124

kuisathaverat: IIRC this only happens when you start Kibana for dev, on the Docker container it does not happens or at least we do not see this behavior in the test clusters

EricDavisX:
@mdelapenya hi - to check the fleet setup is done right you can poll / call that api again:
/api/ingest_manager/fleet/setup
 and when its done and ready it should give this output response:
{
"isReady": true,
"missing_requirements": []
}

That's a first step.  :)  The product ensuring us its super great and ready to go.  The details of what it is doing that we might check with data in ES or Kibana I'm not sure of... I know it creates 'the Fleet user' which we could query for.  @jfsiii or @ruflin or @nchaulet do you know what else we could do here to make the test more concrete in asserting on the details?

mdelapenya:
> @mdelapenya hi - to check the fleet setup is done right you can poll / call that api again:
> /api/ingest_manager/fleet/setup
> and when its done and ready it should give this output response:
> {
> "isReady": true,
> "missing_requirements": []
> }

I think this check is enough for our purpose 👍 

> That's a first step. :) The product ensuring us its super great and ready to go. The details of what it is doing that we might check with data in ES or Kibana I'm not sure of... I know it creates 'the Fleet user' which we could query for. @jfsiii or @ruflin or @nchaulet do you know what else we could do here to make the test more concrete in asserting on the details?

If we wanted to add a scenario for the configuration, I would check those items for sure. Maybe we can do it in a later iteration. wdyt?

mdelapenya:
Hey @EricDavisX I executed the above POST request (http://localhost:5601/api/ingest_manager/fleet/setup), right after kibana status returns a 200 OK, and got this:

```
INFO[0040] {"statusCode":404,"error":"Not Found","message":"Not Found"} 
```

Do you know if we have to wait for something or do any other action to activate Fleet?

mdelapenya:
As an example, I added the CI execution https://apm-ci.elastic.co/blue/organizations/jenkins/stack%2Fe2e-testing-mbp/detail/PR-132/5/pipeline/206#step-303-log-498

Here you can check that Kibana does not recognises that URL. Any guess?

EricDavisX:
I apologize I failed to give you the payload for the post.
{
"admin_username": "string",
"admin_password": "string"
}

I don't know how to get the values there for the uid/pwd at the moment.  @nchaulet can you help with this specific request?  make sure we have given Manu what he needs?

mdelapenya:
What I'm worried about is the 404 error when doing that POST request. A GET request (the one used to check status returns a 404 too, so I guess the endpoint URL is not there.

mdelapenya:
> I apologize I failed to give you the payload for the post.
> {
> "admin_username": "string",
> "admin_password": "string"
> }

I guess they come from the elasticsearch setup, so it's on the test framework spinning up the elasticsearch side. Nevertheless, a confirmation would be superb :)

e2e/_suites/ingest-manager/ingest-manager_test.go
jfsiii: I believe you can use the `get-fleet-setup` route (`GET /api/ingest_manager/fleet/setup`) which ~will return an object like `{ isInitialized: true }` or `{ isInitialized: false }`~

~From the OpenAPI spec file
https://github.com/elastic/kibana/blob/974dcab91733ff7adeabb760e0373d5bbcbfb1cf/x-pack/plugins/ingest_manager/common/openapi/spec_oas3.json#L1417-L1459~

The spec file is incorrect. I believe the actual response shape is 

```typescript
{
  isReady: boolean;
  missing_requirements: Array<'tls_required' | 'api_keys' | 'fleet_admin_user'>;
}
```

From https://github.com/elastic/kibana/blob/5e972e14d147e3c5232114690f61a96a493a28ae/x-pack/plugins/ingest_manager/common/types/rest_spec/fleet_setup.ts#L11-L15

mdelapenya: The POST request for creating the setup returns `isInitialized: true`, so I guess this could be enough

e2e/_suites/ingest-manager/features/ingest_manager.feature
michalpristas: ++ for reusability

ruflin:
An additional test (for later) that I do in this PR here is to validate if all the available packages can be installed: https://github.com/elastic/package-storage/pull/54/files#diff-6a13f5c76026f2845117497754dc63e1R95

.ci/.e2e-tests.yaml
```diff
@@ -7,7 +7,7 @@ SUITES:
   - suite: "helm"
     feature: "metricbeat"
   - suite: "ingest-manager"
-    feature: "all"
+    feature: "stand_alone_mode"
```
mdelapenya, line 10: Running only the scenarios of interest

Issue #130, opened 2020-06-05
Execute tests suites based on the modified files


mdelapenya:
It would be great if we can decide which test suite to run

mdelapenya:
This has already been implemented

![image](https://media.giphy.com/media/AOBSDDpXmIOIM/giphy.gif)

PR #133, opened 2020-06-09
feat(ci): support selecting which suite to run

mdelapenya:
## What does this PR do?
It adds an input parameter to the CI pipeline, supporting running all or one of the existing test suites, setting 'all' as default.

## Why is it important?
It will allow running this pipeline with less parallel branches, being the case we want to run just one branch. This filter will be provided by a Jenkins input parameter.

## Related issues
- Relates #130

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -24,6 +24,7 @@ pipeline {
     issueCommentTrigger('(?i).*(?:jenkins\\W+)?run\\W+(?:the\\W+)?tests(?:\\W+please)?.*')
   }
   parameters {
+    choice(name: 'runTestsSuite', choices: ['all', 'helm', 'metricbeat'], description: 'Choose which test suite to run (default: all)')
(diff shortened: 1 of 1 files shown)


.ci/Jenkinsfile
```diff
@@ -24,6 +24,7 @@ pipeline {
     issueCommentTrigger('(?i).*(?:jenkins\\W+)?run\\W+(?:the\\W+)?tests(?:\\W+please)?.*')
   }
   parameters {
+    choice(name: 'runTestsSuite', choices: ['all', 'helm', 'metricbeat'], description: 'Choose which test suite to run (default: all)')
```
cachedout, line 27: Nit: It seems impossible to run helm+metricbeat but not all. Is that intentional?

cachedout: I don't have any opinion about what's right here. :) I was just raising the point in case it was something you wanted to handle. Thanks!

mdelapenya: Sure! At this moment I could not say we are not going to need it, although given the current  parallel implementation of the CI in separate workers, I would not see potential collision issues while running different combinations of them.

PR #140, opened 2020-06-18
[Ingest-Manager] Implementation for the Deploy scenario of the stand-alone mode

mdelapenya:
## What does this PR do?
It adds the code for checking that there are documents in the index. To achieve it, we use the query provided here: https://github.com/elastic/e2e-testing/pull/126#discussion_r435898461

To perform this query, we need the hostname, which in a docker container matches the container id. Nevertheless, we execute the `hostname` command in the agent container.

To query Elasticsearch when security is enabled, we had to refactor a bit the existing helper methods for querying, adding authentication to the creation of the ES client. Besides that, we refactored the helper method to check for hits in an ES response.

## Why is it important?
It puts in green the three BDD scenarios for the stand-alone mode of the Elastic Agent

## Follow-up considerations
I noticed in Kibana that the hits always belong to this index: "logs-agent-default", that's why I'm hardcoding it.


e2e/_suites/ingest-manager/stand-alone.go
```diff
@@ -66,7 +70,110 @@ func (sats *StandAloneTestSuite) aStandaloneAgentIsDeployed() error {
...
+	hostname, err := getContainerName(serviceName)
+	if err != nil {
+		log.WithFields(log.Fields{
+			"error":   err,
+			"service": serviceName,
+		}).Error("Could not retrieve container name from the Docker client")
+		return err
+	}
+
+	esQuery := map[string]interface{}{
```
mdelapenya, line 87: Translated query from: https://github.com/elastic/e2e-testing/pull/126#discussion_r435898461

e2e/elasticsearch.go
```diff
@@ -83,6 +83,8 @@ func getElasticsearchClientFromHostPort(host string, port int) (*es.Client, erro
 
 	cfg := es.Config{
 		Addresses: []string{fmt.Sprintf("http://%s:%d", host, port)},
+		Username:  "elastic",
```
mdelapenya, line 86: I ran locally other tests using the ES client, and they pass!

e2e/_suites/ingest-manager/ingest-manager_test.go
```diff
@@ -21,12 +21,22 @@ var stackVersion = "8.0.0-SNAPSHOT"
...
 var profileEnv map[string]string
 
+// queryMaxAttempts is the number of attempts to query elasticsearch before aborting
+// It can be overriden by OP_QUERY_MAX_ATTEMPTS env var
+var queryMaxAttempts = 5
+
+// queryRetryTimeout is the number of seconds between elasticsearch retry queries.
+// It can be overriden by OP_RETRY_TIMEOUT env var
+var queryRetryTimeout = 3
+
```
mdelapenya, line 31: As a follow-up improvement, I'm considering converting this basic retry into the backoff strategy. So it's in the TODO

.ci/Jenkinsfile
```diff
@@ -26,8 +26,7 @@ pipeline {
   parameters {
     choice(name: 'runTestsSuite', choices: ['all', 'helm', 'ingest-manager', 'metricbeat'], description: 'Choose which test suite to run (default: all)')
     choice(name: 'LOG_LEVEL', choices: ['INFO', 'DEBUG'], description: 'Log level to be used')
-    choice(name: 'QUERY_MAX_ATTEMPTS', choices: ['5', '10', '20'], description: 'Number of attempts to create the connection to Elasticsearch')
```
mdelapenya, line 29: Removed this argument, as the code is not using a fixed retry option

e2e/_suites/ingest-manager/ingest-manager_test.go
```diff
@@ -75,6 +80,8 @@ func IngestManagerFeatureContext(s *godog.Suite) {
 				"minutes": minutesToBeHealthy,
 			}).Error("The Kibana instance could not get the healthy status")
 		}
+
+		imts.StandAlone.RuntimeDependenciesStartDate = time.Now()
```
mdelapenya, line 84: Marking when the runtime deps are started. We want to have here a lower boundary for queries

e2e/_suites/ingest-manager/stand-alone.go
```diff
@@ -66,13 +81,156 @@ func (sats *StandAloneTestSuite) aStandaloneAgentIsDeployed() error {
...
+	}
+	sats.AgentStoppedDate = time.Now()
+
+	return nil
 }
 
 func (sats *StandAloneTestSuite) thereIsNoNewDataInTheIndexAfterAgentShutsDown() error {
-	return godog.ErrPending
+	maxTimeout := time.Duration(30) * time.Second
+	minimumHitsCount := 1
```
mdelapenya, line 116: With just one hit found in 30 secs, we should fail the assertion

e2e/_suites/metricbeat/metricbeat_test.go
```diff
@@ -403,12 +398,25 @@ func (mts *MetricbeatTestSuite) thereAreEventsInTheIndex() error {
...
+	minimumHitsCount := 5
+	maxTimeout := time.Duration(queryRetryTimeout) * time.Minute
+
+	result, err := e2e.WaitForNumberOfHits(mts.getIndexName(), esQuery, minimumHitsCount, maxTimeout)
 	if err != nil {
 		return err
 	}
 
-	return e2e.AssertHitsArePresent(result, mts.Query)
+	err = e2e.AssertHitsArePresent(result)
```
EricDavisX, line 409: updating Metricbeats test at same time - you're doing good things here!  Very Elastic of you.

e2e/elasticsearch.go
```diff
@@ -256,3 +257,48 @@ func WaitForElasticsearchFromHostPort(host string, port int, maxTimeoutMinutes t
 
 	return true, nil
 }
+
+// WaitForNumberOfHits waits for an elasticsearch query to return more than a number of hits,
+// returning false if the query does not reach that number in a defined number of time.
+func WaitForNumberOfHits(indexName string, query map[string]interface{}, desiredHits int, maxTimeout time.Duration) (SearchResult, error) {
```
EricDavisX, line 263: this is fabulous.  I wonder if (in a coming PR) the next step will be to parametrize this further and make it even more re-usable than it is.  I see we have the index, and # of hits, timeout, and the query.  

It may not be easy or quite appropriate, but I expect keeping fewer copies of the big 'query' listed in code the better... we can chat it, and its certainly mergeable as is, I'm just thinking long term (the first thing someone will do is another query check for something different in ES I'm betting!)

e2e/_suites/ingest-manager/stand-alone.go
```diff
@@ -66,13 +81,156 @@ func (sats *StandAloneTestSuite) aStandaloneAgentIsDeployed() error {
 }
 
 func (sats *StandAloneTestSuite) thereIsNewDataInTheIndexFromAgent() error {
-	return godog.ErrPending
+	maxTimeout := time.Duration(queryRetryTimeout) * time.Minute
+	minimumHitsCount := 100
+
+	result, err := searchAgentData(sats.Hostname, sats.RuntimeDependenciesStartDate, minimumHitsCount, maxTimeout)
```
mdelapenya, line 87: We use the moment the runtime deps (kibana + ES) started to checc for documents

e2e/_suites/ingest-manager/stand-alone.go
```diff
@@ -66,13 +81,156 @@ func (sats *StandAloneTestSuite) aStandaloneAgentIsDeployed() error {
...
+
+	return nil
 }
 
 func (sats *StandAloneTestSuite) thereIsNoNewDataInTheIndexAfterAgentShutsDown() error {
-	return godog.ErrPending
+	maxTimeout := time.Duration(30) * time.Second
+	minimumHitsCount := 1
+
+	result, err := searchAgentData(sats.Hostname, sats.AgentStoppedDate, minimumHitsCount, maxTimeout)
```
mdelapenya, line 118: We use the moment the agent stopped to check for documents

EricDavisX: I should have caught this earlier, but indeed - we should put in some minimal (1 second or 2??) padding on the chance that a doc was sent right as agent is stopping and is in ES microseconds after the agent stop time is recorded  @mdelapenya what do you think?

mdelapenya: Flakiness at sight!!

TBH not sure, we can add a few seconds to that time

mdelapenya: Let's merge as is, so we can observe how it behaves

PR #143, opened 2020-06-24
chore(ci): add a pipeline for the daily trigger of the ingest manager tests

### LINKED ISSUE #142: Automate the schedule of ingest-manager tests
## Specification
We'd like to run the e2e tests for ingest-manager once a day, or even better, coupling the execution with the release of the different snapshots for the elastic-agent (i.e. 8.0.0-SNAPSHOT) (See https://beats-ci.elastic.co/job/Beats/job/packaging).

Finally, we'd like to post the job results in the Slack channel for Ingest Manager: `#ingest-management`

1) Daily run:
```gherkin
When it's 23:59 CEST
Then the CI will run the ingest-manager test suite
    And will notify to the "ingest-management" slack channel
```
2) Coupled with the packaging release:
```gherkin
When there is a new release of the elastic-agent
Then the CI will run the ingest-manager test suite
    And will notify to the "ingest-management" slack channel
```

## What we want to do?
We should use the second option only once the test suite is stable and mature enough to avoid adding noise to the slack channel. For that reason we should firstly go with the daily schedule option.

## Follow-up improvements
Once the process is mature enough, we could add this test suite to the PR stage, so any PR would run them. For that, we'd need a priorization of the existing scenarios, so that we run those with the highest priority in this PR suite. That priorization needs to be done by the functional team (product owner, team leads, team).

cc/ @elastic/observablt-robots @EricDavisX

mdelapenya:
## What does this PR do?
It adds a regular Jenkins pipeline that triggers the already existing job for the e2e tests, customising the execution with predefined input parameters, setting the ingest-manager test suite to be run.

It's also possible to manually run this job with different configurations:
- Go version
- Stack version
- Retry timeout
- Log level

If not set, it will use the triggered build's defaults for configuration

Besides that, all JJBB definitions have been added here, as we opened a request to infra to discover JJBB jobs in this repo.

## Why is it important?
We want to schedule the execution of the ingest-manager tests in a daily manner.

JJBB definitions in the same repo ==> self-content repo ==> 👍 

## Related issues
- Closes #142

### DIFF
--- a/.ci/e2eTestingIngestManagerDaily.groovy
+++ b/.ci/e2eTestingIngestManagerDaily.groovy
@@ -59,7 +59,7 @@ pipeline {
             string(name: 'STACK_VERSION', value: "${params.STACK_VERSION.trim()}"),
             string(name: 'GO_VERSION', value: "${params.GO_VERSION.trim()}")
           ],
-          propagate: true,
+          propagate: false,
           wait: false
         )
       }
--- a/.ci/e2eTestingIngestManagerDaily.groovy
+++ b/.ci/e2eTestingIngestManagerDaily.groovy
@@ -1,0 +1,74 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+@Library('apm@current') _
+
+pipeline {
+  agent { label 'linux && immutable' }
+  environment {
+    REPO = 'e2e-testing'
+    BASE_DIR = "src/github.com/elastic/${env.REPO}"
+    GOPATH = "${env.WORKSPACE}"
+    HOME = "${env.WORKSPACE}"
+    JOB_GCS_BUCKET = credentials('gcs-bucket')
+    JOB_GIT_CREDENTIALS = "f6c7695a-671e-4f4f-a331-acdce44ff9ba"
+    NOTIFY_TO = credentials('notify-to')
+    PATH = "${env.GOPATH}/bin:${env.PATH}"
+    PIPELINE_LOG_LEVEL='INFO'
+  }
+  options {
+    timeout(time: 1, unit: 'HOURS')
+    buildDiscarder(logRotator(numToKeepStr: '20', artifactNumToKeepStr: '20'))
+    timestamps()
+    ansiColor('xterm')
+    disableResume()
+    durabilityHint('PERFORMANCE_OPTIMIZED')
+    rateLimitBuilds(throttle: [count: 60, durationName: 'hour', userBoost: true])
+    quietPeriod(10)
+  }
+  triggers {
+    cron '@daily'
+  }
+  parameters {
+    choice(name: 'LOG_LEVEL', choices: ['INFO', 'DEBUG'], description: 'Log level to be used')
+    choice(name: 'RETRY_TIMEOUT', choices: ['3', '5', '7', '11'], description: 'Max number of minutes for timeout backoff strategies')
+    string(name: 'STACK_VERSION', defaultValue: '8.0.0-SNAPSHOT', description: 'SemVer version of the stack to be used.')
+    string(name: 'GO_VERSION', defaultValue: '1.13.4', description: "Go version to use.")
+  }
+  stages {
+    stage('Run Tests') {
+      steps {
+        build(job: 'stack/e2e-testing-mbp',
+          parameters: [
+            string(name: 'runTestsSuite', value: 'ingest-manager'),
+            string(name: 'LOG_LEVEL', value: "${params.LOG_LEVEL.trim()}"),
+            string(name: 'RETRY_TIMEOUTRETRY_TIMEOUT', value: "${params.RETRY_TIMEOUT.trim()}"),
+            string(name: 'STACK_VERSION', value: "${params.STACK_VERSION.trim()}"),
+            string(name: 'GO_VERSION', value: "${params.GO_VERSION.trim()}")
+          ],
+          propagate: true,
+          wait: true
+        )
+      }
+    }
+  }
+  post {
+    cleanup {
+      notifyBuildResult()
+    }
+  }
+}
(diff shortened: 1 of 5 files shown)


.ci/e2eTestingIngestManagerDaily.groovy
```diff
@@ -0,0 +1,74 @@
...
+          ],
+          propagate: true,
+          wait: true
+        )
+      }
+    }
+  }
+  post {
+    cleanup {
+      notifyBuildResult()
```
mdelapenya, line 71: We have to send the result to Slack too. Do we have any example on how to do it?

v1v: I asked infra to configure the slack for the apm-ci, I need to look for the reference,  i think it was configured for the #apm-agent-rumjs slack channel. There is a slack plugin that provides the pipeline step, so theoretically it will be as simple as using it explicitly in the pipeline itself. I will look for the issue and comment it here.

.ci/e2eTestingIngestManagerDaily.groovy
```diff
@@ -0,0 +1,74 @@
...
+    buildDiscarder(logRotator(numToKeepStr: '20', artifactNumToKeepStr: '20'))
+    timestamps()
+    ansiColor('xterm')
+    disableResume()
+    durabilityHint('PERFORMANCE_OPTIMIZED')
+    rateLimitBuilds(throttle: [count: 60, durationName: 'hour', userBoost: true])
+    quietPeriod(10)
+  }
+  triggers {
+    cron '@daily'
```
cachedout, line 44: Nit: If this is to be run daily, we can probably dial up the timeout and dial down the rate limit on L34 and L40 respectively.

v1v, line 44: NIT: if no weekends then this might be changed

.ci/e2eTestingIngestManagerDaily.groovy
```diff
@@ -0,0 +1,74 @@
...
+    string(name: 'GO_VERSION', defaultValue: '1.13.4', description: "Go version to use.")
+  }
+  stages {
+    stage('Run Tests') {
+      steps {
+        build(job: 'stack/e2e-testing-mbp',
+          parameters: [
+            string(name: 'runTestsSuite', value: 'ingest-manager'),
+            string(name: 'LOG_LEVEL', value: "${params.LOG_LEVEL.trim()}"),
+            string(name: 'RETRY_TIMEOUTRETRY_TIMEOUT', value: "${params.RETRY_TIMEOUT.trim()}"),
```
v1v, line 59:
```suggestion
            string(name: 'RETRY_TIMEOUT', value: "${params.RETRY_TIMEOUT.trim()}"),
```

.ci/e2eTestingIngestManagerDaily.groovy
```diff
@@ -0,0 +1,74 @@
...
+  parameters {
+    choice(name: 'LOG_LEVEL', choices: ['INFO', 'DEBUG'], description: 'Log level to be used')
+    choice(name: 'RETRY_TIMEOUT', choices: ['3', '5', '7', '11'], description: 'Max number of minutes for timeout backoff strategies')
+    string(name: 'STACK_VERSION', defaultValue: '8.0.0-SNAPSHOT', description: 'SemVer version of the stack to be used.')
+    string(name: 'GO_VERSION', defaultValue: '1.13.4', description: "Go version to use.")
+  }
+  stages {
+    stage('Run Tests') {
+      steps {
+        build(job: 'stack/e2e-testing-mbp',
```
v1v, line 55: if this is an MBP, you might need to use the branch to be used:

```suggestion
        build(job: 'stack/e2e-testing-mbp/master',
```

.ci/e2eTestingIngestManagerDaily.groovy
```diff
@@ -0,0 +1,74 @@
...
+
+pipeline {
+  agent { label 'linux && immutable' }
+  environment {
+    REPO = 'e2e-testing'
+    BASE_DIR = "src/github.com/elastic/${env.REPO}"
+    GOPATH = "${env.WORKSPACE}"
+    HOME = "${env.WORKSPACE}"
+    JOB_GCS_BUCKET = credentials('gcs-bucket')
+    JOB_GIT_CREDENTIALS = "f6c7695a-671e-4f4f-a331-acdce44ff9ba"
```
v1v, line 28:
```suggestion
```

.ci/e2eTestingIngestManagerDaily.groovy
```diff
@@ -0,0 +1,74 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+@Library('apm@current') _
+
+pipeline {
+  agent { label 'linux && immutable' }
```
v1v, line 21: If this is just an orchestrator then you can use `none`, IIRC:

```suggestion
  agent none
```

Some of the environments might not be required

.ci/e2eTestingIngestManagerDaily.groovy
mdelapenya: Good catch on a bad copy&paste bug!

.ci/e2eTestingIngestManagerDaily.groovy
mdelapenya: Changed!

mdelapenya: I changed it, thanks!

mdelapenya: About the new rate limit... any suggestion?

.ci/e2eTestingIngestManagerDaily.groovy
```diff
@@ -0,0 +1,73 @@
...
+      steps {
+        build(job: 'stack/e2e-testing-mbp/master',
+          parameters: [
+            string(name: 'runTestsSuite', value: 'ingest-manager'),
+            string(name: 'LOG_LEVEL', value: "${params.LOG_LEVEL.trim()}"),
+            string(name: 'RETRY_TIMEOUT', value: "${params.RETRY_TIMEOUT.trim()}"),
+            string(name: 'STACK_VERSION', value: "${params.STACK_VERSION.trim()}"),
+            string(name: 'GO_VERSION', value: "${params.GO_VERSION.trim()}")
+          ],
+          propagate: true,
```
kuisathaverat, line 62: I think that the daily scheduler should not wait for the job, if it waits we'd have 2 failure notifications one for the scheduler and one for the test, but the scheduler really did not fail because it only launches another jobs

mdelapenya: Agreed. I'll change it now. I added this wait to synchronise both jobs, but I see your point and this is a pure trigger, that works when the downstream build is simply triggered.

mdelapenya: Wdyt about the propagation?

mdelapenya: I'm going to use `false` too

.ci/e2eTestingIngestManagerDaily.groovy
mdelapenya: I'll work on this in a follow-up PR, thanks!

PR #145, opened 2020-06-29
fix(ci): control git checks using a parameter

### LINKED ISSUE #144: Skip isGitRegionMatch when triggering a scheduled build
When triggering a build from the scheduled job, then the `isGitRegionMatch` causes to skip any e2e test stage

mdelapenya:
## What does this PR do?
It adds a boolean param to control if the isGitRegionMatch takes effect. This param, which is defined in the downstream job, will be set as TRUE by the scheduled caller job, bein FALSE by default

We will trigger the job from the scheduled one setting the parameter to TRUE

## Why is it important?
Otherwise the isGitRegionMatch won't detect any change for triggered, scheduled builds.

## Related issues
- Fixes #144

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -25,7 +25,7 @@ pipeline {
   }
   parameters {
     choice(name: 'runTestsSuite', choices: ['all', 'helm', 'ingest-manager', 'metricbeat'], description: 'Choose which test suite to run (default: all)')
-    booleanParam(name: "SKIP_GIT_CHECKS", defaultValue: "false", description: "If it's needed to check for Git changes to filter by modified sources")
+    booleanParam(name: "SKIP_GIT_CHECKS", defaultValue: false, description: "If it's needed to check for Git changes to filter by modified sources")
     choice(name: 'LOG_LEVEL', choices: ['INFO', 'DEBUG'], description: 'Log level to be used')
     choice(name: 'RETRY_TIMEOUT', choices: ['3', '5', '7', '11'], description: 'Max number of minutes for timeout backoff strategies')
     string(name: 'STACK_VERSION', defaultValue: '7.7.0', description: 'SemVer version of the stack to be used.')
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -49,7 +49,7 @@ pipeline {
         GOPROXY = 'https://proxy.golang.org'
         METRICBEAT_VERSION = "${params.METRICBEAT_VERSION.trim()}"
         STACK_VERSION = "${params.STACK_VERSION.trim()}"
-        SKIP_GIT_CHECKS = "${params.SKIP_GIT_CHECKS.trim()}"
+        SKIP_GIT_CHECKS = "${params.SKIP_GIT_CHECKS}"
         HELM_CHART_VERSION = "${params.HELM_CHART_VERSION.trim()}"
         HELM_VERSION = "${params.HELM_VERSION.trim()}"
         KIND_VERSION = "${params.KIND_VERSION.trim()}"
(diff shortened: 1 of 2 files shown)


.ci/Jenkinsfile
```diff
@@ -125,7 +127,7 @@ pipeline {
                     def feature = item.feature
                     if (suiteParam == "all" || suiteParam == suite) {
                       def regexps = [ "^e2e/_suites/${suite}/.*", "^.ci/.*", "^cli/.*", "^e2e/.*\\.go" ]
-                      if (isGitRegionMatch(patterns: regexps, shouldMatchAll: false)) {
+                      if ("${SKIP_GIT_CHECKS}" == "true" || isGitRegionMatch(patterns: regexps, shouldMatchAll: false)) {
```
mdelapenya, line 130: Not sure if it's possible to do a boolean comparison here in Groovish

.ci/e2eTestingIngestManagerDaily.groovy
```diff
@@ -45,6 +45,7 @@ pipeline {
   parameters {
     choice(name: 'LOG_LEVEL', choices: ['INFO', 'DEBUG'], description: 'Log level to be used')
     choice(name: 'RETRY_TIMEOUT', choices: ['3', '5', '7', '11'], description: 'Max number of minutes for timeout backoff strategies')
+    booleanParam(name: "SKIP_GIT_CHECKS", defaultValue: "true", description: "If it's needed to check for Git changes to filter by modified sources")
```
mdelapenya, line 48: Default value is true. Maybe we can hide this param and force the variable to true. wdyt?

.ci/Jenkinsfile
```diff
@@ -25,6 +25,7 @@ pipeline {
   }
   parameters {
     choice(name: 'runTestsSuite', choices: ['all', 'helm', 'ingest-manager', 'metricbeat'], description: 'Choose which test suite to run (default: all)')
+    booleanParam(name: "SKIP_GIT_CHECKS", defaultValue: "false", description: "If it's needed to check for Git changes to filter by modified sources")
```
cachedout, line 28:
```suggestion
    booleanParam(name: "SKIP_GIT_CHECKS", defaultValue: false, description: "If it's needed to check for Git changes to filter by modified sources")
```

v1v, line 28: IIRC, we use the `force` parameter in some other cases, what do you think to rename it?

cachedout: Shouldn't this be the bool of `true` instead of the string?

mdelapenya: I like the idea, being consistent across projects will bring maintainability!

mdelapenya: I copied it from apm-pipeline, and there almost all occurrences are quoted or double quoted.

mdelapenya: I'm still thinking about this: what do we want to `force` here? When triggering the build, we want to force not checking for modified paths, so not sure if `force` is providing context enough to understand what is going on.


I'd prefer using `forceSkipGitChecks` instead, as it clearly indicates what will happen. Keeping the `force` as a prefix brings consistency, although not sure if the variable name is still too long.

wdyt?

mdelapenya: Thinking about it, I do prefer having boolean values without quotes. Removed them

mdelapenya: Ups, I did not realise there was a suggestion commit. My fault  😞

v1v: I'd say something generic and atomic could help with, but we are not there yet, there are certain cases where we use "run_as_the_master_branch". So no strong opinion in my end

mdelapenya: Mmmm, interesting. That's making me think that the `master` branch for this project does not get special behavior, and maybe it should.

Anyway, thanks to this healthy discussion, I'll use `forceSkipGitChecks` to be consistent with the `force` parameter. We can rethink the behavior for master branch later on.

Thanks!

PR #147, opened 2020-06-29
chore(ci): rename variable for controlling git checks

mdelapenya:
## What does this PR do?
It renames the variables using the `force` prefix, as it's a common pattern of us.

## Why is it important?
Consistency matters

## Related issues
- Relates #144


mdelapenya:
Merging, as this is the follow-up of #145 and the CI is passing

Issue #148, opened 2020-06-29
Scenarios for Fleet-mode: initial steps


mdelapenya:
Using the existing Fleet-mode scenarios, let's implement them.

We will use [Centos:7](https://github.com/elastic/beats/blob/master/dev-tools/packaging/packages.yml#L187) as base image, as this is the distro used by the release team of Beats.

mdelapenya:
Merged, thanks!

![image](https://media.giphy.com/media/D0WOL0ogZIoG4/giphy.gif)

PR #149, opened 2020-06-29
feat: implement the installation of the agent

### LINKED ISSUE #169: Update fleet setup on Kibana
Check https://github.com/elastic/observability-test-environments/pull/690

cc/ @EricDavisX

mdelapenya:
## What does this PR do?
We are adding implementation steps for the first scenario, which installs the agent and enrolls it in Kibana.
    
We are spinning up a Centos:7 box (which is kept alive using `tail -f /dev/null` as entrypoint), and we are following the installation process of the Elastic Agent from the guide: https://www.elastic.co/guide/en/ingest-management/7.8/elastic-agent-installation.html:
1. download the binary for the target platform
1. untar it
1. create a symlink to enable "elastic-agent" in the PATH (we did this to simplify invoking the agent from outside the container)

To download the agent, we are inspecting the JSON response of querying Elastic's API for artifacts (https://artifacts-api.elastic.co), extracting the download URL for the current 8.0.0-SNAPSHOT version of the elastic-agent

>If the ARTIFACT_HASH environment variable is present, the test suite will use it as part of the URL to download, not inspecting the above JSON, therefore downloading a custom existing version of the agent. This ability is needed to bypass potential issues when working with the latest snapshot.

We enroll the agent with a new token (we added support to create token when deploying a new agent, so it's possible to revoke the token without affecting other test scenarios), and run the agent, which is shown in Kibana as online. After a few seconds, around 20, datastreams will be present in Kibana.

As a side note, we are reusing the "waitForProcess" method to check that the installation commands finished, checking that they are not there when executing "pgrep -n -l PROCESS_NAME" in the agent container

For the stop scenario, we're killing the process with `pkill`, which comes bundled into the Centos box.

## Why is it important?
Making progress with Fleet mode

## Related issues
- Relates #148
- Closes #169


cli/config/compose/profiles/ingest-manager/docker-compose.yml
```diff
@@ -29,13 +29,13 @@ services:
...
       interval: 1s
-    image: "docker.elastic.co/kibana/kibana:8.0.0-SNAPSHOT"
+    image: "docker.elastic.co/kibana/kibana:${stackVersion:-8.0.0-SNAPSHOT}"
     ports:
       - "5601:5601"
     volumes:
       - ${kibanaConfigPath}:/usr/share/kibana/config/kibana.yml
   package-registry:
-    image: docker.elastic.co/package-registry/package-registry:master
+    image: docker.elastic.co/package-registry/package-registry:41c150c8020efc53ab16e3bba774c62a419b51ea
```
EricDavisX, line 38: can we get a comment as to how folks would know how and when to update this... I think some details will be here: https://github.com/elastic/package-storage/issues/86

cli/shell/curl.go
```diff
@@ -82,9 +108,9 @@ func request(r HTTPRequest) (string, error) {
 	bodyBytes, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
 		log.WithFields(log.Fields{
-			"error":  err,
-			"method": r.method,
-			"url":    r.URL,
+			"error":      err,
+			"method":     r.method,
+			"escapedURL": escapedURL,
```
EricDavisX, line 113: I recall discussing this, glad you got it sorted.

cli/config/compose/profiles/ingest-manager/docker-compose.yml
mdelapenya: We replaced master branch because there was a combination of errors:
- Kibana was unable to start
- the enrollment process of agents did not work because of different agent versions compared on Kibana

I noticed the team changed the version here: https://github.com/elastic/integrations/commit/25a4bd4e160afd3ed763c88745d265cde30941e3, and I shadowed it.

e2e/_suites/ingest-manager/stand-alone.go
```diff
@@ -82,7 +83,7 @@ func (sats *StandAloneTestSuite) aStandaloneAgentIsDeployed() error {
 
 func (sats *StandAloneTestSuite) thereIsNewDataInTheIndexFromAgent() error {
 	maxTimeout := time.Duration(queryRetryTimeout) * time.Minute
-	minimumHitsCount := 100
+	minimumHitsCount := 50
```
EricDavisX, line 86: i agree - I might even lower it more... but this is helpful and still keeps high value

e2e/_suites/ingest-manager/stand-alone.go
```diff
@@ -230,7 +237,7 @@ func searchAgentData(hostname string, startDate time.Time, minimumHitsCount int,
 		},
 	}
 
-	indexName := "logs-agent-default"
+	indexName := ".ds-logs-elastic.agent-default-000001"
```
EricDavisX, line 240: we're not supposed to have to use the .ds in order to find things in a search.  is this used only to assert that a new index is being accessed?   I don't know the Data Streams feature well to be 100% sure of usage just confirming.  How is it used?  We can ask @michalpristas to look at the usage and help confirm...

e2e/_suites/ingest-manager/stand-alone.go
mdelapenya: It takes 10 seconds to reach 50 hits

e2e/_suites/ingest-manager/stand-alone.go
mdelapenya: I'd like to know if there is a way to define an alias. For metricbeat we do this: https://github.com/elastic/e2e-testing/blob/a9fe641faf722ac487e9a6423b5bb321ca29d2d2/cli/config/compose/services/metricbeat/docker-compose.yml#L7 We define an index for each metricbeat that is run. Maybe it's not doable in the same way, but I'd love to define an index to be used during the entire test suite.

mdelapenya:
The two existing failures are directly related to https://github.com/elastic/beats/issues/20006

I'm gonna merge this to have the Fleet scenarios on CI. Thanks @EricDavisX for your review!

PR #158, opened 2020-07-03
fix: use proper secret when accessing the docker registry

mdelapenya:
## What does this PR do?
It overrides the default value for getting access to the Docker registry

## Why is it important?
After moving to the Beats CI, its user does not have access to the default secret


mdelapenya:
Closing this because we prefer using defaults

PR #161, opened 2020-07-07
Create agent_endpoint_integration.feature

EricDavisX:
This is an initial start to a smoke test for Endpoint / Agent integration in the e2e test framework we're building out.  The integration should be feasible now if we want to try to code it out, so I am proposing this feature file to start us off.  I hope that the Endpoint team can contribute?!


EricDavisX:
closed in favor of new pr set from proper forked branch

PR #168, opened 2020-07-14
Revert to using the binary to generate integrations files

### LINKED ISSUE #167: e2e tests integrations is failing on CI
The docker-compose files for all integrations that use a `supported-versions.yml` from the Beats repo (apache, mysql, redis) are failing, and the error on CI is because their docker-compose files are not present.

mdelapenya:
## What is this PR doing?
~~It expands the volume used to download the Beats repo, so the entire workspace fir is used, making possible for the integrations compose files to be present in the host.~~

It reverts back all commits to the change that replaced using the binary.

## Why is it important?
~~The sync command was getting compose files out of the container. This was causing the CI, and local execution using the Make command, to not find the compose files for the integrations, causing consistent failures.~~

We realised that using Docker to run Go will cause that any file generated inside the Docker container will have same permissions as inside the container (root). We tried to use same user than in the host inside the container, but we found this error: https://github.com/golang/go/issues/26280.

Applying this code https://github.com/fabric8-services/toolchain-operator/pull/3 fixed that error, but then we found that, because the user was different than root, the Go tool was nott able to write its workspace to $HOME/.op (/root/.op).

We then changed the HOME env var of the container used to run Go to point to /tmp, but that did not work either, because /tmp belongs to root too. 

And that's why we decided to revert all commits affecting the Makefiles to go back to the point we used the binary (previously built) to generate the integrations files.
>Jenkins build discovering files as expected, although **Ingest manager tests are still broken**, because of Kibana not able to start : https://apm-ci.elastic.co/blue/organizations/jenkins/stack%2Fe2e-testing-mbp/detail/PR-168/9/pipeline/256/?start=0

We preferred to revert back to green before merging new changes, expecting better local testing would help with this.

## How to tests this?
Does not apply any more, as we reverted back to old behavior
#### Before applying this changes:
~~1) from project's root dir, execute `make -C cli sync-integrations`~~
~~**Expected behavior:** the ~/.op/compose/services dir contains all (11) integrations (apache, mysql, redis...)~~
~~**Actual behavior:** the ~/.op/compose/services dir contains only metricbeat, which is already bundled in the CLI~~

#### After applying this changes:
~~1) from project's root dir, execute `make -C cli sync-integrations`~~
~~**Expected behavior:** the ~/.op/compose/services dir contains all (11) integrations (apache, mysql, redis...)~~
~~**Actual behavior:** the ~/.op/compose/services dir contains all (11) integrations~~

## Related issues
- Fixes #167
- ~~Caused by~~ Reverting affecting commits back to 3a8837243ce6a7918c520f7117fe8c37cb64b49b

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -41,6 +41,7 @@ pipeline {
       agent { label 'ubuntu-18.04 && immutable && docker' }
       options { skipDefaultCheckout() }
       environment {
+        HOME = "${env.WORKSPACE}"
         GOPATH = "${env.WORKSPACE}"
         GO_VERSION = "${params.GO_VERSION.trim()}"
         PATH = "${env.PATH}:${env.WORKSPACE}/bin:${env.WORKSPACE}/${env.BASE_DIR}/.ci/scripts"
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -41,7 +41,6 @@ pipeline {
       agent { label 'ubuntu-18.04 && immutable && docker' }
       options { skipDefaultCheckout() }
       environment {
-        HOME = "${env.WORKSPACE}"
         GOPATH = "${env.WORKSPACE}"
         GO_VERSION = "${params.GO_VERSION.trim()}"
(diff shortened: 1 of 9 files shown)


mdelapenya:
Converted to draft again while debugging the errors

PR #172, opened 2020-07-15
feat: adds a scheduled pipeline for integrations

### LINKED ISSUE #171: Schedule Integrations daily
cc/ @jsoriano

mdelapenya:
## What does this PR do?
It adds a regular pipeline for the scheduled execution of Integrations.

Besides that, we are bumping to 7.8.0 the default version for stack binaries (ES, Kibana and Metricbeat)

## Why is it important?
Scheduled build for Metricbeat for regression testing.

## Related issues
- Closes #171

## Follow ups
- Notify the integrations team on failures (Slack)


.ci/e2eTestingIntegrationsDaily.groovy
```diff
@@ -0,0 +1,72 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
```
v1v, line 1: I wonder if the daily should use the default values from the downstream job, it might be easy to maintain, what do you think?

mdelapenya: Yes, that would make it easier, indeed. In the case the teams want to execute their dedicated pipeline, should they go to this scheduled, or the downstream one? I added the parameters here trying to cover the first option, but maybe it makes sense to avoid that, and explain how to run the Jenkins general one for each use case. Thoughts?

v1v: I think they should use only the downstream, the daily pipeline is the approach we put in place to make the downstream pipeline a bit simpler when using different build causes in a Multibranch Pipeline, otherwise it might be a bit complicated to exclude when to run for branches, PRs and so on

mdelapenya: I agree. Let me add a commit on top of this PR, thanks!!

mdelapenya: Added!

PR #150, opened 2020-06-30
issues/146 - review Fleet test and sync with stand-alone test

EricDavisX:
https://github.com/elastic/e2e-testing/issues/146
 - per issue I am reviewing and syncing this up to what we checked in for stand-alone agent.  

 - I also added 3 new scenarios, the last of which can be copy / pasted into a new feature file for packages assessment, one for each package we want to assess, tho I don't think we can expect data to show up unless we actually have that software running so we'll have to chop off the last bit of the assessment. 

@new-configuration-new-package
Scenario: Add a new config and a new package and assign an agent
  Given an agent is deployed to Fleet
  When a new configuration named "Test - custom logs" is created
    And the "custom logs" package datasource is added to the "Test - custom logs" configuration
    And the Agent is assigned to the configuration "Test - custom logs"
    And the "Test - custom logs" configuration shows the "custom logs" datasource added
  Then there is new data in the index from agent from "custom logs" stream

PR #174, opened 2020-07-17
feat: enrich the unenroll scenario

mdelapenya:
## What is this PR doing?
It adds an existing step to the unenroll scenario, checking the elastic-agent process is there.

## Why is it important?
We want to ensure the elastic-agent process is still running in the host, as described in the original test script

## Related issues
- Relates #149


mdelapenya:
@EricDavisX, I'm merging this to make progress and continue with #150

PR #178, opened 2020-07-17
Ingest fleet test scenarios updates

mdelapenya:
## What is this PR doing?
It uses the original work in https://github.com/elastic/e2e-testing/pull/150, resolving conflicts after merging it with current master (basically removing the steps using the `...there is data in the index...`, as we do not have further clarification about what this means.

After that, we have refined the steps in the original work, to use a more fluent BDD style. Each commit represents one of those refinements, so please look at each of them to understand each change.

## Why is it important?
Make progress with Fleet mode.

## Related issues
- Superseds #150 

## Follow-ups
I've not modified the scenarios on purpose: I only did a refinement of the steps but without adding/removing the scenarios. I want to discuss about them in this PR, to add comments as a result of the discussion.


e2e/_suites/ingest-manager/features/fleet_mode_agent.feature
```diff
@@ -6,13 +6,9 @@ Feature: Fleet Mode Agent
 Scenario: Deploying an agent
   When an agent is deployed to Fleet
   Then the agent is listed in Fleet as online
-    And system package dashboards are listed in Fleet
-
-@start-agent
```
mdelapenya, line 11: I like this removal, because deploying implicitly means starting an agent. Neat!

e2e/_suites/ingest-manager/fleet.go
```diff
@@ -40,7 +40,15 @@ func (fts *FleetTestSuite) contributeSteps(s *godog.Suite) {
 	s.Step(`^the agent is not listed as online in Fleet$`, fts.theAgentIsNotListedAsOnlineInFleet)
 	s.Step(`^the agent is re-enrolled on the host$`, fts.theAgentIsReenrolledOnTheHost)
 	s.Step(`^the enrollment token is revoked$`, fts.theEnrollmentTokenIsRevoked)
-	s.Step(`^an attempt to enroll a new agent fails$`, fts.anAttemptToEnrollANewAgentFails)
+	s.Step(`^an attempt to enroll an agent with the old token fails$`, fts.anAttemptToEnrollANewAgentFails)
```
mdelapenya, line 43: This step was rephrased, that's why it's here

e2e/_suites/ingest-manager/features/fleet_mode_agent.feature
```diff
@@ -40,5 +36,33 @@ Scenario: Re-enrolling an agent
...
+    And there is new data in the index from agent
+
+@new-configuration-new-package
+Scenario: Add a new config and a new package and assign an agent
+  Given an agent is deployed to Fleet
+  When a new configuration named "Test - custom logs" is created
+    And the "custom logs" package datasource is added to the "Test - custom logs" configuration
+    And the Agent is assigned to the configuration "Test - custom logs"
+  Then the "Test - custom logs" configuration shows the "custom logs" datasource added
+    And there is new data in the index from agent from "custom logs" stream
```
mdelapenya, line 68: I moved L67 from the `When` clause to the `Then` because after reading it, it makes me expect the datasource to be added, so it's seems a verification/assertion. Therefore, a `Then`

e2e/_suites/ingest-manager/features/fleet_mode_agent.feature
```diff
@@ -40,5 +36,33 @@ Scenario: Re-enrolling an agent
...
+    And the Agent is assigned to the configuration "Test Fleet"
+  Then a new enrollment token is created
+    And there is new data in the index from agent
+
+@new-configuration-new-package
+Scenario: Add a new config and a new package and assign an agent
+  Given an agent is deployed to Fleet
+  When a new configuration named "Test - custom logs" is created
+    And the "custom logs" package datasource is added to the "Test - custom logs" configuration
+    And the Agent is assigned to the configuration "Test - custom logs"
```
mdelapenya, line 66: @EricDavisX wdyt about rephrasing this as:

```gherkin
Given an agent is deployed to Fleet
    And a new configuration named "Test - custom logs" includes the "custom logs" package datasource
When the Agent is assigned to the configuration "Test - custom logs"
```

The creation of the configuration and the datasource seems to me more of a preparation step. wdyt?

e2e/_suites/ingest-manager/features/fleet_mode_agent.feature
```diff
@@ -40,5 +36,33 @@ Scenario: Re-enrolling an agent
...
+  Then the "default" configuration shows the "Cisco" datasource added
+
+@new-agent-configuration
+Scenario: Assign an Agent to a new configuration
+  Given an agent is deployed to Fleet
+    And the agent is listed in Fleet as online
+  When a new configuration named "Test Fleet" is created
+    And the Agent is assigned to the configuration "Test Fleet"
+  Then a new enrollment token is created
+    And there is new data in the index from agent
```
mdelapenya, line 59: What about rephrasing this to:

```gherkin
Given a new configuration named "Test Fleet" is created
When an agent is deployed to Fleet
    And the Agent is assigned to the configuration "Test Fleet"
Then a new enrollment token is created
```

I'd remove the `there is new data...` step for the reasons described in the issue: we do not have more information about what to check

EricDavisX: I think we can remove this and re-add it when possible... I do however think that every few weeks I'm learning more about how to assess the data flowing into ES.  I think we can craft a better query now, adding in the source of the data (being either of Metricbeat, Filebeat, or Agent) to assert on this.  At a high level we don't care which is sending data, but a good short term goal would be to assert each of them has at least sent 1 document.  Agent may not send much more than that, unless there are problems.

e2e/_suites/ingest-manager/features/fleet_mode_agent.feature
EricDavisX: yes, i agree - please go with that, its good.  👍

e2e/_suites/ingest-manager/features/fleet_mode_agent.feature
mdelapenya: Ok, let's use this new scenario. We can create an issue to add the step for checking specific data. We can use it to fill in with your research

mdelapenya: Is it required that the agent is listed on fleet for this scenario?

EricDavisX:
Manu had asked me 3 specific questions about the underlying Ingest Manager APIs that drive these 3 steps:
  When the "Cisco" "latest" package version is installed successfully
    And the "Cisco" package datasource is added to the "default" configuration
  Then the "default" configuration shows the "Cisco" datasource added

I'd like to propose an edit and I will then answer the questions 3:
Let's change "Cisco" to 'Custom logs' because unless we're really integrating with a given software (Cisco in this case) then it can have unexpected integration errors and is better avoided until it can be properly tested (a separate discussion ticket is being started for that...).

1) installing an Integration like 'Custom logs' can be achieved as calling this:
POST (with empty body) to /api/ingest_manager/epm/packages/<package_name>-<desired_package_version>
which in our case will be something like: log-0.3.4

 - I propose that we can use a manual version citation as a possibility, but presuming that "latest" is found in the step that we'll call the below scripting to get the actual version number.  This is best, because the package can be updated at any time and we'll want to keep the test running, and running with the latest without any manual input.  The scripting to get the version of the package is this:

GET api/ingest_manager/epm/packages?experimental=true&category=

and capture the "version" field from the returned "response[]" list where the "name" field is 'log' for that list item. see screenshot:
logs-version-retrieval

we can assert after we GET the packages list but before we install 'logs' that the 'logs' item includes the field "status": "not_installed"

we can assert after we GET the packages list and after we have installed the Integration that the 'logs' item includes the field "status": "installed"

item 2) Regarding: And the "Custom log" integration is added to the "default" configuration

it can be achieved with:
POST /api/ingest_manager/package_configs
with body below (some parameterization is required).  I suggest we leave the body as an input to the call so we can easily install other Integrations, pls.  Each Integration has zero, 1 or many fields that will need to be specified in the body so we can leave it open ended and as simple as we want to for now, as a text blob perhaps?

body:
{"name":"e2e-test","description":"e2e-test description","namespace":"default","config_id":"8e8de190-d72b-11ea-8faf-d5798ef2bf04","enabled":true,"output_id":"","inputs":[{"type":"logfile","enabled":true,"streams":[{"id":"logfile-log.log","enabled":true,"dataset":{"name":"log.log","type":"logs"},"vars":{"paths":{"type":"text","value":["\\var\\log\\e2e-test.log"]},"dataset":{"value":"generic","type":"text"},"custom":{"value":"","type":"yaml"}}}]}],"package":{"name":"log","title":"Custom logs","version":"0.3.4"}}

The 'config_id' we should have documented already, this is the id of the 'default config'.
The 'version' is the same version as noted above.
The 3 separate text inputs for this Integration are the name, description and value.
The value of "\\var\\log\\e2e-test.log" is a location we can update on the agent host as a test step addition to validate the Integration is able to send data, specially that a corresponding document is added to ES.  Shall we add a 4th test step now as the below?

  And the "Custom logs" Integration is validated to "send" data to ES
(search query to be posted separately, follow up [1] if we decide to)

And if we're going to take it that far, we ideally would also add a pair of test steps that remove the Integration (scripting posted separately, follow up [2]) and validate that it does 'not send' data into ES as:
  And the "Custom logs" Integration is validated to "not send" data to ES

We can leave this as a separate ticket.  Some small tickets are good to put in for team members to use to get familiar with the framework.

Item 3)
  Then the "default" configuration shows the "Cisco" datasource added
we should update this to be:
  Then the "default" configuration shows the "Custom logs" Integration added

This can be asserted to be found in the config by calling:
GET on /api/ingest_manager/agent_configs/<default-configuration-id>
and checking that the 2nd item in the response is the 'Custom logs' integration added, at this data location in the response:
"item": { "package_configs": [1] { "name": "e2e-test"

lets also validate the description in the same location
"item": { "package_configs": [1] { "description": "e2e-test description"
there are more asserts we could do on the actual data, lets leave it as such with this for now.  Some of this is changing here at the end of 7.9 which is another reason I don't want to add a test that is imminently going to be 'wrong' 

screenshot:
config-api-response-data

e2e/_suites/ingest-manager/features/fleet_mode_agent.feature
```diff
@@ -40,5 +36,29 @@ Scenario: Re-enrolling an agent
...
+
+@package-added-to-default-config
+Scenario: Adding packages to configurations
+  Given an agent is deployed to Fleet
+    And the package list service is available
+  When the "Cisco" "latest" package version is installed successfully
+    And the "Cisco" package datasource is added to the "default" configuration
+  Then the "default" configuration shows the "Cisco" datasource added
+
+@new-agent-configuration
```
EricDavisX, line 52: this is covered now in the agent-endpoint section, not needed to be repeated here when we resolve conflicts - we can just delete this scenario

e2e/_suites/ingest-manager/features/fleet_mode_agent.feature
```diff
@@ -40,5 +36,29 @@ Scenario: Re-enrolling an agent
...
+Scenario: Assign an Agent to a new configuration
+  Given a new configuration named "Test Fleet" is created
+  When an agent is deployed to Fleet
+    And the Agent is assigned to the configuration "Test Fleet"
+  Then a new enrollment token is created
+
+@new-configuration-new-package
+Scenario: Add a new config and a new package and assign an agent
```
EricDavisX, line 60: this is covered as well with agent-endpoint integration file we got merged recently.  additionally, all tests are using a new agent policy (config), specific to the scenario so we don't need ot test this as it is is written.  we are going down the path of having at least 1 implied policy config for each Integration, if and when we want to make it more configurable it will likely still need to be enumerated as such because the policy JSON to post is a lot to try to manage as a step in the scenario

e2e/_suites/ingest-manager/features/fleet_mode_agent.feature
```diff
@@ -40,5 +36,29 @@ Scenario: Re-enrolling an agent
...
 Scenario: Revoking the enrollment token for an agent
   Given an agent is deployed to Fleet
+    And the agent is un-enrolled
+    And the "agent" process is "stopped" on the host
   When the enrollment token is revoked
-  Then an attempt to enroll a new agent fails
+  Then an attempt to enroll an agent with the old token fails
+
+@package-added-to-default-config
+Scenario: Adding packages to configurations
```
EricDavisX, line 45: we've decided that the default policy ( configuration ) is the same as any newly created configuration so this test is not needed in that angle, if we do want to try to set up individual Integrations, we can certainly do that - but would be better to file them to their own individual file(s).  I'd say we can leave this out for now.

e2e/_suites/ingest-manager/features/fleet_mode_agent.feature
EricDavisX: yeah, i still like this too - i think I'll rebase on my branch and push up the remaining changes that still make sense and aren't already implemented (most of it is).  when done we can close this out I think. will post back

EricDavisX (changes requested):
its so out of date I think its easier to re-base and re-work the 2 (first 2) items that are still needed.  i'll do a pr for it.

EricDavisX:
in the end there was only one minor refactor I think we really need to keep from here, pushed to https://github.com/elastic/e2e-testing/pull/312
 - Manu, when you're free (finally) of other urgent tasks, we can review this and likely close it I think?  If you have other code we want to keep, by all means please do push it on, after rebasing, etc.

EricDavisX:
laying it to rest.  it shall be worked no more.  lol

PR #179, opened 2020-07-17
feat: download the agent once, mounting it as a volume

### LINKED ISSUE #177: Download elastic-agent once
We'd like to download it from the test runner instead of doing it inside the container, then we would copy it to the running container when deploying the elastic-agent to fleet.

This will add two things:
1) reliability, as we will download the file once, and possibly using retries, which is not possible (or more difficult) to do it in the container
2) reduce build time: it takes around 7 seconds to download the file in my local machine. Simply counting the times the Deploy agent step is used (6 times for fleet at this moment) will give us the build time reduction.

mdelapenya:
## What is this PR doing?
It changes the compose file for the vanilla box defining a volume to mount the elastic-agent binary on it. We will use environment variables to configure the src and target of the volume.

That way, we will download the agent binary to a temporary dir before the test suite runs (only once), and will remove it from the file system after the tests suite runs.

## Why is this important?
It will reduce the number of HTTP connections to download the binary to 1, compared to one for each scenario installing an agent. It usually takes from 5 to 10 seconds, so the gain will be (in average) 7'5 times the number of scenarios (6 at this moment).

It will also bring reliability, because the download of the binary is done from the test runner, so it has the retry-with-backoff capabilities, instead of running it from inside the container.

## Related issues
- Closes #177

### DIFF
--- a/e2e/_suites/ingest-manager/fleet.go
+++ b/e2e/_suites/ingest-manager/fleet.go
@@ -24,12 +24,14 @@ const ingestManagerDataStreamsURL = kibanaBaseURL + "/api/ingest_manager/data_st
 
 // FleetTestSuite represents the scenarios for Fleet-mode
 type FleetTestSuite struct {
-	EnrolledAgentID string // will be used to store current agent
-	BoxType         string // we currently support Linux
-	Cleanup         bool
-	ConfigID        string // will be used to manage tokens
-	CurrentToken    string // current enrollment token
-	CurrentTokenID  string // current enrollment tokenID
+	AgentDownloadName string // the name for the binary
+	AgentDownloadPath string // the path where the agent for the binary is installed
+	EnrolledAgentID   string // will be used to store current agent
+	BoxType           string // we currently support Linux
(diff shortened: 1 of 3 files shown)


e2e/_suites/ingest-manager/fleet.go
```diff
@@ -83,6 +85,25 @@ func (fts *FleetTestSuite) anAgentIsDeployedToFleet() error {
 	return err
 }
 
+// downloadAgentBinary it downloads the binary and stores the location of the downloaded file
+// into the Fleet struct, to be used else where
+func (fts *FleetTestSuite) downloadAgentBinary() error {
+	artifact := "elastic-agent"
+	version := "8.0.0-SNAPSHOT"
```
mdelapenya, line 92: This version should be configurable, to support running the test suite against multiple versions of the agent.

I could think of an env var with the URL. If it is present in the execution, simply use it.

mdelapenya:
Failing tests are the ones failing in upstream

PR #181, opened 2020-07-20
fix: use the status of the agent by hostname

mdelapenya:
## What is this PR doing?
It uses the hostname to get the status of the agent, because the hostname value is used as agentID. That way we are able to look up the status of the agent in the response of the API call, which returns a JSON with the list of agents including their active status.

We also revoke/remove the token that is created for the enrollment process (we create it on each scenario)

## Why is it important?
It moves from a testing approach that is based in the existence of only one agent, to an approach that is based in the status of a specific agent (identified by hostname). We need this change because the API should be able to unenroll/remove an agent whenever the container representing it is removed (docker-compose down), and this is not currently happening.

## Related issues
- Relates https://github.com/elastic/beats/issues/20006
- Relates https://github.com/elastic/kibana/pull/72386

### DIFF
--- a/e2e/_suites/ingest-manager/fleet.go
+++ b/e2e/_suites/ingest-manager/fleet.go
@@ -140,23 +140,23 @@ func (fts *FleetTestSuite) setup() error {
 func (fts *FleetTestSuite) theAgentIsListedInFleetAsOnline() error {
 	log.Debug("Checking agent is listed in Fleet as online")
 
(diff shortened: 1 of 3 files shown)


mdelapenya:
The failed tests complains about Go lint in a multi module project, so it's not related.

mdelapenya:
Merging, as the tests for fleet are now green!

image

PR #183, opened 2020-07-20
fix: do not check for modifications in the downstream job

### LINKED ISSUE #182: Scheduled jobs are not triggering the e2e tests because it's not finding modifications in the code
As it should be, but we added the `forceSkipGitChecks` param to avoid this.

mdelapenya:
## What does this PR do?
It forces the downstream job to skip for git changes

## Why is it important?
Because otherwise the triggered job, which uses master, won't discover changes in the execution, not triggering the e2e jobs. I.e. https://beats-ci.elastic.co/blue/organizations/jenkins/e2e-tests%2Fe2e-testing-mbp/detail/master/69/pipeline/173/

## Related issues
- Fixes #182

### DIFF
--- a/.ci/e2eTestingHelmDaily.groovy
+++ b/.ci/e2eTestingHelmDaily.groovy
@@ -42,6 +42,7 @@ pipeline {
       steps {
         build(job: 'e2e-tests/e2e-testing-mbp/master',
           parameters: [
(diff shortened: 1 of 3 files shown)


mdelapenya:
Merging, as I'd like to have this running for today's nightly schedule

PR #190, opened 2020-07-21
chore(ci): support overriding the Elastic Agent binary to be used on CI

### LINKED ISSUE #188: Support parametrising the elastic-agent binary on CI
It's possible to change the binary to be used: if the `ARTIFACT_HASH` env var is present, the test framework will use it to calculate the URL of the binary, based on the version.

```go
// GetElasticArtifactURL returns the URL of a released artifact from
// Elastic's artifact repository, bbuilding the JSON path query based
// on the desired OS, architecture and file extension
// i.e. GetElasticArtifactURL("elastic-agent", "8.0.0-SNAPSHOT", "linux", "x86_64", "tar.gz")
// If the environment variable ARTIFACT_HASH exists, then the artifact to be downloaded will
// be defined by that value
func GetElasticArtifactURL(artifact string, version string, OS string, arch string, extension string) (string, error) {
	hash := os.Getenv("ARTIFACT_HASH")
	if hash != "" {
		hashedVersion := strings.ReplaceAll(version, "-SNAPSHOT", "")
		return fmt.Sprintf("https://snapshots.elastic.co/%s-%s/downloads/beats/%s/%s-%s-%s-%s.%s", hashedVersion, hash, artifact, artifact, version, OS, arch, extension), nil
	}
...
}
```

If the variable is not present, a cURL request to `https://artifacts-api.elastic.co/v1/search/$version/$artifact` is used to get the package URL to be downloaded.

mdelapenya:
## What does this PR do?
This PR adds a new parameter to the Jenkinsfile, which will set an environment variable to the pipeline. This new parameter will allow the user to configure which version of the Agent will be installed during a build. By default, this variable will be empty, which means the tests will use the agent for the current version (i.e. 8.0.0-SNAPSHOT)

In this PR we are refactoring the logic to get that variable, and instead of partially expecting a hash and build the download URL from it, we expect the entire download URL to be configurable.

## Why is it important?
This flexibility in the definition of the download URL will help in executing the job with multiple values, also simplifying the calculation of the URL.

Besides that, triggering a job with a specific version of the agent will be simpler, as it would be just a matter of getting it from our artifact system.

## Checklist

~~- [ ] I have commented my code, particularly in hard-to-understand areas~~
~~- [ ] I have made corresponding change to the default configuration files~~
~~- [ ] I have added tests that prove my fix is effective or that my feature works~~
~~- [ ] I have noticed new Go dependencies (run `make notice` in the proper directory)~~

Run tests with this new environment variable:
```shell
$ export ELASTIC_AGENT_DOWNLOAD_URL=https://snapshots.elastic.co/8.0.0-59098054/downloads/beats/elastic-agent/elastic-agent-8.0.0-SNAPSHOT-linux-x86_64.tar.gz
$ godog -t fleet_mode
```

## Related issues

- Closes #188 

## Use cases

Running the tests with different released versions of the agent will be possible.

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -26,7 +26,7 @@ pipeline {
   parameters {
     choice(name: 'runTestsSuite', choices: ['all', 'helm', 'ingest-manager', 'metricbeat'], description: 'Choose which test suite to run (default: all)')
     booleanParam(name: "forceSkipGitChecks", defaultValue: false, description: "If it's needed to check for Git changes to filter by modified sources")
-    string(name: 'ELASTIC_AGENT_HASH', defaultValue: '', description: 'If present, it will override the released version of the Elastic agent artifact.')
+    string(name: 'ELASTIC_AGENT_DOWNLOAD_URL', defaultValue: '', description: 'If present, it will override the download URL for the Elastic agent artifact. (I.e. https://snapshots.elastic.co/8.0.0-59098054/downloads/beats/elastic-agent/elastic-agent-8.0.0-SNAPSHOT-linux-x86_64.tar.gz')
     choice(name: 'LOG_LEVEL', choices: ['INFO', 'DEBUG'], description: 'Log level to be used')
     choice(name: 'RETRY_TIMEOUT', choices: ['3', '5', '7', '11'], description: 'Max number of minutes for timeout backoff strategies')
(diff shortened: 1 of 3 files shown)


mdelapenya:
Failing test not related... Will merge as the changes in this PR affect to the ingest-manager test suite, only

PR #193, opened 2020-07-21
feat: install test dependencies for each test suite

### LINKED ISSUE #170: Support each test suite to contribute the installation of its tests dependencies
We are installing kubectl, helm, kind, etc across all test suites, when we need different one per suite.

An approach could be generate a dedicated shell script per suite, so the automation already in place would execute the one for the test suite.

mdelapenya:
## What does this PR do?
When installing the test dependencies, it checks for the existence of a specialised script for installing the dependencies for the current test suite, identified by the $SUITE variable. This variable will be set on CI.

## Why is it important?
It will reduce the build time in those parallel branches that don't need to install some test deps.

## Checklist

## Related issues

- Closes #170

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -205,7 +205,7 @@ def generateFunctionalTestStep(Map params = [:]){
         unstash 'build'
         retry(3){
           dir("${BASE_DIR}"){
-            sh script: """.ci/scripts/install-test-dependencies.sh "${GO_VERSION}" """, label: "Install test dependencies for ${suite}:${feature}"
+            sh script: """.ci/scripts/install-test-dependencies.sh "${GO_VERSION}" "${suite}" """, label: "Install test dependencies for ${suite}:${feature}"
           }
         }
         dir("${BASE_DIR}"){
--- a/.ci/scripts/install-test-dependencies.sh
+++ b/.ci/scripts/install-test-dependencies.sh
@@ -22,4 +22,9 @@ SUITE=${2:?SUITE is not set}
 source .ci/scripts/install-go.sh "${GO_VERSION}"
 
 # execute specific test dependencies if it exists
-[ -f .ci/scripts/install-${SUITE}-dependencies.sh ] && .ci/scripts/install-${SUITE}-dependencies.sh
+if [ -f .ci/scripts/install-${SUITE}-dependencies.sh ]
+then
+    .ci/scripts/install-${SUITE}-dependencies.sh
+else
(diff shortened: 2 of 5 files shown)


.ci/scripts/install-test-dependencies.sh
```diff
@@ -10,42 +10,19 @@ set -euxo pipefail
...
-curl -o ${HELM_TAR_GZ_FILE} "https://get.helm.sh/${HELM_TAR_GZ_FILE}"
-tar -xvf ${HELM_TAR_GZ_FILE}
-mv linux-amd64/helm ${HELM_CMD}
-chmod +x "${HELM_CMD}"
-${HELM_CMD} version --client
-rm -fr linux-amd64 ${HELM_TAR_GZ_FILE}
+# execute specific test dependencies if it exists
+if [ -f .ci/scripts/install-${SUITE}-test-dependencies.sh ]
+then
+    .ci/scripts/install-${SUITE}-test-dependencies.sh
```
v1v, line 25: I'd say to add `chmod 755 .ci/scripts/install*` of something similar in case those files got some permission issues for whatever reason, or using source as an alternative

.ci/Jenkinsfile
```diff
@@ -205,11 +205,11 @@ def generateFunctionalTestStep(Map params = [:]){
         unstash 'build'
         retry(3){
           dir("${BASE_DIR}"){
-            sh script: """.ci/scripts/install-test-dependencies.sh "${GO_VERSION}" """, label: "Install test dependencies for ${suite}:${feature}"
+            sh script: """.ci/scripts/install-test-dependencies.sh "${GO_VERSION}" "${suite}" """, label: "Install test dependencies for ${suite}:${feature}"
           }
         }
         dir("${BASE_DIR}"){
-          sh script: """.ci/scripts/functional-test.sh "${GO_VERSION}" "${suite}" "${feature}" "${STACK_VERSION}" "${METRICBEAT_VERSION}" """, label: "Run functional tests for ${suite}:${feature}"
```
mdelapenya, line 212: I'm removing the GO_VERSION from the `functional-test.sh` script because it is used to install Go, which is already done in the `install-test-dependencies.sh` script. And as the CI executes both in a decoupled way, I consider the 1st script is not self-contained. As a follow-up, we could include the `install-test-dependencies` script into the 1st script, becoming self-contained.

mdelapenya: Mmm, I think I see why both scripts are executed in a decoupled manner: to support retries installing things. We'd need then that the Go binary is still available to the second execution.

PR #191, opened 2020-07-21
docs: enhance documentation

mdelapenya:
## What does this PR do?
This PR is simplifies the existing docs, creating a main README file per test suite, moving some of the sections to other docs, and removing others because they are outdated or even present in a more up-to-date doc.

~~Finally, we added the CI badges for the scheduled pipeline for each test suite~~~ Edited: See https://github.com/elastic/e2e-testing/pull/191#discussion_r459612392 

## Why is it important?
CCC: Clearness, Conciseness and Concreteness

We are also preparing the project for transferring it to the functional teams

## Checklist

~~- [ ] My code follows the style guidelines of this project~~
~~- [ ] I have commented my code, particularly in hard-to-understand areas~~
- [x] I have made corresponding changes to the documentation
~~- [ ] I have made corresponding change to the default configuration files~~
~~- [ ] I have added tests that prove my fix is effective or that my feature works~~
~~- [ ] I have run the Unit tests for the CLI, and they are passing locally~~
~~- [ ] I have run the End-2-End tests for the suite I'm working on, and they are passing locally~~
~~- [ ] I have noticed new Go dependencies (run `make notice` in the proper directory)~~

## Related issues

- Closes #185 
- Merge after #194

PR #194, opened 2020-07-22
chore(ci): use withGoEnv build step to install Go

### LINKED ISSUE #189: Use GVM in the CI, as it's already bundled in the build agent

mdelapenya:
## What does this PR do?
This PR removes the script to install Go. Instead, we use the existing `withGoEnv` build step from the shared library, which installs Go using GVM (already bundled in the CI workers) and apply the proper environment variables needed by Go (GOPATH and GOROOT).

The removal of the script caused the removal of the GO_VERSION variable from every shell script. Instead, we are reading the Go version from a single file in the workspace, a `.go-version` file, applying that value everywhere.
>I've not updated the docs using the environment variable to avoid merge conflicts on #191  

## Why is it important?
The first goal is to reuse existing building blocks for the pipeline, without reinventing the wheel every time. The second one is about simplicity, as we remove a dependency on an input parameter for the GO_VERSION: TBH, we are not going to run the test framework in different versions of Go; we are going to use the one set at the project level.

## Checklist

~~- [ ] My code follows the style guidelines of this project~~
~~- [ ] I have commented my code, particularly in hard-to-understand areas~~
~~- [ ] I have made corresponding changes to the documentation~~
~~- [ ] I have made corresponding change to the default configuration files~~
~~- [ ] I have added tests that prove my fix is effective or that my feature works~~
~~- [ ] I have run the Unit tests for the CLI, and they are passing locally~~
~~- [ ] I have run the End-2-End tests for the suite I'm working on, and they are passing locally~~
~~- [ ] I have noticed new Go dependencies (run `make notice` in the proper directory)~~

## Related issues

- Closes #189


.ci/Jenkinsfile
```diff
@@ -71,17 +67,21 @@ pipeline {
...
               steps {
                 withGithubNotify(context: 'Sanity checks', tab: 'tests') {
                   deleteDir()
                   unstash 'source'
-                  dir(BASE_DIR){
-                    sh script: """.ci/scripts/install-dependencies.sh "${GO_VERSION}" """, label: 'Install dependencies'
-                    preCommit(commit: "${GIT_BASE_COMMIT}", junit: true)
+                  script {
+                    env.GO_VERSION = readFile("${env.WORKSPACE}/${env.BASE_DIR}/.go-version").trim()
+                  }
```
kuisathaverat, line 79:
```suggestion
                 setEnvVar("GO_VERSION", readFile("${env.WORKSPACE}/${env.BASE_DIR}/.go-version").trim())
```

kuisathaverat: if you out the setEnvVaron the `Checkout` stage you do not have to set it anymore.

mdelapenya: Grooming time! I'll refactor it as you suggest!

image

PR #202, opened 2020-07-27
docs: enhance documents about docker errors

### LINKED ISSUE #201: Document docker issues for troubleshooting local errors

mdelapenya:
## What does this PR do?

We are explicitly adding Docker configurations that are needed on Mac, plus some minor improvements in the existing docs.

## Why is it important?
Improve the user experience while using this framework

## Related issues

- Closes #201

Paste here output logs discovered while creating this PR, such as stack traces or integration logs, or any other output you consider important to be shared with the team.
-->

Add here any thought that you consider could be identified as an actionable step once this PR is merged.
-->


mdelapenya:
> What If running the commands with `TMP=$(pwd)/.tmp <command>` then the temporary folder location is changed on the fly without the need to configure the mac for docker setup?

I think this will work! Although I would prefer a more standard execution, not adding another variable to the party. Nevertheless, I think we can cover this in a follow-up, if we see the above option very intrusive

PR #205, opened 2020-07-28
feat: support using Beats CI snapshots

### LINKED ISSUE #204: Move from release artifacts to those generated by the packaging job in Beats
We create artifacts for every merge to master, and for PRs we are able to generate the artifacts on demand.

Consuming these artifacts will bring more reliability, as the artifacts will contain in current code, not depending on other teams.

cc/ @elastic/observablt-robots @EricDavisX @ph

mdelapenya:
## What does this PR do?
It supports downloading the elastic-agent binaries from the Google Cloud Storage bucket used by Beats CI on every merge-to-master. For that, as the objects in the bucket are not public, we are listing bucket's contents, looking for a match for the agent we want to use in tests.

To enable this feature, set the `ELASTIC_AGENT_USE_CI_SNAPSHOTS` environment variable to `true`, and the tests will not fetch the binaries from the default location.

Besides, we are using the Docker image for all Elastic components from Elastic's docker registry, using the observability-ci namespace. This namespace contains snapshots from each merge-to-master, and all PRs for Beats.

To enable this feature, set the `ELASTIC_AGENT_STAND_ALONE_VERSION` environment variable to a valid and puushed tag i.e. `78a762c76080aafa34c52386341b590dac24e2df` or eventually `pr-20323`, and the tests will not fetch the default image (8.0.0-SNAPSHOT).

Finally, we are arranging the existing env vars into a different name convention, prefixing them with the test suite that the var belongs to, removing the `OP_` prefix when needed.

## Why is it important?
The release process for the binaries is not as reliable as we need, as it could take days until we have a new snapshot, blocking this test suite. For that reason we want to support a way to switch to another location if the official binaries are not working or does not contain latest code or fixes.

About the small refactors, consistent names that are easy to read are names easy to understand, that's the motivation for that change.

## Checklist

## Author's Checklist

- [x] Executed e2e tests with and without the new variable
- [x] Verified that refactored code keeps existing behavior

#### Using the latest snapshot
```shell
$ OP_LOG_LEVEL=DEBUG ELASTIC_AGENT_USE_CI_SNAPSHOTS=true godog -t "fleet_mode && unenroll"
```

You should see something similar to these log entries, related to Google Cloud Storage APIs:
>DEBU[0088] Google Cloud Storage API is available         bucket=beats-ci-artifacts elapsedTime=1.529925991s object=elastic-agent-8.0.0-SNAPSHOT-linux-x86_64.tar.gz retries=1 statusEndpoint="https://storage.googleapis.com/storage/v1/b/beats-ci-artifacts/o"
DEBU[0088] Downloading file                              path="/var/folders/wn/zl67dnhn27s0jkdlgw4bbts40000gn/T/snapshots%2Felastic-agent-8.0.0-SNAPSHOT-linux-x86_64.tar.gz?generation=1594737904872374&alt=media569500323" url="https://storage.googleapis.com/download/storage/v1/b/beats-ci-artifacts/o/snapshots%2Felastic-agent-8.0.0-SNAPSHOT-linux-x86_64.tar.gz?generation=1594737904872374&alt=media"
DEBU[0088] File downloaded                               elapsedTime=752.260024ms path="/var/folders/wn/zl67dnhn27s0jkdlgw4bbts40000gn/T/snapshots%2Felastic-agent-8.0.0-SNAPSHOT-linux-x86_64.tar.gz?generation=1594737904872374&alt=media569500323" retries=1 url="https://storage.googleapis.com/download/storage/v1/b/beats-ci-artifacts/o/snapshots%2Felastic-agent-8.0.0-SNAPSHOT-linux-x86_64.tar.gz?generation=1594737904872374&alt=media"

On the other hand, if not setting that variable, or setting to something not true:
```shell
$ OP_LOG_LEVEL=DEBUG godog -t "fleet_mode && unenroll"
```
You should see something similar to these logs entries, related to the official artifacts endpoint:
>DEBU[0088] The Elastic artifacts API is available        elapsedTime=784.608597ms retries=1 statusEndpoint="https://artifacts-api.elastic.co/v1/search/8.0.0-SNAPSHOT/elastic-agent"
DEBU[0088] Downloading file                              path=/var/folders/wn/zl67dnhn27s0jkdlgw4bbts40000gn/T/elastic-agent-8.0.0-SNAPSHOT-linux-x86_64.tar.gz697951930 url="https://snapshots.elastic.co/8.0.0-f762d16e/downloads/beats/elastic-agent/elastic-agent-8.0.0-SNAPSHOT-linux-x86_64.tar.gz"

#### Stand-alone mode: using another Docker tag

```shell
$ OP_LOG_LEVEL=DEBUG ELASTIC_AGENT_STAND_ALONE_VERSION=78a762c76080aafa34c52386341b590dac24e2df godog -t "deploy-stand-alone"
```
You should see something similar to these logs entries, related to the docker-compose pull command fetching the tag you selected:
>Pulling elastic-agent (docker.elastic.co/observability-ci/elastic-agent:78a762c76080aafa34c52386341b590dac24e2df)...

## Related issues

- Closes #204
- Relates https://github.com/elastic/beats/issues/20274

## Follow-ups
An interesting follow-up will be consuming the snapshot for a PR, instead of the merge-to-master snapshot. See https://github.com/elastic/beats/issues/20274 and https://github.com/elastic/beats/pull/20323 for related work on Beats side.


cli/shell/shell.go
```diff
@@ -54,6 +56,42 @@ func Execute(workspace string, command string, args ...string) (string, error) {
...
+	}
+
+	return defaultValue
+}
+
+// GetEnvBool returns an environment variable as boolean
+func GetEnvBool(key string) bool {
+	s := os.Getenv(key)
+	if s == "" {
+		return false
```
cachedout, line 72: I kind of wonder if this should return something else here instead of `false`. (Perhaps `nil`?) It seems like this behavior might introduce hard-to-handle error conditions, where a value is incorrectly interpreted as being truly false instead of an error in processing the lookup itself.

e2e/utils.go
```diff
@@ -150,6 +141,72 @@ func GetElasticArtifactURL(artifact string, version string, OS string, arch stri
 	return downloadURL, nil
 }
 
+// GetObjectURLFromBucket extracts the media URL for the desired artifact from the
+// Google Cloud Storage bucket used by the CI to push snapshots
+func GetObjectURLFromBucket(bucket string, object string) (string, error) {
+	exp := GetExponentialBackOff(1 * time.Minute)
```
cachedout, line 147: It seems surprising to multiply by 1. :) What's the reason for this?

mdelapenya: It's because the method needs a time.Time argument, and we want to ensure it belongs to the order of minutes. If this is not clear, we can make it more explicit to 60 seconds. Any suggestion is welcome!

cli/shell/shell.go
mdelapenya: Mmmm, I see your point: so if the env var is not present, then do not consider it as false... In this particular use case, we are going to check for true, only. If true, then execute the proper code, otherwise use default. But it could be the case that we use this function again checking for false, and if the var is not present, it will satisfy the condition

mdelapenya: I think returning (bool, error) will do the trick, although we'll have to handle errors on the consume side.

mdelapenya: I added your suggestion in https://github.com/elastic/e2e-testing/pull/205/commits/3881a99b251b92ee8fd52f05acd8c9e4ed9c14bc, including unit tests for verification.

mdelapenya:
@EricDavisX with these changes now we will be able to run tests with the artifacts for a PR :)

e2e/utils.go
mdelapenya: I added 7038847e56430337aebc7299b24dab36828a77d5, removing the need of multiplying by the neutral element: simply using the default duration. Thanks for this improvement!

.ci/Jenkinsfile
```diff
@@ -28,15 +30,17 @@ pipeline {
...
-    string(name: 'STACK_VERSION_METRICBEAT', defaultValue: '7.8.0', description: 'SemVer version of the stack to be used for Metricbeat tests.')
+    string(name: 'INGEST_MANAGER_STACK_VERSION', defaultValue: '8.0.0-SNAPSHOT', description: 'SemVer version of the stack to be used for Ingest Manager tests.')
+    string(name: 'METRICBEAT_STACK_VERSION', defaultValue: '7.8.0', description: 'SemVer version of the stack to be used for Metricbeat tests.')
     string(name: 'METRICBEAT_VERSION', defaultValue: '7.8.0', description: 'SemVer version of the metricbeat to be used.')
     string(name: 'HELM_CHART_VERSION', defaultValue: '7.6.1', description: 'SemVer version of Helm chart to be used.')
     string(name: 'HELM_VERSION', defaultValue: '2.16.3', description: 'SemVer version of Helm to be used.')
-    string(name: 'KIND_VERSION', defaultValue: '0.7.0', description: 'SemVer version of Kind to be used.')
-    string(name: 'KUBERNETES_VERSION', defaultValue: '1.15.3', description: 'SemVer version of Kubernetes to be used.')
+    string(name: 'HELM_KIND_VERSION', defaultValue: '0.7.0', description: 'SemVer version of Kind to be used.')
```
kuisathaverat, line 42:
```suggestion
    string(name: 'HELM_KIND_VERSION', defaultValue: '0.8.1', description: 'SemVer version of Kind to be used.')
```

.ci/scripts/install-helm-test-dependencies.sh
```diff
@@ -10,29 +10,29 @@ set -euxo pipefail
...
 #
 
 MSG="parameter missing."
 HOME=${HOME:?$MSG}
 
 HELM_VERSION="${HELM_VERSION:-"2.16.3"}"
 HELM_TAR_GZ_FILE="helm-v${HELM_VERSION}-linux-amd64.tar.gz"
-KIND_VERSION="v${KIND_VERSION:-"0.7.0"}"
-KUBERNETES_VERSION="${KUBERNETES_VERSION:-"1.15.3"}"
+HELM_KIND_VERSION="v${HELM_KIND_VERSION:-"0.7.0"}"
```
kuisathaverat, line 22:
```suggestion
HELM_KIND_VERSION="v${HELM_KIND_VERSION:-"0.8.1"}"
```

.ci/Jenkinsfile
mdelapenya: Could we send this bump in another PR, so that we do not add side-effects to it?

mdelapenya: I created https://github.com/elastic/e2e-testing/issues/207 to track progress

PR #219, opened 2020-08-18
chore: fix gherkin lint

mdelapenya:
## What does this PR do?
We added a Make target to run gherkin/lint locally, disabling some rules that we consider less useful.

These rules have been added as arguments to the pre-commit stage, so that it will happen there too.

We also refactored a bit the script to lint Go code, for the sake of reusability 

## Why is it important?
Refinement of the scenarios, following best practices

## Checklist

Run this make target:

```shell
cd e2e
make lint
```

The number of BDD warning/errors is very reduced now.


e2e/Makefile
```diff
@@ -41,6 +41,10 @@ functional-test: install-godog
 	STACK_VERSION=${STACK_VERSION} \
 	godog --format=${FORMAT} ${FEATURE_FLAG} ${FEATURE}
 
+.PHONY: lint
+lint:
+	@docker run -ti --rm -v $(PWD):/src -w /src gherkin/lint **/*.feature --disable AvoidOutlineForSingleExample,TooClumsy,TooManySteps,TooManyDifferentTags,TooLongStep
```
v1v, line 46: TTY is not enabled in the CI workers:

```suggestion
	@docker run -t --rm -v $(PWD):/src -w /src gherkin/lint **/*.feature --disable AvoidOutlineForSingleExample,TooClumsy,TooManySteps,TooManyDifferentTags,TooLongStep
```

v1v, line 46: Will it be possible to run the pre-commit in the CI? If so, we don't need this particular make goal

mdelapenya: Yes, it's still possible to run pre-commit in the CI, I wanted to have an easy manner to run it locally, setting the exclusions and getting a better output:

Screenshot 2020-08-18 at 18 24 33

PR #223, opened 2020-08-21
add slack post for successful jobs too

EricDavisX:
This is a change the the reporting / notification to slack when a job passes.  If it fails, we already have notification, but I realize that passing tests should be celebrated and shown off (and keep folks in the habit of thinking of them, even if they pass all the time)

## What does this PR do?

Changes the Jenkins file with a 'success' section.

## Why is it important?
Because it allows for better visibility of test execution

## Checklist

## Author's Checklist
I can admit I don't know how to test this.  Team, can you help confirm?  I suppose it will create a branch in the relating Jenkins and we can poke it there?


mdelapenya:
I like the idea, although I would add logic to control this to accept only the ingest manager test suite, as other teams could not want to receive green build in the channel

PR #235, opened 2020-08-25
chore: use released versions

mdelapenya:
## What does this PR do?
It replaced the Docker images for the stack: from the observability-ci, to the official Docker registry.

## Why is it important?
Let's use released images to avoid sync issues

## Checklist

## Author's Checklist

- [x] Checked that observability-ci images have same SHA as the official repositories

### DIFF
--- a/cli/config/compose/profiles/ingest-manager/docker-compose.yml
+++ b/cli/config/compose/profiles/ingest-manager/docker-compose.yml
@@ -35,7 +35,7 @@ services:
     volumes:
       - ${kibanaConfigPath}:/usr/share/kibana/config/kibana.yml
   package-registry:
-    image: docker.elastic.co/package-registry/package-registry:41c150c8020efc53ab16e3bba774c62a419b51ea
+    image: docker.elastic.co/package-registry/distribution:staging
     healthcheck:
       test: ["CMD", "curl", "-f", "http://localhost:8080"]
       retries: 300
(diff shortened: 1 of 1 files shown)


cachedout (approved):

cli/config/compose/profiles/ingest-manager/docker-compose.yml
```diff
@@ -16,7 +16,7 @@ services:
       - xpack.security.authc.api_key.enabled=true
       - ELASTIC_USERNAME=elastic
       - ELASTIC_PASSWORD=changeme
-    image: "docker.elastic.co/observability-ci/elasticsearch:${stackVersion:-7.9.0}"
+    image: "docker.elastic.co/elasticsearch/elasticsearch:${stackVersion:-7.9.0}"
```
EricDavisX, line 19: @mdelapenya do you think we can make this dynamic to 'latest avialable of the 7.9.X line' as opposed to hard-coded to the 7.9.0 Beta build?  It is rather implied that its pulling .X latest else we can create a new branch '7.9.0' in e2e-testing ?

mdelapenya: If we have a "moving target" for that version, I'd hardcoded here. I chose 7.9.0 or the released one. We can always override for local or CI specific runs. If we find ourselves replacing it many times, it's the time to update the default value.

PR #236, opened 2020-08-25
fix: use systemctl to stop the agent process, otherwise pkill

mdelapenya:
## What does this PR do?
It uses systemctl to stop the elastic-agent service

Thanks to @michalpristas, we also noticed that the order of the execution matters. In the test framework we did (in this particular order):

1. a centos container is spun up
2. we copied the RPM installer and install it
3. we execute systemctl enable
4. we execute systemctl start
5. we create a fleet token
6. we enrol the agent with that token
7. we check the online status <- here it fails

We moved 4. after 6., so that the service is actually started AFTER the enrollment process. Can we check with @dedemorton if this is documented?

## Why is it important?
Let's stop services using the process manager, and fix a possible issue with how the service must be installed/configured.

## Checklist

- Relates https://github.com/elastic/beats/issues/20759

### DIFF
--- a/e2e/_suites/ingest-manager/fleet.go
+++ b/e2e/_suites/ingest-manager/fleet.go
@@ -92,6 +92,11 @@ func (fts *FleetTestSuite) anAgentIsDeployedToFleet(image string) error {
 		return err
 	}
 
+	err = systemctlRun(profile, image, image, "start")
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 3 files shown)


cachedout (approved):

mdelapenya:
Checked that the official docs explain the steps in that particular order: https://www.elastic.co/guide/en/ingest-management/current/run-elastic-agent.html

PR #244, opened 2020-08-26
[7.9.x] notify green builds for ingest-manager

mdelapenya:
Backports the following commits to 7.9.x:
 - chore: use debug in CI by default (6341a9a7)
 - chore: support notifying on green builds using a param (dac43608)
 - chore: notify green builds for ingest management scheduled builds (e9bbb2c3)
 - chore: optimise removing services (#241)


mdelapenya:
@cachedout, reminder for us: we should create (and refresh) the index for 7.9.0

PR #242, opened 2020-08-26
feat: support for RedHat Ubi 7

mdelapenya:
## What does this PR do?
It adds support for RedHat UBI, but to achieve it the developer (or the CI) has to login in RedHat registry, so docker pull works.

We simply added a new installer, reusing the Centos one, and added a new example in the scenario ooutlline for the restart agent.

## Why is it important?
It adds coverage in another platform.

## Checklist

- [x] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have run the Unit tests for the CLI, and they are passing locally
- [x] I have run the End-2-End tests for the suite I'm working on, and they are passing locally
- [ ] I have noticed new Go dependencies (run `make notice` in the proper directory)

## Author's Checklist

- [x] Created personal account in RedHat Developers to gain access to the registry, checking that the image is pulled

## Related issues

- Closes #239

## Follow-ups
@elastic/observablt-robots we need to enable login access to the RedHat registry for the CI: 
```shell
docker login registry.redhat.io -u CI_USER
```

We could add a cache for the installers, as with this PR we are downloading RPM artifact twice.

We could also run all Fleet scenarios using Outline, so all of them are run in the different supported platforms

Issue #252, opened 2020-08-28
[Ingest Manager] Add support for environment tests for Agent / Endpoint support matrix certification


EricDavisX:
This is in support of https://github.com/elastic/integrations-dev/issues/204
 and its original issue: https://github.com/elastic/integrations-dev/issues/22
 a new test addition is also the Red Hat UBI image: https://github.com/elastic/e2e-testing/issues/239

we can add these individually if needed, but I would prefer and will attempt to design some system that removes as much maintenance needs as possible.

Procuring all of the Docker images for the different environments may be a separate part of the chore.

Thoughts on the feature file + implementation code:

- At a minimum we could start a new feature file that has all of the environmental support.  If needed, it may be easier to start a new folder for the feature file (if it makes it easier to 'skip' them when not desired, like during every pr or every night).

- In the new file, we can link to the main Ingest feature file section that has Agent and Endpoint deploy (which is nearly merged!) for good reference.

- We could abstract the Go implementation side code into just a few feature file steps, with a parameterization akin to:
 Given an Agent is deployed to "<environment>" and Endpoint "is / isnt" included in Agent policy
 Then the "Elastic Agent" process is "running" on the host
 Then the "Metricbeat" process is "running" on the host
 Then the "Filebeat" process is "running" on the host
 And the "Elastic Endpoint" process is "running" on the host
 And Agent is unenrolled from Fleet
 And the "Elastic Agent" process is "running" on the host
 And the "Metricbeat" process is "stopped" on the host
 And the "Filebeat" process is "stopped" on the host
 And the "Elastic Endpoint" process is "stopped" on the host

We could go a little deeper with validating the data shows up, but that may need to be worked on at the feature side first.  Starting with this would be helpful.

EricDavisX:
the above PR has an example of how we could add in more support for different environment / OSes to run a regimen of tests against them.

EricDavisX:
In another ticket the open question as to how much of retesting do we need to do with various variables in the system came up.  Like for methods of install 1) enroll vs install, 2)  container usage vs standard host vm, 3) deb vs tar package, amd vs arm, and combinations of architecture and packaging and (thankfully we don't have to worry about x86 anymore as support was discontinued before we released).  

[...]

PR #269, opened 2020-09-04
feat: support for UBI images

mdelapenya:
## What does this PR do?
This PR adds official UBI images to the existing stand-alone tests, as they are basically the same: a docker image with the agent already bundled in it.

The change is very minimal: convert the existing Cucumber scenarios into scenario outlines with Examples, where we will use as examples the default stand-alone agent and the UBI image.

## Why is it important?
It adds regression testing coverage for the recently created UBI images.

## Checklist

```shell
$ OP_LOG_LEVEL=DEBUG DEVELOPER_MODE=true godog -t "stand_alone_mode"
```

At this moment, there are 2 failures, caused by an error with the index used to search for documents.

## Related issues

- Relates https://github.com/elastic/beats/issues/20992
- Relates #242


e2e/_suites/ingest-manager/features/stand_alone_agent.feature
```diff
@@ -5,18 +5,30 @@ Feature: Stand-alone Agent
...
 @start-agent
-Scenario: Starting the agent starts backend processes
-  When a stand-alone agent is deployed
+Scenario Outline: Starting the <image> agent starts backend processes
+  When a "<image>" stand-alone agent is deployed
   Then the "filebeat" process is in the "started" state on the host
     And the "metricbeat" process is in the "started" state on the host
+Examples:
+| image   |
+| default |
```
mdelapenya, line 14: We use "default", although the code will only "massage" the service name when `image == "ubi"`

cli/config/compose/services/elastic-agent-ubi/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+version: '2.3'
+services:
+  elastic-agent-ubi:
+    image: docker.elastic.co/beats/elastic-agent-ubi8:${elasticAgentTag:-8.0.0-SNAPSHOT}
```
jsoriano, line 4: Could we reuse the same scenario as for "elastic-agent"? Maybe somehow parameterizing the suffix here.
```suggestion
    image: docker.elastic.co/beats/elastic-agent${elasticAgentDockerImageSuffix}:${elasticAgentTag:-8.0.0-SNAPSHOT}
```

e2e/_suites/ingest-manager/features/stand_alone_agent.feature
```diff
@@ -5,18 +5,30 @@ Feature: Stand-alone Agent
   with them
 
 @start-agent
-Scenario: Starting the agent starts backend processes
-  When a stand-alone agent is deployed
+Scenario Outline: Starting the <image> agent starts backend processes
+  When a "<image>" stand-alone agent is deployed
```
jsoriano, line 9: I find confusing the use of "stand-alone" here. Does it refer to deploying the agent using a docker image? If that is the case, could we specifically say that?
This way it would be more clear that this feature tests the agent deployed with a docker image, while other features test the agent installed using some package.

```suggestion
Scenario Outline: Starting the <image> docker image of the agent starts backend processes
  When an agent is deployed using the "<image>" docker image
```

Then we could rename the feature to `docker_agent.feature` or something like this. If we don't want to couple the feature to specific technology we can use "container" instead of "docker".

jsoriano (approved):
Thanks for covering ubi images!

Changes LGTM, I have added some suggestions, but they can be seen as food for thought, or possible follow ups.

Also, more food for thought :slightly_smiling_face: It would be nice if docker images (and tar.gzs, zips...) could be seen as different installation methods, and we could do the same tests on them, so we can have scenarios like the ones we have in the `fleet_mode` feature to cover all supported platforms.

Imagine long-term having something like this:
```
Scenario Outline: Starting the agent installed in <os> as <package> starts backend processes
  When an agent installed in <os> as <package> is deployed to Fleet
  Then the "filebeat" process is in the "started" state on the host
   And the "metricbeat" process is in the "started" state on the host
Examples:
| os     | package |
| centos | rpm |
| centos | targz |
| centos | docker |
| centos | docker-ubi |
| debian | deb |
| debian | targz |
| debian | docker |
| debian | docker-ubi |
| windows | zip |
| windows | msi |
| windows | docker |
| mac | targz |
```

e2e/elasticsearch.go
```diff
@@ -316,6 +316,15 @@ func WaitForNumberOfHits(indexName string, query map[string]interface{}, desired
...
 		if err != nil {
+			log.WithFields(log.Fields{
+				"desiredHits": desiredHits,
+				"elapsedTime": exp.GetElapsedTime(),
+				"error":       err,
+				"index":       indexName,
+				"retry":       retryCount,
+			}).Warn("There was an error executing the query")
+
+			retryCount++
```
cachedout, line 327: Why does the retry counter get incremented here? It doesn't look like it will have an effect, since we just return below it.

kuisathaverat (approved):

e2e/_suites/ingest-manager/features/stand_alone_agent.feature
mdelapenya: @EricDavisX wdyt?

e2e/elasticsearch.go
mdelapenya: The `retryCount` variable is defined outside the retry function. This function, which returns an error, will be automatically retried by the backoff library when an error occurs, so that the `retryCount` variable is incremented with each failure.

cli/config/compose/services/elastic-agent-ubi/docker-compose.yml
mdelapenya: Yes, we can do this. I like it!

mdelapenya:
> Thanks for covering ubi images!
> 
> Changes LGTM, I have added some suggestions, but they can be seen as food for thought, or possible follow ups.
> 
> Also, more food for thought 🙂 It would be nice if docker images (and tar.gzs, zips...) could be seen as different installation methods, and we could do the same tests on them, so we can have scenarios like the ones we have in the `fleet_mode` feature to cover all supported platforms.
> 
> Imagine long-term having something like this:
> 
> ```
> Scenario Outline: Starting the agent installed in <os> as <package> starts backend processes
>   When an agent installed in <os> as <package> is deployed to Fleet
>   Then the "filebeat" process is in the "started" state on the host
>    And the "metricbeat" process is in the "started" state on the host
> Examples:
> | os     | package |
> | centos | rpm |
> | centos | targz |
> | centos | docker |
> | centos | docker-ubi |
> | debian | deb |
> | debian | targz |
> | debian | docker |
> | debian | docker-ubi |
> | windows | zip |
> | windows | msi |
> | windows | docker |
> | mac | targz |
> ```

I see this as a new feature file: `installers.feature`. Wdyt?

jsoriano:
> I see this as a new feature file: `installers.feature`. Wdyt?

Yeah, it can be a feature with a minimal feature with a reduced number of scenarios :+1:

mdelapenya:
What I'm not 100% sure is about the "Docker" install, as we run everything in docker

e2e/_suites/ingest-manager/features/stand_alone_agent.feature
EricDavisX: I appreciate and understand the idea.  To me, using Docker is just a means to test the 'Stand-alone' Agent mode which is a feature of Elastic Agent, as opposed to 'Docker' or 'Container' which is just how we've implemented the test.  I think I'd like to keep it as is, if it were up to me - it keeps the test scenarios feature oriented instead of technology aligned.

Further, it kinda highlights that we could implement other support for 'stand-alone' mode in that we could start with a blank Centos (or Debian or Red Hat ubi) and deploy the Agent as part of the test.  Doing that would require picking apart the current example / default config of the agent as it is packaged and replacing parts of it and so is a bit more work than what we have already done here which meets the needs of a basic test (and I admit we kind of stumbled into this usage on accident, using Docker that is, but it fits fine).  

Finally, the pure 'stand-alone' mode is not a focus now, so I am content to leave it as is and not spend time on reimplementing or changing test support - the real goal is to move faster to build in more OS support (in any fashion) and more features etc.  So, we're doing great!

EricDavisX (approved):
lgtm - thanks

e2e/_suites/ingest-manager/features/stand_alone_agent.feature
jsoriano: Ok, I get it, so it is just a feature that happens to be tested with the docker image, but it could be tested with other deployment approaches.

EricDavisX:
> What I'm not 100% sure is about the "Docker" install, as we run everything in docker, at least until we get to a clear separation of concerns for stand-alone (docker-based) versus fleet-mode (using installers)

I started a ticket to track the e2e-testing work prior - https://github.com/elastic/e2e-testing/issues/252
 - I had included Agent 'unenroll' and the 'Endpoint' integration as part of the test need.  I love the idea of a new file.

There are multiple parts to getting the coverage we want... running this outside of Docker is the big step, and it was first discussed in integrations-dev/ issue 32 and is now logged separately as integrations-dev/ issue 275

e2e/_suites/ingest-manager/stand-alone.go
```diff
@@ -131,6 +147,39 @@ func (sats *StandAloneTestSuite) getContainerLogs() error {
...
 
+// installTestTools we need the container name because we use the Docker Client instead of Docker Compose
+// we are going to install those tools we use in the test framework for checking
+// and verifications
+func (sats *StandAloneTestSuite) installTestTools(containerName string) error {
+	if sats.Image != "ubi8" {
+		return nil
+	}
+
+	cmd := []string{"microdnf", "install", "procps-ng"}
```
mdelapenya, line 158: @jsoriano we had to do this to be able to check for process existence in the test container.

I think it won't cause any overhead

mdelapenya:
The failed tests in https://beats-ci.elastic.co/blue/organizations/jenkins/e2e-tests%2Fe2e-testing-mbp/detail/PR-269/6/tests/ are considered flaky:

- no data streams
- metricbeat process not starting

The new tests for the UBI8 image passed (although with certain flakiness)

e2e/_suites/ingest-manager/stand-alone.go
jsoriano: Yeah, I don't think it will cause any issue. How are the checks we do? Have we alternatively considered reading the procfs directly? (something like `readlink /proc/*/exe | grep /metricbeat` might work).

mdelapenya: we are running `pgrep -n -l $PROCCESS`, with PROCESS=[elastic-agent | filebeat | metricbeat]

mdelapenya: The thing is, we are using the docker client to send the command, so any pipe `|` could be interpolated in the host instead of in the container

jsoriano: Oh ok, I see. Well, the "grep" could be done in the client side, as we check now with `strings.Contains` after `pgrep`.

mdelapenya: I'm gonna give it a try, although not sure about the glob expression with the star. Will post results, thanks!

mdelapenya: Running that command from inside the container works:

```shell
docker exec -ti ingest-manager_elastic-agent_1 bash
[root@d571995e1f80 elastic-agent]# readlink /proc/*/exe
/usr/share/elastic-agent/data/elastic-agent-b595bf/elastic-agent
/usr/share/elastic-agent/data/elastic-agent-b595bf/install/metricbeat-8.0.0-SNAPSHOT-linux-x86_64/metricbeat
/usr/share/elastic-agent/data/elastic-agent-b595bf/install/filebeat-8.0.0-SNAPSHOT-linux-x86_64/filebeat
/usr/share/elastic-agent/data/elastic-agent-b595bf/install/metricbeat-8.0.0-SNAPSHOT-linux-x86_64/metricbeat
/usr/bin/bash
/usr/bin/readlink
/usr/bin/readlink
```

But running it from outside, does not:
```shell
$ docker exec ingest-manager_elastic-agent_1 readlink /proc/*/exe
zsh: no matches found: /proc/*/exe
$ docker exec ingest-manager_elastic-agent_1 readlink "/proc/*/exe"
$ docker exec  ingest-manager_elastic-agent_1 "readlink /proc/*/exe"
OCI runtime exec failed: exec failed: container_linux.go:349: starting container process caused "exec: \"readlink /proc/*/exe\": stat readlink /proc/*/exe: no such file or directory": unknown
```

jsoriano: Interesting, it seems that `zsh` is trying to resolve the wildcard in the host. This should work no matter the shell in the host:
```
$ docker exec ingest-manager_elastic-agent_1 sh -c 'readlink /proc/*/exe'
```
Or directly:
```
$ docker exec ingest-manager_elastic-agent_1 sh -c 'grep /metricbeat /proc/*/exe'
```
And check exit status.

jsoriano: But don't spend a lot of time with this, installing tools if needed looks good too :slightly_smiling_face:

mdelapenya:
If you all agree I'd like to merge this support, although we'll find the same test errors as described above

jsoriano (approved):

Issue #274, opened 2020-09-08
Create maintenance branches


mdelapenya:
I think some of the APIs have changed from 7.9 to 7.10 / 8.0... I know of one, but I'm not sure if we're done making changes to the API or the responses at this point in the cycle.  I'd like to back-port it to 7.9 of course, but I'm not sure its worth the effort.

I do want to backport / create a 7.10 branch of the tests, if we can tho.  @mdelapenya

_Originally posted by @EricDavisX in https://github.com/elastic/e2e-testing/pull/222#issuecomment-688955547_

mdelapenya:
Done

PR #280, opened 2020-09-10
chore: bump package registry to staging

### LINKED ISSUE #253: [Ingest Manager] Wrong package versions, need to update to using a :distribution image (tested w 7.10.0)
I ran the test recently with a 7.10 variable citation as below:

export DEVELOPER_MODE=true && ELASTIC_AGENT_VERSION=7.10.0-SNAPSHOT STACK_VERSION=7.10.0-SNAPSHOT godog -t @enroll

This pulled what seems to be the 7.8.x list of packages, just the original 9 of them that we had then.

I need help reviewing how the branches in e2e-testing are set up and what we might be able to do here... at a minimum, we could make sure the Registry image we use is always refreshed with latest Integration packages from the storage repo it points to (snapshot, staging, or production).

Its a bit of research to figure out, I don't know yet whats up.  Maybe Ruflin can help or SKH who has worked with getting the registry docker container in use in Kibana side.

screenshots:

Endpoint on at 0.8.0 which is old
Screen Shot 2020-08-28 at 10 26 45 AM

base 9 integrations shown:
Screen Shot 2020-08-28 at 10 26 29 AM

mdelapenya:
## What does this PR do?
This PR bumps the package registry to `staging`. After the change, we noticed the title for the "Elastic Endpoint" integration changed to "Elastic Endpoint Security", so we renamed the feature files and other usages in consequence.

## Why is it important?
The staging tag is updated more frequently and, for master branch, we want to use latest code.

About the usage of the title instead of the name, we are using the title because the feature files have to be super readable, and the title is more readable than the name, which is usually an identifier that could bring less value from business side.

## Checklist

To run the endpoint integration scenarios:

```shell
$ OP_LOG_LEVEL=TRACE DEVELOPER_MODE=true godog -t "agent_endpoint_integration"
```

## Related issues

- Closes #253

## Use cases

```gherkin
Given the master branch
When we run the tests
Then the current development version of the registry is used

Given a maintenance branch (7.x.x)
When we run the tests
Then the released version ffor that branch is used
```

### DIFF
--- a/cli/config/compose/profiles/ingest-manager/docker-compose.yml
+++ b/cli/config/compose/profiles/ingest-manager/docker-compose.yml
@@ -35,7 +35,7 @@ services:
     volumes:
       - ${kibanaConfigPath}:/usr/share/kibana/config/kibana.yml
   package-registry:
-    image: docker.elastic.co/package-registry/package-registry:41c150c8020efc53ab16e3bba774c62a419b51ea
+    image: docker.elastic.co/package-registry/distribution:staging
     healthcheck:
       test: ["CMD", "curl", "-f", "http://localhost:8080"]
       retries: 300
--- a/e2e/_suites/ingest-manager/features/agent_endpoint_integration.feature
+++ b/e2e/_suites/ingest-manager/features/agent_endpoint_integration.feature
@@ -6,8 +6,8 @@ Feature: Agent Endpoint Integration
 Scenario: Adding the Endpoint Integration to an Agent makes the host to show in Security App
   Given a "centos" agent is deployed to Fleet
     And the agent is listed in Fleet as "online"
-  When the "Elastic Endpoint" integration is "added" in the "default" policy
-  Then the "Elastic Endpoint" datasource is shown in the "default" policy as added
+  When the "Elastic Endpoint Security" integration is "added" in the "default" policy
+  Then the "Elastic Endpoint Security" datasource is shown in the "default" policy as added
     And the host name is shown in the Administration view in the Security App as "online"
     
 @endpoint-policy-check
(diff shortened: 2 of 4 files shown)


cachedout (approved):

cli/config/compose/profiles/ingest-manager/docker-compose.yml
```diff
@@ -35,7 +35,7 @@ services:
     volumes:
       - ${kibanaConfigPath}:/usr/share/kibana/config/kibana.yml
   package-registry:
-    image: docker.elastic.co/package-registry/package-registry:41c150c8020efc53ab16e3bba774c62a419b51ea
```
ruflin, line 38: I'm tempted to say we should use here `distribution:snapshot` and use `distribution:staging` for the 7.x e2e tests.

mdelapenya: I we have this tag, I'm OK with it.

For the 7.x branch, we do not have that branch, only 7.9.x and eventually 7.10.x. Should we have a 7.x branch?

ruflin: As we have 7.* and 8.* branches I assume we are aligned with the release branches. So I think yes. Then it becomes self explaining on what monitors which part.

mdelapenya: Good. We'd need then to define which versions for the stack are needed, as there is no https://container-library.elastic.co/r/elasticsearch/elasticsearch:7.x image

mdelapenya: We could use this issue for reference: https://github.com/elastic/e2e-testing/issues/274

ruflin: There actually is one: `docker.elastic.co/elasticsearch/elasticsearch:7.x-SNAPSHOT`

PR #281, opened 2020-09-10
fix: remove references to the default policy

mdelapenya:
## What does this PR do?
This PR updates the specs/scenarios removing any reference to the default policy, because since #279 we are creating a policy for each scenario.

This change does not affect the tests, as we are internally using the policyID, so it should work as expected.

## Checklist

Run:
```shell
$ OP_LOG_LEVEL=TRACE DEVELOPER_MODE=true godog -t "agent_endpoint_integration"
```

## Related issues

- Relates #279

### DIFF
--- a/e2e/_suites/ingest-manager/features/agent_endpoint_integration.feature
+++ b/e2e/_suites/ingest-manager/features/agent_endpoint_integration.feature
@@ -6,8 +6,8 @@ Feature: Agent Endpoint Integration
 Scenario: Adding the Endpoint Integration to an Agent makes the host to show in Security App
   Given a "centos" agent is deployed to Fleet
     And the agent is listed in Fleet as "online"
-  When the "Elastic Endpoint" integration is "added" in the "default" policy
-  Then the "Elastic Endpoint" datasource is shown in the "default" policy as added
+  When the "Elastic Endpoint" integration is "added" in the policy
+  Then the "Elastic Endpoint" datasource is shown in the policy as added
     And the host name is shown in the Administration view in the Security App as "online"
     
 @endpoint-policy-check
@@ -32,7 +32,7 @@ Scenario: Un-enrolling Elastic Agent stops Elastic Endpoint
 @deploy-endpoint-then-remove-it-from-policy
 Scenario: Removing Endpoint from Agent policy stops the connected Endpoint
   Given an Endpoint is successfully deployed with a "centos" Agent
-  When the "Elastic Endpoint" integration is "removed" in the "default" policy
+  When the "Elastic Endpoint" integration is "removed" in the policy
   Then the agent is listed in Fleet as "online"
     But the host name is not shown in the Administration view in the Security App
     And the "elastic-endpoint" process is in the "stopped" state on the host
--- a/e2e/_suites/ingest-manager/fleet.go
+++ b/e2e/_suites/ingest-manager/fleet.go
@@ -124,8 +124,8 @@ func (fts *FleetTestSuite) contributeSteps(s *godog.Suite) {
 	s.Step(`^the "([^"]*)" process is "([^"]*)" on the host$`, fts.processStateChangedOnTheHost)
 
 	// endpoint steps
-	s.Step(`^the "([^"]*)" integration is "([^"]*)" in the "([^"]*)" policy$`, fts.theIntegrationIsOperatedInThePolicy)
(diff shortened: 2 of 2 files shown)


cachedout (approved):

EricDavisX:
Hi, thanks Manu - so the implication here is that each scenario runs in a separate, newly created (for that test) policy?  That is certain to avoid failures / clean-up related problems we may have had, its a good addition.

I see 2 areas to discuss / confirm:
- this will require us to implement a new step syntax and supporting implementation to allow testing the change of Agent from one policy to another, which is a core test the Endpoint team (or myself) should intend to write soon.  If we could keep the parameterized "policy name" usage that might be wise and save ourselves more work later (if it is already implemented, it may not be.  Maybe its not hard to fix / change later... this isn't a must-have, just wanted us to think.

- this also precludes us from testing the default policy itself - which is a unique case.  I actually don't know if its created programmatically at Fleet setup time or if its hard-coded in any way.  If its at all hard-coded we'd want to implement some new specific test that uses it, but with the assurance that no test scenario or code changes it in any way (which should be easy enough).  @ph can you or someone help confirm that (about the default policy creation at Fleet setup time) ?

mdelapenya:
> Hi, thanks Manu - so the implication here is that each scenario runs in a separate, newly created (for that test) policy? That is certain to avoid failures / clean-up related problems we may have had, its a good addition.

Yes, the policy (master)/configuration (7.9.x) will be created before each scenario, so that each agent will be enrolled using this new policy.

> 
> I see 2 areas to discuss / confirm:
> 
> * this will require us to implement a new step syntax and supporting implementation to allow testing the change of Agent from one policy to another, which is a core test the Endpoint team (or myself) should intend to write soon.  If we could keep the parameterized "policy name" usage that might be wise and save ourselves more work later (if it is already implemented, it may not be.  Maybe its not hard to fix / change later... this isn't a must-have, just wanted us to think.

I'd suggest working on this at the moment we need it, not now.

> * this also precludes us from testing the default policy itself - which is a unique case.  I actually don't know if its created programmatically at Fleet setup time or if its hard-coded in any way.  If its at all hard-coded we'd want to implement some new specific test that uses it, but with the assurance that no test scenario or code changes it in any way (which should be easy enough).  @ph can you or someone help confirm that (about the default policy creation at Fleet setup time) ?

It's created at the beginning of the test execution, in the beforeSuite (before any scenario) hook. We have a Fleet setup, which calls a kibana URL for that. I guess it creates the default policy. I'd ask about what the implications of using the default policy are, to understand if we need specific scenarios for it. As of today, where a policy seems to be exactly the same as the default one (or at least I cannot distinguish any difference), I'd treat default as any other one, so I do not a reason to not use the created ones... Unless we find a specific use case for the default policy

EricDavisX:
> > * this also precludes us from testing the default policy itself - which is a unique case.  I actually don't know if its created programmatically at Fleet setup time or if its hard-coded in any way.  If its at all hard-coded we'd want to implement some new specific test that uses it, but with the assurance that no test scenario or code changes it in any way (which should be easy enough).  @ph can you or someone help confirm that (about the default policy creation at Fleet setup time) ?
> 
> [The policy the test creates is] created at the beginning of the test execution, in the beforeSuite (before any scenario) hook. We have a Fleet setup, which calls a kibana URL for that. I guess it creates the default policy. I'd ask about what the implications of using the default policy are...

@ph or @jen-huang maybe - Manu's question is exactly what I would like to confirm.  The policies are supposed to be the same, and are currently - if there is any need to test the 'default' policy especially we need to know it now so we can create a new test case for it.

ph:
@EricDavisX @mdelapenya What you are describing is exactly what is happening behind the scene when setup is run the default policy is created and necessary packages are installed for the system integration. Testing the default policy would be testing logs and metrics are collected on the machine and I think we should have a test for that, just for sanity.

Note, I would keep the "assertion" as minimal as possible, because the system integration is in flux at the moment and we are iterating to reduce the scope of that integration. Reduce the metrics and the logs we are collecting and dealing with the difference between the operating system.

mdelapenya:
I think we can merge this, and move the discussion about default policies to another issue, as we already merged the usage of a new policy for each scenario in #279

EricDavisX:
agree - thanks.  and it seems (still confirming) that we need not re-build any specific test for the Ingest Manager provided 'pre-built' policy, I believe it is programmatically created and is like any other policy so this change was quite harmless to test coverage (just needed to be confirmed).  Also, indeed, we can continue tests of the System Integration at all speed.  I found we didn't have a ticket for that, so I added one: https://github.com/elastic/e2e-testing/issues/284

PR #289, opened 2020-09-14
fix: use not hidden index name

mdelapenya:
## What does this PR do?
It renames the index used for storing the logs sent by the agent, used in the stand-alone mode test suite.

## Why is it important?
Stabilise the test suite, which is failing in master and 7.9.x (we have to backport this)

## Checklist

- [x] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have run the Unit tests for the CLI, and they are passing locally
- [x] I have run the End-2-End tests for the suite I'm working on, and they are passing locally
- [ ] I have noticed new Go dependencies (run `make notice` in the proper directory)

## How to test this PR locally

```shell
cd e2e/_suites/ingest-manager
OP_LOG_LEVEL=TRACE DEVELOPER_MODE=true godog -t "stand_alone_mode"
```

## Related issues

- Closes #288

PR #285, opened 2020-09-14
[7.9.x] chore: bump 7.9.0 to 7.9.1

mdelapenya:
## What does this PR do?
It bumps the versions for the stack and the agent, to use the latest 7.9.1 version

## Why is it important?
The 7.9.0 binaries for the elastic-agent dissapeared (https://artifacts-api.elastic.co/v1/search/7.9.0/elastic-agent returns a 404 not-found error), causing all tests to fail.

```log
[2020-09-14T07:49:39.868Z] time="2020-09-14T07:49:39Z" level=warning msg="The Elastic artifacts API is not available yet" arch=x86_64 artifact=elastic-agent elapsedTime=27.858337442s error="GET request failed with 404" extension=rpm os=linux retry=10 statusEndpoint="https://artifacts-api.elastic.co/v1/search/7.9.0/elastic-agent" version=7.9.0
```

Bumping the stack and the agent to 7.9.1 brings back the tests again.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have run the Unit tests for the CLI, and they are passing locally
- [x] I have run the End-2-End tests for the suite I'm working on, and they are passing locally
- [ ] I have noticed new Go dependencies (run `make notice` in the proper directory)

## Author's Checklist

- [x] Run the CI with the 7.9.1 parameters

PR #291, opened 2020-09-14
fix: use not hidden index name (#289) backport for 7.9.x

mdelapenya:
Backports the following commits to 7.9.x:
 - fix: use not hidden index name (#289)

- Needs #285


mdelapenya:
Failing because the 7.9.x branch is using 7.9.0 as the version used by stack and agent. Waiting to merge #285 into 7.9.x to merge this

PR #299, opened 2020-09-16
feat: log elastic-agent logs before terminating its container

### LINKED ISSUE #298: [Elastic-Agent] get agent's logs before removing the underlying coontainer
As the native `docker logs CONTAINER` is not retrieving the agent logs, we have to inspect the internals of the agent too fetch the log.

mdelapenya:
## What does this PR do?
As the native `docker logs CONTAINER` is not retrieving the agent logs, we have to inspect the internals of the agent too fetch the log. For that reason this PR adds a function to retrieve the logs from the internals of the elastic-agent structure, and to achieve it we have to get the build hash of the agent, using a short version of that hash to calculate the path to the log file.

It will be the installer struct the responsible for doing these tasks, as it knows the internals of the agent.

The way we consume this logs is in the afterScenario test life cycle hook, right before removing the docker service representing the agent.

## Why is it important?
We want to troubleshoot problems on CI

## Checklist

```shell
$ OP_LOG_LEVEL=TRACE DEVELOPER_MODE=true godog -t "fleet_mode && start-agent"
```

## Related issues

- Closes #298

## Follow-ups
The backport to 7.9.x must be done in a manual mode, because agent's new structure (including build hash) is not backported to 7.9.x., so the log file can be retrieved directly from `/var/lib/elastic-agent/logs/elastic-agent-json.log`.

### DIFF
--- a/cli/docker/docker.go
+++ b/cli/docker/docker.go
@@ -102,6 +102,19 @@ func ExecCommandIntoContainer(ctx context.Context, containerName string, user st
 
 	output = strings.ReplaceAll(output, "\n", "")
 
+	patterns := []string{
+		"\x01\x00\x00\x00\x00\x00\x00\r",
+		"\x01\x00\x00\x00\x00\x00\x00)",
+	}
+	for _, pattern := range patterns {
+		if strings.HasPrefix(output, pattern) {
+			output = strings.ReplaceAll(output, pattern, "")
+			log.WithFields(log.Fields{
+				"output": output,
+			}).Trace("Output name has been sanitized")
+		}
+	}
+
 	return output, nil
 }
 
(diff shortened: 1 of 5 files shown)


cli/docker/docker.go
```diff
@@ -102,6 +102,19 @@ func ExecCommandIntoContainer(ctx context.Context, containerName string, user st
 
 	output = strings.ReplaceAll(output, "\n", "")
 
+	patterns := []string{
+		"\x01\x00\x00\x00\x00\x00\x00\r",
```
mdelapenya, line 106: The docker client is adding this garbage

cachedout (approved):

PR #310, opened 2020-09-16
[7.9.x] chore: log agent logs before terminating the container

mdelapenya:
## What does this PR do?
Backport of #299

## Checklist

## Related issues

- Backports #299  

## Follow-ups
When merging this PR, let's not squash the two commits, in the case they backport the new structure for the agent and we have to use the first commit, only

### DIFF
--- a/cli/docker/docker.go
+++ b/cli/docker/docker.go
@@ -102,6 +102,19 @@ func ExecCommandIntoContainer(ctx context.Context, containerName string, user st
 
 	output = strings.ReplaceAll(output, "\n", "")
 
+	patterns := []string{
+		"\x01\x00\x00\x00\x00\x00\x00\r",
+		"\x01\x00\x00\x00\x00\x00\x00)",
+	}
+	for _, pattern := range patterns {
+		if strings.HasPrefix(output, pattern) {
+			output = strings.ReplaceAll(output, pattern, "")
+			log.WithFields(log.Fields{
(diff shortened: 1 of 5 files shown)


cli/docker/docker.go
```diff
@@ -102,6 +102,19 @@ func ExecCommandIntoContainer(ctx context.Context, containerName string, user st
 
 	output = strings.ReplaceAll(output, "\n", "")
 
+	patterns := []string{
```
kuisathaverat, line 105: my binary is a little rust, What is it for?

kuisathaverat: ok, I saw it https://github.com/elastic/e2e-testing/pull/299/files#r489219873

kuisathaverat, line 105:
```suggestion
        // Docker command garbage cleaner :D 
	patterns := []string{
```

mdelapenya: I guess the docker client is returning something weird when converting the bytes. Maybe I'm doing something weird stuff with the buffer and the reader. I'd need to debug it.

mdelapenya: There is a trace log https://github.com/elastic/e2e-testing/pull/310/files#diff-0d15ecc74d96858feb62aab759545abeR114

Should we add the comment too?

kuisathaverat (approved):

PR #315, opened 2020-09-22
feat: support contributing a GH check to the upstream job that triggered the build

### LINKED ISSUE #314: [CI] SUpport contributing GH checks to the upstream PRs
As we are consuming this pipeline from Beats PRs, we want it to contribute a GH check to the Beats pipeline with the result of the e2e tests.

cc/ @EricDavisX @ph

mdelapenya:
## What does this PR do?
It adds three input parameters to the Jenkinsfile, empty by default, which will be populated by upstream jobs (more specifically PRs triggering this job).

It will post the check after the job is finished, using the upstream parameters to configure the check.

## Why is it important?
It will allow Beats PRs to be notified about the E2E tests, providing a fast feedback about the status of the original PR.

>We'd like to warn about the risks of merging the upstream PR without checking the results: we must wait for all checks to be green/passing.

## Checklist

## Author's Checklist

- [x] Used same approach than in the APM ITs

## Related issues

- Closes #314 
- Relates https://github.com/elastic/beats/pull/21100

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -9,6 +9,7 @@ pipeline {
     BASE_DIR = "src/github.com/elastic/${env.REPO}"
     NOTIFY_TO = credentials('notify-to')
     JOB_GCS_BUCKET = credentials('gcs-bucket')
+    JOB_GIT_CREDENTIALS = "f6c7695a-671e-4f4f-a331-acdce44ff9ba"
     DOCKER_ELASTIC_SECRET = 'secret/observability-team/ci/docker-registry/prod'
     DOCKER_REGISTRY = 'docker.elastic.co'
   }
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -43,6 +43,9 @@ pipeline {
     string(name: 'HELM_VERSION', defaultValue: '2.16.3', description: 'SemVer version of Helm to be used.')
     string(name: 'HELM_KIND_VERSION', defaultValue: '0.8.1', description: 'SemVer version of Kind to be used.')
     string(name: 'HELM_KUBERNETES_VERSION', defaultValue: '1.18.2', description: 'SemVer version of Kubernetes to be used.')
+    string(name: 'GITHUB_CHECK_NAME', defaultValue: '', description: 'Name of the GitHub check to be updated. Only if this build is triggered from another parent stream.')
+    string(name: 'GITHUB_CHECK_REPO', defaultValue: '', description: 'Name of the GitHub repo to be updated. Only if this build is triggered from another parent stream.')
+    string(name: 'GITHUB_CHECK_SHA1', defaultValue: '', description: 'Name of the GitHub repo to be updated. Only if this build is triggered from another parent stream.')
   }
   stages {
     stage('Initializing'){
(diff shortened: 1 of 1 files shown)


.ci/Jenkinsfile
```diff
@@ -227,6 +230,7 @@ pipeline {
   }
   post {
     cleanup {
+      githubCheckNotify(currentBuild.currentResult == 'SUCCESS' ? 'SUCCESS' : 'FAILURE')
```
v1v, line 233: I'd say to add the 'PENDING' check in [L77](https://github.com/elastic/e2e-testing/pull/315/files#diff-8db37e8fcea0f1a8f2f39667e94ebcc4L77)

.ci/Jenkinsfile
```diff
@@ -300,3 +304,16 @@ def generateFunctionalTestStep(Map params = [:]){
...
+/**
+ Notify the GitHub check of the parent stream
+**/
+def githubCheckNotify(String status) {
+  if (params.GITHUB_CHECK_NAME?.trim() && params.GITHUB_CHECK_REPO?.trim() && params.GITHUB_CHECK_SHA1?.trim()) {
+    githubNotify context: "${params.GITHUB_CHECK_NAME}",
+                 description: "${params.GITHUB_CHECK_NAME} ${status.toLowerCase()}",
+                 status: "${status}",
+                 targetUrl: "${env.RUN_DISPLAY_URL}",
+                 sha: params.GITHUB_CHECK_SHA1, account: 'elastic', repo: params.GITHUB_CHECK_REPO, credentialsId: env.JOB_GIT_CREDENTIALS
```
v1v, line 317: `JOB_GIT_CREDENTIALS` is not defined in the environment section [L7](https://github.com/elastic/e2e-testing/pull/315/files#diff-8db37e8fcea0f1a8f2f39667e94ebcc4R7)

.ci/Jenkinsfile
mdelapenya: It's set here in the upstream job: https://github.com/elastic/beats/pull/21100/files#diff-bdaa5a61de14efda6eb2bbb2a38fa9a8R269

Should you suggest adding it here and remove it from there? I like the idea of having the control in one single place

.ci/Jenkinsfile
mdelapenya: Added, good catch!

.ci/Jenkinsfile
```diff
@@ -9,6 +9,7 @@ pipeline {
     BASE_DIR = "src/github.com/elastic/${env.REPO}"
     NOTIFY_TO = credentials('notify-to')
     JOB_GCS_BUCKET = credentials('gcs-bucket')
+    JOB_GIT_CREDENTIALS = "f6c7695a-671e-4f4f-a331-acdce44ff9ba"
```
mdelapenya, line 12: Taken from Beats repo

.ci/Jenkinsfile
v1v: I'm trying to recall if there is any scenario that will be required, but nothing on the top off my head, let's then leave it as is

v1v (approved):

.ci/Jenkinsfile
mdelapenya: In the other hand, I think it's harmless to add it here instead of in the packaging job: for PRs in this repo, it will no trigger this check; for upstream PRs, it will trigger the check

mdelapenya: I've checked that the opbeans pipeline follows the same approach (https://github.com/elastic/apm-pipeline-library/blob/f0b506f783e0fe5b468d034d6459937fc981f806/vars/opbeansPipeline.groovy#L202-L208):
- trigger the downstream job
- start the pending check
which is the same as we do for this use case: the packaging job triggers this job and starts the pending check.

PR #333, opened 2020-09-29
chore: extract integrations' interactions with kibana to a dedicated client

mdelapenya:
## What does this PR do?
It extracts existing integrations HTTP requests to Kibana to a dedicated client, which will handle any HTTP request (GET, POST, PUT) to Kibana APIs.

It basically moves the code responsible of creating the HTTP request from the e2e test suite to the client.

## Why is it important?
This approach will allow to unit-test the HTTP requests, supporting detecting API changes the soonest.

## Checklist

```shell
$ OP_LOG_LEVEL=TRACE DEVELOPER_MODE=true TIMEOUT_FACTOR=3  godog -t "agent_endpoint_integration"
```

## Follow-ups
We are extracting only the inteegrations API calls, but we can move Fleet calls too.

### DIFF
--- a/cli/services/kibana.go
+++ b/cli/services/kibana.go
@@ -25,18 +25,19 @@ const ingestManagerIntegrationURL = "/api/ingest_manager/epm/packages/%s-%s"
 
 // KibanaClient manages calls to Kibana APIs
 type KibanaClient struct {
-	url string
+	baseURL string
+	url     string
 }
 
 // NewKibanaClient returns a kibana client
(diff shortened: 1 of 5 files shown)


cli/services/kibana.go
```diff
@@ -0,0 +1,239 @@
+package services
```
mdelapenya, line 1: This file is in the `cli` side so that the unit tests are run in CI with no change.

kuisathaverat (approved):

PR #339, opened 2020-10-01
[PACT] feat: add Contract tests for Kibana Fleet

mdelapenya:
## What does this PR do?
It adds a contract tests with Pact.io, allowing two things:
- From consumer perspective (the e2e tests), define the expected behaviours in the format of contracts
- From provider perspective (Kibana Fleet), verify that any change in the code is not breaking the contracts provided by the consumers

It simply adds 3 tests in the contract:
1. when we query Fleet APIs, then there are integrations, including the proper status code (200 OK)
1. when we look for an existing integration, it's returned including the proper status code (200 OK)
1. when we look for a non-existing integration, the right status code is returned (404 not found)

> With these tests we detected that in kibana-7.9.1 the error code was 500 server error

The generated contracts are stored locally for simplicity, instead of using a Pact Broker (see docs).

## Why is it important?
This is a PoC, but it will allow decoupling Kibana from its consumers, verifying they are not broken in a proactive manner, instead of in a reactive one (when their tests fail, if executed)

## Checklist

## Author's Checklist

- [x] Checked the life cicle of creating contracts and verifying them

From root directory:
```shell
$ make pact-consumer
$ make verify-provider
```

## Follow-ups

Move the verification process to Kibana (or the sotware identified as provider)


mdelapenya:
Pinging @MadameSheema and @jfsiii, as they were interested in contract testing

jfsiii:
@mdelapenya Thanks for the ping. I'm excited to watch this and learn more.

Is there any tooling to go from OpenAPI to the Pact JSON format? I don't know much about Pact and would like to avoid describing the HTTP two different ways in two different locations. i.e. OpenAPI in the plugin and Pact in this repo

mdelapenya:
@jfsiii https://docs.pact.io/faq/convinceme/#but-i-use-swaggeropenapi

BTW, did not mention this pr is the result of an space time project, probably not going into the project...

CONTRACT_TESTING.md
```diff
@@ -0,0 +1,65 @@
...
+- **meta**: Pact specification and version (2.0.0)
+
+
+#### Verify the contracts from provider side
+We want to verify that the provider satisfies the contracts. For that reason, we need the provider to be started first, and to achieve it, we are providing the following Make commands:
+
+```shell
+$ make prepare-pact-provider-deps # will run an Elasticsearch, Kibana andd Package Registry
+$ make pact-provider              # will run the verification of the contracts by itself
+$ make verify-provider            # will tear down the runtime dependencies
```
v1v, line 62: `verify-provider` seems to call some other make goals -> https://github.com/elastic/e2e-testing/pull/339/files#diff-b67911656ef5d18c4ae36cb6741b7965R47

Is the comment correct?

mdelapenya: Yes, it's on purpose: verify is the last piece in the chain, defining a well-known life cycle: prepare > verify > destroy

I did this that way to support decoupling the destroy goal in the post stage of the CI pipeline (or to be run manually)

cachedout (approved):
💯

PR #340, opened 2020-10-01
[PACT] [7.9.x] feat: add Contract tests for Kibana Fleet

mdelapenya:
## What does this PR do?
It adds a contract tests with Pact.io, allowing two things:
- From consumer perspective (the e2e tests), define the expected behaviours in the format of contracts
- From provider perspective (Kibana Fleet), verify that any change in the code is not breaking the contracts provided by the consumers

It simply adds 3 tests in the contract:
1. when we query Fleet APIs, then there are integrations, including the proper status code (200 OK)
1. when we look for an existing integration, it's returned including the proper status code (200 OK)
1. when we look for a non-existing integration, the right status code is returned (500 server error)

> With these tests we detected that in kibana-8.0.0-SNAPSHOT the error code was 404 not found

The generated contracts are stored locally for simplicity, instead of using a Pact Broker (see docs).

## Why is it important?
This is a PoC, but it will allow decoupling Kibana from its consumers, verifying they are not broken in a proactive manner, instead of in a reactive one (when their tests fail, if executed)

## Checklist

## Author's Checklist

- [x] Checked the life cicle of creating contracts and verifying them

From root directory:
```shell
$ make pact-consumer
$ make verify-provider
```

## Related issues

- Backport of #339 

## Follow-ups

Move the verification process to Kibana (or the sotware identified as provider)


CONTRACT_TESTING.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Fleet contracts
+We are already running E2E tests for Fleet, defining the scenarios and behaviours using Gherkin (see [here](./e2e/_suites/ingest-manager/features)). As explained above, we now want to verify that the contracts between Kibana Fleet and this project is not broken because of changes in the APIs the tests consume.
+
+For that reason, we are going to use Pact to get notified whenever the contracts are not satisfied. The steps that we are folowing are:
+
+1. Create unit tests mocking the HTTP requests/responses for Fleet's APIs. As we all know, this approach will always work, as we predefine the mocks, but on the other hand we won't get notified if/when the provider changes, so we must be reactively checking for updates.
```
cachedout, line 33:
```suggestion
1. Create unit tests mocking the HTTP requests/responses for Fleet APIs. As we all know, this approach will always work, as we predefine the mocks, but on the other hand we won't get notified if/when the provider changes, so we must be reactively checking for updates.
```

CONTRACT_TESTING.md
```diff
@@ -0,0 +1,65 @@
...
+1. Create a contract/pact for the consumer (the e2e tests), with the behaviours we expect from Fleet. The result will be a JSON file with the specs.
+1. Store the contract in some place that is accesible by the provider. In this case, we are using the file system to store them, but Pact recommends using a [Broker](https://github.com/pact-foundation/pact_broker).
+1. Verify the contract/pact from the provider side. The provider is able to check if a change breaks all its consumers by itself, as it has access to the contracts they provided (via file system or the afore mentioned Broker).
```
cachedout, line 36:
```suggestion
1. Verify the contract/pact from the provider side. The provider is able to check if a change breaks all its consumers by itself, as it has access to the contracts they provided (via file system or the aforementioned Broker).
```

cachedout (approved):
This is really cool. 💯

PR #378, opened 2020-10-21
feat: support installing the agent with TAR installer and the install subcommand

### LINKED ISSUE #326: [Fleet / Agent ] support install / uninstall command from Agent
I'm going to put some thoughts down for this.

From this issue: https://github.com/elastic/beats/issues/21019
 - its merging in the next day or two so would be good to prep some feature file work and implementation as needed.

briefly:
the feature adds in new support, and leaves the old in place.  
So we'll still want to test the enroll and unenroll as we are.  but we can add in new feature tests for 'install' and 'uninstall' from the Agent side.

mdelapenya:
## What does this PR do?
This PR adds the the TAR installer so that it's possible to install the agent from a TAR file. For that, we had to define a proper life cycle for agent commands (install, enroll), as depending on the installer type (RPM, DEB, TAR), the life cycle (preInstall > install > enroll > postInstall) depends on the installer:

**preInstall:**
- `TAR` leverages preInstall phase to extract the TAR file and create the proper layout

**install:**
- `RPM` and `DEB` leverage install phase to use package managers to install the artifact
- `TAR` leverages install phase to actually install and enroll the agent

**enroll:**
- `RPM` and `DEP` needs an extra enrollment call
- `TAR` defines an enroll method to be used outside the life

**postInstall:**
- `RPM` and `DEB` leverages postInstall phase to enable and run the service using systemd

This PR also refactors the scenarios (feature files) merging the subcommand feature file with Fleet one. In these files, and for the sake of a green CI build, we are removing any interaction with the RPM and DEB installers, because we prefer the TAR install process. We could at any time turn them back as Outline examples in Gherkin:

```gherkin
Examples:
| os     | installer |
| centos | tar       |
| centos | systemd   |
| debian | tar       |
| debian | systemd   |
```
>We know that systemd is not an installer by itself, but it installs the elastic-agent service in Centos/Debian.

## Why is it important?
The first main thing is that the PR matches the e2e tests code with the Elastic Agent way of installing the agent, which is critical. The second thing, the tests pass locally and we expect to fix the tests on CI too.

## Checklist

```shell
SUITE=fleet TAGS="fleet_mode" DEVELOPER_MODE=false TIMEOUT_FACTOR=3 LOG_LEVEL=TRACE make -C e2e functional-test
```

## Related issues

- Closes #326

## Follow-ups
The test for uninstalling the agent is failing. Need to check. I pushed the PR to receive feedback from the team.

We could refactor the download of the agents and add it when installing the agent instead of at the beginning of the test suite. Why? Because we download them at that time to create a volume and mount it into the container, at startup. This is causing that, when running the tests in DEV mode (DEVELOPER_MODE=true) locally, where the services are not destroyed, docker-compose tries to reuse the service without recreating it (we are not using the `--force-recreate` option). Therefore, that installer is already there and could lead to unexpected errors because of garbage in the target container for the agent.


e2e/_suites/fleet/services.go
```diff
@@ -270,32 +385,196 @@ func newDebianInstaller() (ElasticAgentInstaller, error) {
...
+			log.WithFields(log.Fields{
+				"containerName": containerName,
+				"error":         err,
+			}).Error("Could not get agent hash in the container")
+
+			return err
+		}
+
+		// install the elastic-agent to /usr/bin/elastic-agent using command
+		binary := fmt.Sprintf("/elastic-agent/data/elastic-agent-%s/", hash) + artifact
```
michalpristas, line 488: you dont have to invoke from subdir `/elastic-agent/elastic-agent install ...` should be fine

e2e/_suites/fleet/services.go
```diff
@@ -270,32 +385,196 @@ func newDebianInstaller() (ElasticAgentInstaller, error) {
...
+		args := []string{"--force", "--insecure", "--enrollment-token", token, "--kibana-url", "http://kibana:5601"}
+
+		err = runElasticAgentCommand(profile, image, service, binary, "install", args)
+		if err != nil {
+			return fmt.Errorf("Failed to install the agent with subcommand: %v", err)
+		}
+		return nil
+	}
+	enrollFn := func(token string) error {
```
michalpristas, line 497: enroll is performed out of the box using install so this should not be necessary

e2e/_suites/fleet/fleet.go
```diff
@@ -1004,46 +1025,6 @@ func createDefaultHTTPRequest(url string) curl.HTTPRequest {
 	}
 }
 
-// createFleetPolicy() sends a POST request to Fleet creating a new test policy
```
mdelapenya, line 1007: We are not creating the policy because of https://github.com/elastic/kibana/pull/81236

e2e/_suites/fleet/services.go
mdelapenya: great! that will simplify it, as there is no need to calculate the hash

e2e/_suites/fleet/services.go
```diff
@@ -270,32 +385,196 @@ func newDebianInstaller() (ElasticAgentInstaller, error) {
...
+		return runElasticAgentCommand(profile, image, service, ElasticAgentProcessName, "enroll", args)
+	}
+	postInstallFn := func() error {
+		log.Trace("No postinstall commands for TAR installer")
+		return nil
+	}
+	unInstallFn := func() error {
+		args := []string{"-f"}
+
+		return runElasticAgentCommand(profile, image, service, ElasticAgentProcessName, "uninstall", args)
```
michalpristas, line 509: i believe ElasticAgentProcessName is `elastic-agent`, this should not be only process name but `/opt/Elastic/Agent/elastic-agent`

e2e/_suites/fleet/services.go
mdelapenya: If you take a look, we are providing the enroll command for further usage in the re-enroll scenario. As defined in the life cycle (see PR description), we use install command to initial enroll, but enroll for following ones

e2e/_suites/fleet/services.go
michalpristas: also this should equal to how user runs it

mdelapenya: Done in https://github.com/elastic/e2e-testing/pull/378/commits/8208b567bdc0291ed24ea0eba439238db8bbc797

e2e/_suites/fleet/services.go
mdelapenya: I would expect running the installed linux service `elastic-agent` from PATH

e2e/_suites/fleet/features/fleet_mode_agent.feature
```diff
@@ -94,7 +94,7 @@ Scenario Outline: Un-installing the installed <os> agent
     And the "filebeat" process is in the "stopped" state on the host
     And the "metricbeat" process is in the "stopped" state on the host
     And the file system Agent folder is empty
-    And the agent is listed in Fleet as "inactive"
+    And the agent is listed in Fleet as "offline"
```
mdelapenya, line 97: cc/ @EricDavisX

e2e/_suites/fleet/services.go
mdelapenya: With https://github.com/elastic/e2e-testing/pull/378/commits/017e4d0b203147af7357f218ab7dbe9c38647de9 we are fixing the uninstalling scenario

e2e/_suites/fleet/features/fleet_mode_agent.feature
mdelapenya: It takes +2 minutes in reaching the `offline` status. Is that correct?

michalpristas (approved):
change LGTM

cachedout (approved):

PR #389, opened 2020-10-22
fix: double the timeout to check security events

### LINKED ISSUE #388: Step the policy will reflect the change in the Security App: No endpoint-security events where found for the agent in the policy
This step is consistently failing:

```gherkin
Scenario: Changing an Agent policy is reflected in the Security App
  Given an Endpoint is successfully deployed with a "centos" Agent using "tar" installer
  When the policy is updated to have "malware" in "detect" mode
  Then the policy will reflect the change in the Security App
```

We have noticed that 3 minutes (default timeout) is not enough for this task. We are going to double down the time, as it makes this test to pass locally.

cc/ @EricDavisX @ph

mdelapenya:
## What does this PR do?
It doubles the time used to wait for security events. We are currently checking these conditions:

```go
applicationName := "endpoint-security"
agentID := ""                           // calculated
packagePolicyID := ""                   // calculated
//...
matches := (strings.Contains(message, applicationName) &&
	strings.Contains(message, "["+agentID+"]: State changed to") &&
	strings.Contains(message, "Protecting with policy {"+packagePolicyID+"}"))
```

## Why is it important?
This should fix the CI, as we have seen 3-4 mins delay to get those events.

## Checklist

```shell
$ SUITE="fleet" TAGS="set-policy-and-check-changes" DEVELOPER_MODE=false TIMEOUT_FACTOR=3 LOG_LEVEL=TRACE make -C e2e functional-test
```

## Related issues

- Fixes #388

## Follow-ups
We are noticing that the events are always in the `DEGRADED` state (verified in 7.9.x and master), when ideally they should be in the `RUNNING` state. That's why we are so loose in the string matches.

### DIFF
--- a/e2e/_suites/fleet/fleet.go
+++ b/e2e/_suites/fleet/fleet.go
@@ -796,7 +796,7 @@ func (fts *FleetTestSuite) thePolicyWillReflectTheChangeInTheSecurityApp() error
 		return err
 	}
 
-	maxTimeout := time.Duration(timeoutFactor) * time.Minute
+	maxTimeout := time.Duration(timeoutFactor) * time.Minute * 2
 	retryCount := 1
 
 	exp := e2e.GetExponentialBackOff(maxTimeout)


mdelapenya:
Well, this is glorious, we are going to put CI green again after weeks or a month. 👏 👏 👏 👏 👏 

image

Issue #287, opened 2020-09-14
Document how to run regression tests agains different versions of the test suites


mdelapenya:
Using Jenkins UI for the multiple variables

mdelapenya:
Done in #487

PR #417, opened 2020-11-04
[7.10.x] chore: use official namespace for beats

mdelapenya:
## What does this PR do?
It uses the `beats` namespace for metricbeat and the elastic-agent when running their docker images.

It also bumps Helm version

## Why is it important?
Older versions (i.e. 7.9.2) were not pushed in the CI namespace. Besides that, the 7.9.x branch uses this "released" namespace, so for consistency, we are using it here too.

## Checklist

## Author's Checklist

- [x] The docker images exist in the beats namespace

## Related issues

- Discovered while testing #287

### DIFF
--- a/e2e/_suites/helm/README.md
+++ b/e2e/_suites/helm/README.md
@@ -33,7 +33,7 @@ This is an example of the optional configuration:
 
    ```shell
    # Depending on the versions used, 
-   export HELM_VERSION="3.2.4"        # Helm version: for Helm v2.x.x we have to initialise Tiller right after the k8s cluster
+   export HELM_VERSION="3.2.4"
    export HELM_CHART_VERSION="7.6.1"  # version of the Elastic's Observability Helm charts
    export HELM_KUBERNETES_VERSION="1.18.2" # version of the cluster to be passed to kind
    ```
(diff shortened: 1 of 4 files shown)


v1v (approved):

e2e/_suites/helm/README.md
```diff
@@ -33,14 +33,14 @@ This is an example of the optional configuration:
 
    ```shell
    # Depending on the versions used, 
-   export HELM_VERSION="2.16.3"        # Helm version: for Helm v2.x.x we have to initialise Tiller right after the k8s cluster
+   export HELM_VERSION="3.2.4"        # Helm version: for Helm v2.x.x we have to initialise Tiller right after the k8s cluster
```
v1v, line 36: Will the backport from master might cause some issues with https://github.com/elastic/e2e-testing/pull/419?

mdelapenya: Nope, I'm not backporting that one, because I realised it's partially updating it. I started updating the versions in the maintenance branches thinking that they were properly set in master, which was not the case

mdelapenya, line 36:
```suggestion
   export HELM_VERSION="3.2.4"
```

mdelapenya: remove outdate comment

PR #430, opened 2020-11-09
chore(ci): cache docker images in packer

mdelapenya:
## What does this PR do?
It adds picklesdoc and golang docker images in Packer

## Why is it important?
It will reduce build time

## Checklist

## Author's Checklist

- [x] Checked the new script returns the proper GO_VERSION

## Follow-ups
Identify new images and push updates in the script.

Question: is it possible to run the clean-docker.sh script at packer time? If so, would it be too much size to cache all those images?

### DIFF
--- a/.ci/packer_cache.sh
+++ b/.ci/packer_cache.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+
+_sdir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# shellcheck disable=SC1091
+source /usr/local/bin/bash_standard_lib.sh
+
+readonly GO_VERSION=$(cat ${_sdir}/../.go-version)
+
+DOCKER_IMAGES="docker.elastic.co/observability-ci/picklesdoc:2.20.1
+golang:${GO_VERSION}-stretch
+"
+
+if [ -x "$(command -v docker)" ]; then
+  for di in ${DOCKER_IMAGES}
+  do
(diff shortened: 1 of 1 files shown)


.ci/packer_cache.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+_sdir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# shellcheck disable=SC1091
+source /usr/local/bin/bash_standard_lib.sh
+
+readonly GO_VERSION=$(cat ${_sdir}/../.go-version)
```
v1v, line 8: The script runs in the context of the workspace, therefore if .go-version is in the top level folder, you could simplify the _sdir 

```
 readonly GO_VERSION=$(cat .go-version)
```

v1v (changes requested):

.ci/packer_cache.sh
mdelapenya, line 8:
```suggestion
# shellcheck disable=SC1091
source /usr/local/bin/bash_standard_lib.sh

readonly GO_VERSION=$(cat .go-version)
```

cachedout (approved):

kuisathaverat (approved):

mdelapenya:
@kuisathaverat are we usually backporting packer cache scripts to the maintenance branches?

kuisathaverat:
> @kuisathaverat are we usually backporting packer cache scripts to the maintenance branches?

no, the script is only grabbed from the master so you do not need to backport the change

PR #480, opened 2020-11-30
fix: honor ELASTIC_AGENT_VERSION when downloading the agent

mdelapenya:
## What does this PR do?
It checks if the environment variable used to override Agent's version is set. If it's not set, then it will use the base version, which is obtained from:

a. the Elastic Artifacts URL for aliases (7.x, 7.10.x),
b. fixed to the maintenance branch
c. 8.0.0-SNAPSHOT in the master branch

## Why is it important?
I found that running:

```shell
SUITE="fleet" \
    TAGS="fleet_mode && enroll" DEVELOPER_MODE=true \
    TIMEOUT_FACTOR=3 LOG_LEVEL=TRACE \
    ELASTIC_AGENT_VERSION="7.10.1" \
    make -C e2e functional-test
```
was always downloading the version set to the maintenance branch, not honoring the version set in the environment variable.

Because we evaluated the version just once (init method for the test suite), then further calculations were tied to thee base version.

## Checklist

- [x] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have run the Unit tests for the CLI, and they are passing locally
- [x] I have run the End-2-End tests for the suite I'm working on, and they are passing locally
- [ ] I have noticed new Go dependencies (run `make notice` in the proper directory)

## How to test this PR locally

Download a different agent version (7.10.1)
```shell
git checkout master
SUITE="fleet" \
    TAGS="fleet_mode && enroll" DEVELOPER_MODE=true \
    TIMEOUT_FACTOR=3 LOG_LEVEL=TRACE \
    ELASTIC_AGENT_VERSION="7.10.1" \
    make -C e2e functional-test
```

Download the default agent version (8.0.0-SNAPSHOT)
```shell
git checkout master
SUITE="fleet" \
    TAGS="fleet_mode && enroll" DEVELOPER_MODE=true \
    TIMEOUT_FACTOR=3 LOG_LEVEL=TRACE \
    make -C e2e functional-test
```

## Related issues

- Relates #403, #306
- Discovered while testing #287

PR #489, opened 2020-12-01
fix: honor ELASTIC_AGENT_VERSION when downloading the agent (#480) backport for 7.10.x

mdelapenya:
 - fix: honor ELASTIC_AGENT_VERSION when downloading the agent (#480)


mdelapenya:
I know why this is failing: because we are setting the ELASTIC_AGENT_VERSION to `7.10-SNAPSHOT` in the Jenkinsfile, it's directly used to get the variable, but it's indeed an alias, not a version.

I see two options here:

a. check if the version set in the environment is an alias (more code to maintain)
b. remove the default value in the Jenkinsfile, honouring the base version (simplify the Jenkinsfile, setting default values to empty, making version bumps easier: less locations for the version). My option is this one 🤚 

@elastic/observablt-robots thoughts?

kuisathaverat (approved):

cachedout (approved):

mdelapenya:
As discussed offline, I'll set the Jenkinsfile parameters with empty values

mdelapenya:
> As discussed offline, I'll set the Jenkinsfile parameters with empty values

Upsss... after setting that variable as empty in the pipeline, the job failed with:

>[2020-12-02T16:17:27.052Z] ERROR: groovy.lang.MissingPropertyException: No such property: METRICBEAT_VERSION for class: groovy.lang.Binding

mdelapenya:
jenkins run the tests please

Issue #341, opened 2020-10-02
[Elastic Agent] Agent Upgrade feature test


michalpristas:
Scenario should be as follow:
#### 2 versions of agent are available (must differ in last commit)
  - first agent is the one which is installed
  - second one is made available using some public API or local fileserver (must be available at `{uri}:{port}/beats/elastic-agent/{tar_name}`

#### Second agent artifacts is made available 
for local test i'm using fileserver with artifacts at `~/server/beats/elastic-agent/*`
then from `~/server` i run `python3 -m http.server`

#### Update is initiated either using UI (manual tests) or using API (manual/automated tests)
API for initiating upgrade using fileserver serving at localhost
```
curl -X POST -H 'kbn-xsrf: true'  -u elastic:changeme http://localhost:5601/api/ingest_manager/fleet/agents/{AGENT_ID}/upgrade \
  -H 'Content-Type: application/json' \
 -d '{"source_uri":"http://localhost:8000","version":"8.0.0"}'
```

If fleet does not send SourceURI it is good to configure `agent.download.sourceURI` to `http://localhost:8000` in `elastic-agent.yml`

#### Agent is listed as `UPDATING` in fleet 
- with file server this will be only for a fraction of second or couple of seconds max (update is very fast)

#### Agent version is changed and agent is listed as `ONLINE` in fleet 
- state should be running
- all beats has to be started and collecting again based on config assigned

No ERROR should be reported during this procedure.

### Prerequisites 
https://github.com/elastic/kibana/pull/78810
https://github.com/elastic/beats/pull/21461

EricDavisX:
Thanks Michal!  Do you want me to take a first stab at the feature file?  Manu and I are happy to review.  to start... I'll list out what I see, and post it to the desc above which we can modify. I have a plan I think may be fairly feasible to implement.  The syntax below is contingent on the test being able to introspect what the current snapshot version is and what the prior version is.  I think we can do that.  And the vague 'current' and 'prior' citation allows us to write actual tests in a given GA line that upgrade the current shipped version to the current-in-test version (I think).  Anyhow, below is a first pass draft to consider.  I did (painfully) add back in the steps for the filebeat, metric, and endpoint about 'sending new data' to ES.  we took those steps out, and I submit we shouldn't wait to put them back in, we just needed better depth of understanding to do it and Michal, I think you can help us out here!

file:
agent_upgrades.feature

@agent_endpoint_upgrade
Feature: Agent Endpoint Upgrade tests
[...]

PR #481, opened 2020-11-30
fix: update log path for TAR installer

mdelapenya:
## What does this PR do?
It removes the `logs/` path from the log file filepath.

## Why is it important?
We are _catting_ the log file when an step fails, and this was not failing, although not failing the build (as intended).

## Checklist

- [x] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have run the Unit tests for the CLI, and they are passing locally
- [x] I have run the End-2-End tests for the suite I'm working on, and they are passing locally
- [ ] I have noticed new Go dependencies (run `make notice` in the proper directory)

## Author's Checklist

- [x] Checked with @michalpristas the log file path

## Related issues

- Fixes #479

PR #439, opened 2020-11-18
work for issue #379 - adding back rpm and deb tests

EricDavisX:
## What does this PR do?
2 things:
1) very small change to rename the @ tag to be more accurate per the tests being done and to sync 2 file names to the @ tag name for the whole file (I think this is our best practice pattern?)

2) most primarily, this is work for https://github.com/elastic/e2e-testing/issues/379 - adding in the deb and rpm tests against centos & debian images that were removed prior.

I used existing .go implementation code that remained and which I think is functional and updated so far, only feature files.

## Why is it important?
we need coverage over rpm and deb installs, even tho they are no longer the recommended usage - they are still supported and expected to be used.

In 379 you can see the discussion, we removed this accidentally / temporarily as we implemented new support for 'install' feature.

## Checklist

- [x] My code follows the style guidelines of this project
~~- [ ] I have commented my code, particularly in hard-to-understand areas~~
~~- [ ] I have made corresponding changes to the documentation~~
~~- [ ] I have made corresponding change to the default configuration files~~
~~- [ ] I have added tests that prove my fix is effective or that my feature works~~
~~- [ ] I have run the Unit tests for the CLI, and they are passing locally~~
- [x] I have run the End-2-End tests for the suite I'm working on, and they are passing locally~~
~~- [ ] I have noticed new Go dependencies (run `make notice` in the proper directory)~~

## Author's Checklist

I actually am opening this as a draft to get some help from @mdelapenya or perhaps @blakerouse if we need

## How to test this PR locally
OP_LOG_LEVEL=DEBUG DEVELOPER_MODE=true godog -t enroll > test-debug.log

## Related issues

- Relates https://github.com/elastic/beats/pull/22543/files

## Follow-ups

So, Centos is working as submitted here,  But Debian is throwing an error.. and I am not 100% sure I'm tracing the go code we've implemented right to know what question to ask.  I'll put logs and more details in the comments and we can figure it out I hope.

PR #506, opened 2020-12-02
Introduce tests for elastic-agent upgrade procedure

michalpristas:
## What does this PR do?

Just sharing WIP of Upgrade tests 

```bash
SUITE=fleet DEVELOPER_MODE=true TAGS="fleet_mode && upgrade-agent" LOG_LEVEL=TRACE make -C e2e functional-test > test.upgrade.out
```

## Checklist

cc #341


e2e/_suites/fleet/ingest-manager_test.go
```diff
@@ -43,6 +43,10 @@ var agentVersionBase = "8.0.0-SNAPSHOT"
 // It can be overriden by ELASTIC_AGENT_VERSION env var
 var agentVersion = agentVersionBase
 
+// agentStaleVersion is the version of the agent to use as a base during upgrade
+// It can be overriden by ELASTIC_AGENT_STALE_VERSION env var. Using latest GA as a default.
+var agentStaleVersion = "7.10.0"
```
mdelapenya, line 48: Could you add it to the Jenkinsfile as an input parameter? This way we will allow manual tests on CI simply setting that input parameter in Jenkins UI

e2e/_suites/fleet/services.go
```diff
@@ -497,8 +498,8 @@ func newTarInstaller(image string, tag string) (ElasticAgentInstaller, error) {
 		image:             image,
 		InstallFn:         installFn,
 		installerType:     "tar",
-		logFile:           "elastic-agent.log",
-		logsDir:           "/opt/Elastic/Agent/data/elastic-agent-%s/",
+		logFile:           "elastic-agent-json.log",
+		logsDir:           "/opt/Elastic/Agent/data/elastic-agent-%s/logs/",
```
mdelapenya, line 502: Could you revisit my comment here? https://github.com/elastic/e2e-testing/pull/481#discussion_r533246796

I'd prefer using plain log for readability concerns

michalpristas: if it's ok by you then ok, i assumed that for computer processing json is more suitable.

mdelapenya: Totally agree, although in this case, we are not sending the raw container logs to any other system (it's on our roadmap).

What we are sending to an ES cluster is Jenkins build logs, and because the container log is simply outputted to Jenkins logs, it's stored in the same format as the Jenkins log in our logs cluster.

e2e/_suites/fleet/features/fleet_mode_agent.feature
```diff
@@ -26,6 +26,17 @@ Examples:
 | centos | tar       |
 | debian | tar       |
 
+@upgrade-agent
+Scenario Outline: Upgrading the installed <os> agent
```
EricDavisX, line 30: I have a PR coming https://github.com/elastic/e2e-testing/pull/439 that will update the whole file so linting passes.  you'll find that you can't use the | tar | usage phrasing as it is the only variable used.  Just replace it in the Gherkin with "tar".

EricDavisX (commented):
Thanks for pushing up the draft!

e2e/_suites/fleet/features/fleet_mode_agent.feature
michalpristas: let push your issue i deal with conflicts and update file afterwards

EricDavisX: done

e2e/_suites/fleet/features/fleet_mode_agent.feature
```diff
@@ -40,6 +40,17 @@ Examples:
 | centos |
 | debian |
 
+@upgrade-agent
+Scenario Outline: Upgrading the installed <os> agent
+  Given a "<os>" agent "stale" is deployed to Fleet with "tar" installer
+    And certs for "<os>" are installed
```
EricDavisX, line 46: is there value in leaving this separate?  you could maybe just copy the Go code into the end of the above installer step if we're never ever going to run tests for this feature without certs (I guess that would be a negative test to prove it fails? maybe not helpful...)

e2e/_suites/fleet/features/fleet_mode_agent.feature
```diff
@@ -40,6 +40,17 @@ Examples:
...
+  When agent is upgraded to version "latest"
+  Then process waits for "2m"
```
EricDavisX, line 48: when you are ready... a proper polling state is likely quicker and more stable than the long 2 min wait here.

e2e/_suites/fleet/features/fleet_mode_agent.feature
```diff
@@ -40,6 +40,17 @@ Examples:
...
+    And agent is in version "latest"
```
EricDavisX, line 49: this is great - after this I would like to see some of the other validation steps copied in, to make sure it is sending docs in to ES.  ... and I realize we may not have that for Fleet cases (only for stand-alone) so we may leave this and add it as a separate item if we must.

e2e/_suites/fleet/ingest-manager_test.go
EricDavisX: Adding it to Jenkins as an optional param is even better... where it uses the value if passed... but if not, we can do some parse and math logic to get the current-minus-1 version (from the stack deployed or other e2e-testing variable) and use that instead.  It is more future proof and will require less maintenance, right?  

It is worth some discussion, open to what ever you want to merge, and if not tackled now, we can just drop a ticket in for the reminder of the work.

e2e/_suites/fleet/fleet.go
```diff
@@ -138,6 +145,123 @@ func (fts *FleetTestSuite) contributeSteps(s *godog.Suite) {
 	s.Step(`^the policy will reflect the change in the Security App$`, fts.thePolicyWillReflectTheChangeInTheSecurityApp)
 }
 
+func (fts *FleetTestSuite) anStaleAgentIsDeployedToFleetWithInstaller(image, version, installerType string) error {
+	agentVersionBackup := agentVersion
+	defer func() { agentVersion = agentVersionBackup }()
```
mdelapenya, line 150: I liked this post-condition style!

e2e/_suites/fleet/features/fleet_mode_agent.feature
michalpristas: i wanted to make it separate as having this in installer would mean that it would get executed every time installer is prepared and usually there's no need for that, having this separate dont slow down tests

michalpristas:
@mdelapenya updated steps and code to make it nicer, can we get this in?

e2e/_suites/fleet/features/fleet_mode_agent.feature
EricDavisX: that is sound argument, but I don't see the 'Given' statement used elsewhere. :)  it is ok, no harm done, lets leave it as you've written

e2e/_suites/fleet/features/fleet_mode_agent.feature
EricDavisX: its a really weak validation as it is, no proof of what Agent is or can do after it is upgraded.  But it is a start, nothing wrong with incremental improvements

EricDavisX:
There were some smaller debates, but code fine as is with one item I did want to confirm about branch support over time vs hard-coded (Agent) file version (7.10.0).  

Did we run the other branches yet to see how they would work... we can test them in the backport PRs I guess.

I'm trying to figure if we want to take the low-fi approach and just update the stale-agent value when needed... I suppose over time the worst that will happen is that it uncovers an upgrade bug that we won't care to fix (too old, unsupported case) or maybe if we aren't testing on new enough versions routinely then we may miss an upgrade bug for whatever reason from more recent 'stale' version to current.  The fix in this case is just to update the version in the different branches when those branches are created, we just need to remember to do that, that is if we don't put in some dynamic logic.

EricDavisX:
Michal and I discussed and we'll merge as is.  And I logged a ticket for the remaining work and ideas to resolve the version problem with dynamic version assessment.

e2e/_suites/fleet/features/fleet_mode_agent.feature
mdelapenya: I'd advocate for hiding this technical detail from the specification, and handle it in code, otherwise we will be polluting a reader's mind when understanding the behavior.

OTOH great job here @michalpristas !!!

Issue #373, opened 2020-10-19
Refactor Gherkin tag names


mdelapenya:
Let's use more concise names for tags, trying to reuse tags across different feature files in the same test suite

mdelapenya:
@EricDavisX we discussed yesterday about the desired capability of adding tags by examples. Well, Godog supports it!! https://github.com/cucumber/godog/blob/5c352074bcecb39c9dd17182410be9ea0c081a20/features/tags.feature#L14-L26

We would be able to have something like this, everywhere:

```gherkin
@revoke-token
Scenario Outline: Revoking the enrollment token for the <os> agent
  Given a "<os>" agent is deployed to Fleet with "<installer>" installer
  When the enrollment token is revoked
  Then an attempt to enroll a new agent fails

@centos
Examples:
| os     | installer |
| centos | tar       |

@debian
Examples:
| os     | installer |
| debian | tar       |
```

PR #507, opened 2020-12-02
feat: support more integrations

mdelapenya:
## What does this PR do?
It adds the ability to fetch Beats' config files instead of maintaining a local version of the config file, so the sync integrations CMD will do this for you.

When getting the config files for the integrations, it will "massage" the files changing any occurrence for "localhost" or "127.0.0.1", replacing it with the service name. We do this for the service to be discovered under the docker-compose network.

>Meanwhile Beats PRs are reviewed, I added specific sanitizers for MySQL and dropwizard, so that we are able to add progress. The MySQL one replaces `root:secret` with `root:test`, and the second, and dropwizard's, replaces `/metrics/metrics` with `/test/metrics`.

Finally we are adding the integrations supporting multiple versions (those with the supported-versions.yml descriptor), but we are not adding support for nats and uwsgi, because the `extends` clause fails when composing several compose files with `-f` (needs to investigate). We are adding them as a parallel branch on CI too.

## Why is it important?
It will simplify the addition of new integrations, reducing the amount of code in the form of config files to maintain. To add a new integration:

1. Add a new examples block, including integration name and versions.
1. Add to Jenkins' parallel branch, using the YML descriptor.

## Checklist

## Author's Checklist

- [x] Executed all tests for all integrations
- [x] Added that all integrations with supported versions under Beats' metricbeat/module/.../_meta/supported-versions.yml
- [x] Verified that all integrations has config.yml file under Beats' metricbeat/module/.../_meta
- [x] Ran existing integrations and they works as previously

Update the integration name for each new one, in lowercase.

```shell
export INTEGRATION_NAME=redis
SUITE="metricbeat" \
    TAGS="integrations && $INTEGRATION_NAME" DEVELOPER_MODE=true \
    TIMEOUT_FACTOR=3 LOG_LEVEL=TRACE \
    make -C e2e functional-test
```

## Related issues

- Relates #373
- Related to elastic/beats#22906
- Related to elastic/beats#22934

## Follow-up

We expect some of them to fail, because the tests will find errors in the index


kuisathaverat (approved):

cachedout (approved):

mdelapenya:
@jsoriano not sure if you can take a look at the following errors, which happen consistently (I skipped both integrations in last two commits):

### Kafka
>error in connect: getting cluster client for advertised broker with address kafka:9092: kafka: client has run out of available brokers to talk to (Is your cluster reachable?)

#### Dropwizard
>HTTP error 404 in : 404 Not Found

mdelapenya:
I found the root cause for dropwizard:

> HTTP error 404 in

The config file sets the `metrics_path:` to `/metrics/metrics`, where it should point to `/test/metrics`. See Dockerfile https://github.com/elastic/beats/blob/master/metricbeat/module/dropwizard/_meta/Dockerfile#L15

jsoriano:
@mdelapenya the kafka image needs to know the host it should advertise, there are three ways for that:
* Set the host:port directly in `KAFKA_ADVERTISED_HOST` environment variable.
* Set `KAFKA_ADVERTISED_HOST_AUTO=true` environment variable, then it uses its hostname (`$(dig +short $HOSTNAME):9092`).
* If none of them are set, it waits for a file `/run/compose_env` containing the `KAFKA_ADVERTISED_HOST` variable. This can be used when the host name is not known before hand. We do this to be able to run integration tests locally in beats with `go test`, as then the tests are not run in a docker network.

For this case it will be probably enough with setting `KAFKA_ADVERTISED_HOST=kafka:9092`.

e2e/_suites/metricbeat/features/integrations.feature
```diff
@@ -14,6 +14,44 @@ Examples: Apache
...
+Examples: HAProxy
+| integration | version |
+| haproxy     | 2.0.9   |
+| haproxy     | 1.9.12  |
+| haproxy     | 1.8.22  |
+| haproxy     | 1.7.12  |
+| haproxy     | 1.6.15  |
+
+@kafka
+@skip
```
EricDavisX, line 46: is this the appropriate way to put in multiple tags?!  Neat, I had thought to put them side by side, I didn't think to stack them.

Also, when putting in a skip a common practice is to put a comment with the relating issue, so it easier to review and un-skip when its fixed (especially if it is forgotten about).

mdelapenya: I always trust the guy commit causing the line change, in this case the one adding the skip tag

mdelapenya:
> KAFKA_ADVERTISED_HOST=kafka:9092

Do you think this should be added to the original compose file? or something the tests should provide?

jsoriano:
> Do you think this should be added to the original compose file? or something the tests should provide?

The original compose file is used on scenarios that use the host network, where the `kafka` hostname doesn't exist, so if we set this by default we would need to modify kafka tests in the beats repo to remove this variable so it goes back to the `/run/compose_env` approach.

So we have two possible paths to solve this:
* Add the `KAFKA_ADVERTISED_HOST=kafka:9092` in the E2E tests.
* Modify the docker files and the tests (python and go) in the beats repo so `KAFKA_ADVERTISED_HOST=kafka:9092` is used by default, but other approaches can be defined so it still works in the host network.

The second option will require some time, kafka is quite tricky to test.

mdelapenya:
> > Do you think this should be added to the original compose file? or something the tests should provide?
> 
> The original compose file is used on scenarios that use the host network, where the `kafka` hostname doesn't exist, so if we set this by default we would need to modify kafka tests in the beats repo to remove this variable so it goes back to the `/run/compose_env` approach.
> 
> So we have two possible paths to solve this:
> 
> * Add the `KAFKA_ADVERTISED_HOST=kafka:9092` in the E2E tests.
> * Modify the docker files and the tests (python and go) in the beats repo so `KAFKA_ADVERTISED_HOST=kafka:9092` is used by default, but other approaches can be defined so it still works in the host network.
> 
> The second option will require some time, kafka is quite tricky to test.

We are adding sanitizers in this PR, which replaces localhost and 127.0.0.1 with service name. We could also add a specific saniticer for Kafka, as we did for MySQL: https://github.com/elastic/e2e-testing/pull/507/files#diff-83ea776352dff2d01db5b056913e36608e5034252afa73f85015d0c82a6edf19R48

OTOH, we could add an EnvironmentProvisioner struct following same approach, i.e. for Kafka, where we set up the environment with according variables.

jsoriano:
> We are adding sanitizers in this PR, which replaces localhost and 127.0.0.1 with service name. We could also add a specific saniticer for Kafka, as we did for MySQL: https://github.com/elastic/e2e-testing/pull/507/files#diff-83ea776352dff2d01db5b056913e36608e5034252afa73f85015d0c82a6edf19R48
> 

I understand that this sanitization is for the configuration used by Metricbeat? This wouldn't be enough. One of the acctions mentioned in https://github.com/elastic/e2e-testing/pull/507#issuecomment-739894942 need to be taken so the kafka service reaches a healthy state.

> OTOH, we could add an EnvironmentProvisioner struct following same approach, i.e. for Kafka, where we set up the environment with according variables.

Yes, something like this should work (given that metricbeat and kafka are in the same network and kafka is reachable using the `kafka` hostname).

mdelapenya:
Ok, let me work on that. I'm in PTO but will try to send it ASAP. Thanks!!

mdelapenya:
The existing errors are happening in master, so they are not related.

@jsoriano if you agree, I'd merge this one to provide support for more integrations! 🚀 

About the skipped ones:

- **Redis Enterprise** fails with the following message:
>unable to decode response from prometheus endpoint: error making http request: Get \"https://redisenterprise:8070/\": x509: certificate is valid for cluster.local, *.cluster.local, not redisenterprise

- **Oracle**, even though we are manually installing oracle tools in the container, it fails with the following message:
>error creating connection to Oracle: error doing ping to database: ORA-00000: DPI-1047: Cannot locate a 64-bit Oracle Client library: \"libclntsh.so: cannot open shared object file: No such file or directory\". See https://oracle.github.io/odpi/doc/installation.html#linux for help

e2e/_suites/metricbeat/metricbeat_test.go
```diff
@@ -280,6 +301,45 @@ func (mts *MetricbeatTestSuite) installedUsingConfiguration(configuration string
 	return nil
 }
 
+func (mts *MetricbeatTestSuite) installOracleTools() error {
```
jsoriano, line 304: If this is not working, wdyt about moving it to a different PR?

jsoriano (approved):

e2e/_suites/metricbeat/metricbeat_test.go
mdelapenya: At the end of the day, oracle is skipped, but I like the idea of not having dead code. Let me do it

jsoriano:
> * **Kafka**: thinking about how to populate the environment in the kafka container. Maybe massaging the docker-compose file coming from Beats, to add the env vars needed for Kafka discovery.
> 

Perhaps something easy that what we can try to do is to add `KAFKA_ADVERTISED_HOST` as an empty environment variable in the `docker-compose.yml` file in Beats. Then maybe we can use an [env file](https://docs.docker.com/compose/environment-variables/) here.

> * **Redis Enterprise** fails with the following message:
> 
> 
> > unable to decode response from prometheus endpoint: error making http request: Get "[https://redisenterprise:8070/\](https://redisenterprise:8070/%5C)": x509: certificate is valid for cluster.local, *.cluster.local, not redisenterprise
> 

In beats we set [`ssl.verification_mode: none` in the config](https://github.com/elastic/beats/blob/e5bae4ce5bd2508ae2d40ecb762be072c29a1ece/x-pack/metricbeat/module/redisenterprise/node/node_integration_test.go#L46). Could this option be also added here?

> * **Oracle**, even though we are manually installing oracle tools in the container, it fails with the following message:
> 
> 
> > error creating connection to Oracle: error doing ping to database: ORA-00000: DPI-1047: Cannot locate a 64-bit Oracle Client library: "libclntsh.so: cannot open shared object file: No such file or directory". See https://oracle.github.io/odpi/doc/installation.html#linux for help

According to these docs it seems that ld needs to be configured to find the library even if `yum` is used. They propose two methods: adding a file to `/etc/ld.so.conf.d/` and running `ldconfig`, or using the `LD_LIBRARY_PATH` environment variable. Did you try these options?

mdelapenya:
Adding the patch for the oracle tools removal for future reference:

```patch
From e9941d3f510e81e07e612a58b03de4f259ae5c7c Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?Manuel=20de=20la=20Pe=C3=B1a?= <mdelapenya@gmail.com>
Date: Mon, 4 Jan 2021 19:08:18 +0100
Subject: [PATCH] chore: remove custom code for oracle tools

Because it's skipped, we do not want to maintain dead-code. The failure
is very likely related to the environment, as with this instructions it
should work: see https://oracle-base.com/articles/misc/oracle-instant-client-installation#yum

What we observe is that the OS is not able to load a library for connecting
to Oracle
---
 e2e/_suites/metricbeat/metricbeat_test.go | 44 -----------------------
 1 file changed, 44 deletions(-)

diff --git a/e2e/_suites/metricbeat/metricbeat_test.go b/e2e/_suites/metricbeat/metricbeat_test.go
index 521725f1..aab71a8a 100644
--- a/e2e/_suites/metricbeat/metricbeat_test.go
+++ b/e2e/_suites/metricbeat/metricbeat_test.go
@@ -15,7 +15,6 @@ import (
 	"github.com/cucumber/godog"
 	messages "github.com/cucumber/messages-go/v10"
 	"github.com/elastic/e2e-testing/cli/config"
-	"github.com/elastic/e2e-testing/cli/docker"
 	"github.com/elastic/e2e-testing/cli/services"
 	"github.com/elastic/e2e-testing/cli/shell"
 	"github.com/elastic/e2e-testing/e2e"
@@ -301,45 +300,6 @@ func (mts *MetricbeatTestSuite) installedUsingConfiguration(configuration string
 	return nil
 }
 
-func (mts *MetricbeatTestSuite) installOracleTools() error {
-	containerName := "metricbeat_oracle_1"
-
-	log.WithFields(log.Fields{
-		"containerName": containerName,
-	}).Trace("Installing Oracle tools")
-
-	// see https://oracle-base.com/articles/misc/oracle-instant-client-installation#yum
-	cmds := [][]string{
-		[]string{"yum", "install", "wget", "-y"},
-		[]string{"rm", "-f", "/etc/yum.repos.d/public-yum-ol7.repo"},
-		[]string{"wget", "https://yum.oracle.com/public-yum-ol7.repo", "-O", "/etc/yum.repos.d/public-yum-ol7.repo"},
-		[]string{"yum", "install", "yum-utils", "-y"},
-		[]string{"yum-config-manager", "--enable", "ol7_oracle_instantclient"},
-		[]string{"yum", "install", "oracle-release-el7", "-y"},
-		[]string{"/usr/bin/ol_yum_configure.sh"}, // transition to new Oracle Linux yum server repository
-		[]string{"yum", "install", "oracle-instantclient19.6-basic", "-y"},
-	}
-
-	for _, cmd := range cmds {
-		_, err := docker.ExecCommandIntoContainer(context.Background(), containerName, "root", cmd)
-		if err != nil {
-			log.WithFields(log.Fields{
-				"command":       cmd,
-				"containerName": containerName,
-				"error":         err,
-			}).Error("Could not install Oracle tools using the Docker client")
-			return err
-		}
-	}
-
-	log.WithFields(log.Fields{
-		"commands":      cmds,
-		"containerName": containerName,
-	}).Debug("Oracle tools installed")
-
-	return nil
-}
-
 // runMetricbeatService runs a metricbeat service entity for a service to monitor it
 func (mts *MetricbeatTestSuite) runMetricbeatService() error {
 	// this is needed because, in general, the target service (apache, mysql, redis) does not have a healthcheck
@@ -435,10 +395,6 @@ func (mts *MetricbeatTestSuite) serviceIsRunningForMetricbeat(serviceType string
 	mts.ServiceName = serviceType
 	mts.ServiceVersion = serviceVersion
 
-	if serviceType == "oracle" {
-		err = mts.installOracleTools()
-	}
-
 	return err
 }
 ```

mdelapenya:
@jsoriano I'm gonna merge this one, as it passes and adds a lot of value to any PR from Beats affecting integrations.

As a follow-up, I'm planning to add 6.8.x branch, so that we add coverage to the integrations in that branch. See #574

PR #514, opened 2020-12-02
chore: use empty values for beats versions in Jenkinsfile (#513) backport for 7.x

mdelapenya:
 - chore: use empty values for beats versions in Jenkinsfile (#513)


mdelapenya:
We reverted #513, closing until finding root cause

PR #516, opened 2020-12-02
chore: use empty values for beats versions in Jenkinsfile (#513) backport for 7.10.x

mdelapenya:
 - chore: use empty values for beats versions in Jenkinsfile (#513)


mdelapenya:
Discard until #513 is fixed

PR #540, opened 2020-12-15
update-to-available-version

EricDavisX:
## What does this PR do?
updates minor version in string, so it can be found and that agent artifact used in test
## Screenshots

Screen Shot 2020-12-15 at 2 52 20 PM

need to check if the upgrade test was ever back-ported out of master, it may not have been.  also, it is disabled as of now - hence why the tests are not failing on master due to this.  a future PR could enhance the api call used to not require regular maintenance like this... but we likely won't have many 7.10 hot-fix patch releases so it matters only a little for now


EricDavisX:
there is one test failing and it is unrelated to this pr change as it was seen passing in the first run before updating master branch in.  so i'm merging.

PR #542, opened 2020-12-16
unskip deb and rpm test

EricDavisX:
When this was fixed, we should have unskipped the relating tests that were recently re-added to test this.
issue here: https://github.com/elastic/beats/issues/22744

I plan to use the CI to test it.  Will report back after test run (hopefully) passes.

PR #544, opened 2020-12-17
unskip deb and rpm test (#542) backport for 7.x

EricDavisX:
 - unskip deb and rpm test (#542)


EricDavisX:
the one failed test has nothing to do with unskipping this test which is passing so i'm merging it.

PR #551, opened 2021-01-04
[7.x] backport #537

mdelapenya:
## What does this PR do?
It cherry-picks all commits in that PR, manually resolving conflicts.

## Why is it important?
Keep 7.x branch up-to-date

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have run the Unit tests for the CLI, and they are passing locally
- [ ] I have run the End-2-End tests for the suite I'm working on, and they are passing locally
- [ ] I have noticed new Go dependencies (run `make notice` in the proper directory)

## Author's Checklist

- [ ]

## Related issues

- Backports #537 to 7.x
- Relates #548

## Follow-ups

Once merged using squash, try to backport that resulting commit to 7.10.x too

Issue #548, opened 2021-01-04
Backport https://github.com/elastic/e2e-testing/pull/506


mdelapenya:
#506 was not backported to 7.x nor 7.10.x, so it should be done before we create 7.11.x

mdelapenya:
Fixed, cool!

![image](https://media.giphy.com/media/AvMJCeu1EMmhG/giphy.gif)

PR #537, opened 2020-12-10
disable upgrade tests

michalpristas:
## What does this PR do?

Disables upgrade tests 

## Why is it important?

Upgrade test fails on PR on 
```
[2020-12-10T17:53:01.565Z] time="2020-12-10T17:53:00Z" level=error msg="Could not download the binary for the agent" arch=x86_64 artifact=elastic-agent error="Reached the end of the pages and the snapshots/elastic-agent-7.10.0-x86_64.tar.gz object was not found for the beats-ci-artifacts bucket" extension=tar.gz os=linux version=7.10.0

[2020-12-10T17:53:01.565Z] time="2020-12-10T17:53:00Z" level=fatal msg="Sorry, we could not download the installer" error="Reached the end of the pages and the snapshots/elastic-agent-7.10.0-x86_64.tar.gz object was not found for the beats-ci-artifacts bucket" image=debian installer=tar

```

## Checklist

- [x] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have run the Unit tests for the CLI, and they are passing locally
- [ ] I have run the End-2-End tests for the suite I'm working on, and they are passing locally
- [ ] I have noticed new Go dependencies (run `make notice` in the proper directory)

PR #558, opened 2021-01-05
[7.x] Backport #537

mdelapenya:
## What does this PR do?
It cherry-picks all commits in that PR, manually resolving conflicts.

## Why is it important?
Keep 7.x branch up-to-date

## Checklist

## Author's Checklist

- [ ]

## Related issues

- Backports #537 to 7.x
- Relates #548
- Superseded #551

## Follow-ups

Once merged using squash, try to backport that resulting commit to 7.10.x too


.ci/Jenkinsfile
```diff
@@ -35,6 +35,7 @@ pipeline {
     string(name: 'SLACK_CHANNEL', defaultValue: 'observablt-bots', description: 'The Slack channel(s) where errors will be posted. For multiple channels, use a comma-separated list of channels')
     string(name: 'ELASTIC_AGENT_DOWNLOAD_URL', defaultValue: '', description: 'If present, it will override the download URL for the Elastic agent artifact. (I.e. https://snapshots.elastic.co/7.12.0-069dfaa4/downloads/beats/elastic-agent/elastic-agent-7.12.0-SNAPSHOT-linux-x86_64.tar.gz')
     string(name: 'ELASTIC_AGENT_VERSION', defaultValue: '7.x-SNAPSHOT', description: 'SemVer version of the stand-alone elastic-agent to be used for Fleet tests. You can use here the tag of your PR to test your changes')
+    string(name: 'ELASTIC_AGENT_STALE_VERSION', defaultValue: '7.10.0', description: 'SemVer version of the stale stand-alone elastic-agent to be used for Fleet upgrade tests.')
```
mdelapenya, line 38: @michalpristas what version should we use as stale in 7.x? Same question will come up for the backport to 7.10.x

mdelapenya:
@michalpristas this PR backports your PR. I'll squash the commits and do same backport to 7.10.x. Do you think that branch needs to receive the backport?

v1v (approved):

cachedout (approved):

kuisathaverat (approved):

.ci/Jenkinsfile
michalpristas: 7.x i would keep previous stable 
for 7.10.x we could do 7.10.(x-1) or 7.10.0? what do you think?

mdelapenya: IIUC, the stale version is a previous, valid version, right? So:

- For `7.x` branch: current=7.x-SNAPSHOT and stale=
- For `7.10.x` branch: current=7.10-SNAPSHOT and stale=

Because we are using the alias, which holds a reference to the current moving target, how do we get that (-1)? There is a method to retrieve the current version: `GetElasticArtifactVersion` (see https://github.com/elastic/e2e-testing/commit/4c1cea547851e087ccc47f71fe60bbf038c37989#diff-04a1accd1c12dafffe8d3ecc43a623cd0bf22d03d3561f7e04e296e64962048eR121). What would be the algorithm to rest one to that version? Do you foresee any corner case?

mdelapenya: As discussed in today's conversation, @EricDavisX and I decided we are not going to backport the upgrade tests to 7.10.x

mdelapenya: BTW, @michalpristas @EricDavisX I'd see that in `master` branch, the value of the stale agent it should be 7.x-SNAPSHOT, which is `((master-1) == (8.0.0-SNAPSHOT - 1))`. Wdyt?

EricDavisX:
I agree with 7.x using current minor version minus one as the stale version.  
 - Which means 7.x = 7.12 (currently) so the stale is 7.11 latest snapshot agent.

The exceptions are 7.10 and 8.0.
7.10 can use the .patch version minus 1. 
 - so in this case it is 7.10.x = 7.10.2 and the stale version can be hardcoded to 7.10.1 (only because this is the first minor version where upgrade support exists)

8.0 is 'master' and it can use the 7.x latest major for the 'stale' agent.  If it used the latest minus one (if easier) that is ok too, to avoid too much logic here - what ever is quickest is ok by me.

.ci/Jenkinsfile
mdelapenya, line 38: @EricDavisX @michalpristas is this correct then?
```suggestion
    string(name: 'ELASTIC_AGENT_STALE_VERSION', defaultValue: '7.11.0', description: 'SemVer version of the stale stand-alone elastic-agent to be used for Fleet upgrade tests.')
```

mdelapenya: Ready to merge this if you agree

EricDavisX: looks good

PR #624, opened 2021-01-20
chore: increase timeout in CI

mdelapenya:
## What does this PR do?
It uses 5 as timeout_factor on CI, which means that instead of max 3 minutes of backoff wait,
it will be 5. The time magnitude will be the same: seconds, minutes.

## Why is it important?
We want to add reliability in the CI, and after further investigation from
the functional teams, we are adding this as a counter-measure to avoid
more flaky failures at CI time.

We ideally want to revert this commit once the team knows about the long
timeouts.

## Checklist

## Related issues

- Relates elastic/beats#23593


mdelapenya:
Merging, as we need this change ASAP to increase the timeout on CI. Thanks!

PR #631, opened 2021-01-20
fix: let godog handle the creation of the junit test file (#629) backport for 7.x

mdelapenya:
 - fix: let godog handle the creation of the junit test file (#629)


mdelapenya:
Errors are not related to this PR, merging

PR #630, opened 2021-01-20
fix: do not override compose configurations when syncing from beats

mdelapenya:
## What does this PR do?
It moves the variable holding the references to a single service into the loop-for that is responsible of creating the service.

In the process, we added unit tests to the sanitize compose file feature, and for that we had to change method signature to support writing to a different file in the test side (adding testability).

## Why is it important?
We noticed that ceph was not receiving all the ports declared in its docker-compose file:

```yaml
version: '2.3'

services:
  ceph:
    image: docker.elastic.co/integrations-ci/beats-ceph:${CEPH_VERSION:-master-97985eb-nautilus-centos-7-x86_64}-2
    build:
      context: ./_meta
      dockerfile: Dockerfile.${CEPH_CODENAME:-nautilus}
      args:
        CEPH_VERSION: ${CEPH_VERSION:-master-97985eb-nautilus-centos-7-x86_64}
    ports:
      - 5000
      - 8003
      - 8080
  ceph-api:
    image: docker.elastic.co/integrations-ci/beats-ceph:master-6373c6a-jewel-centos-7-x86_64-1
    build:
      context: ./_meta
      dockerfile: Dockerfile.jewel
      args:
        CEPH_VERSION: master-6373c6a-jewel-centos-7-x86_64
    ports:
      - 5000
```

In the contrary it was overriden by the second service in the compose, ceph-api:

```yaml 
version: "2.3"
services:
  ceph:
    image: docker.elastic.co/integrations-ci/beats-ceph:${CEPH_VERSION:-master-97985eb-nautilus-centos-7-x86_64}-2
    ports:
    - 5000
  ceph-api:
    image: docker.elastic.co/integrations-ci/beats-ceph:master-6373c6a-jewel-centos-7-x86_64-1
    ports:
    - 5000
```

And this could lead to unexpected errors.

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have run the Unit tests for the CLI, and they are passing locally
- [ ] I have run the End-2-End tests for the suite I'm working on, and they are passing locally
- [ ] I have noticed new Go dependencies (run `make notice` in the proper directory)

## Author's Checklist

- [x] Verified ceph is having multiple ports

## How to test this PR locally

```shell 
$ make -C cli test
...
[...]

PR #646, opened 2021-01-25
fix: do not override compose configurations when syncing from beats (#630) backport for 7.x

mdelapenya:
 - fix: do not override compose configurations when syncing from beats (#630)


mdelapenya:
Same errors as in 7.x. Merging

PR #648, opened 2021-01-25
fix: do not override compose configurations when syncing from beats (#630) backport for 7.10.x

mdelapenya:
 - fix: do not override compose configurations when syncing from beats (#630)


mdelapenya:
Seems the test is flaky, as it passed for the default image and failed for the ubi8 one.

PR #651, opened 2021-01-25
chore: decouple installer packages

mdelapenya:
## What does this PR do?
This PR extracts all the code that is specific to the TAR/DEB/RPM installers to a separate file, defining the installation life cycle in a separate component.

For that reason we have moved code from the services.go file to a new file, where the responsabilities of each component are clearer. In this case, we are going to operate the OS packages:
- DEB: use `apt` to install the package
- RPM: use `yum` to install the package
- TAR: use `tar` to extract the package and then automatically install the agent

For the TAR installer, because it flattens the directory layout when extracting it, we need more data: artifact name, version, OS and arch.

## Why is it important?
This PR is not bringing new features but adding maintainability and keeping the code healthier, as we are decoupling basic installation process to a separate file.

## Checklist

### DIFF
--- a/e2e/_suites/fleet/fleet.go
+++ b/e2e/_suites/fleet/fleet.go
@@ -62,7 +62,7 @@ func (fts *FleetTestSuite) afterScenario() {
 
 		err := installer.UninstallFn()
 		if err != nil {
-			log.Warnf("Could not uninstall the agent after the scenario: %w", err)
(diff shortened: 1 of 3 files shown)


kuisathaverat (approved):

mdelapenya:
Same errors as in master. I'd merge once reviewed

cachedout (approved):
💯 Nice in-line documentation!

PR #662, opened 2021-01-26
chore: decouple installer packages (#651) backport for 7.10.x

mdelapenya:
 - chore: decouple installer packages (#651)


mdelapenya:
The only error seems flaky, as it happens on Centos but not on Debian. Merging

PR #673, opened 2021-01-28
fix: notify build result only once

mdelapenya:
## What does this PR do?
It removes the call to `notifyBuildResult` from the cleanUp hook, merging its parameters into the final invocation.

We are also fixing the flaky analyser name, using base branch instead of always using master.

After review, we detected the post success/failure blocks were confusing: we decided to simplify them so that there is only one entrypoint for creating the GH comment, although the slack notification is controlled exclusively by the input parameter (false by default).

## Why is it important?
We detected two documents on Jenkins-Stats, only differentiating in the artifacts field.

## Checklist

## Author's Checklist

- [ ] Flaky parameters: are they correct?

## Screenshots

Screenshot 2021-01-28 at 12 19 32

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -236,7 +236,6 @@ pipeline {
   post {
     cleanup {
       githubCheckNotify(currentBuild.currentResult == 'SUCCESS' ? 'SUCCESS' : 'FAILURE')
-      notifyBuildResult(analyzeFlakey: true, flakyReportIdx: "reporter-e2e-tests-end-2-end-tests-pipeline-master", prComment: true)
     }
     success {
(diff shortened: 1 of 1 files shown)


cachedout (approved):

mdelapenya:
Shouldn't the `notifyBuildResult` step post a GH comment with the build result?

v1v (changes requested):
This won't work for PRs, `whenTrue` and `whenFalse` conditions apply for the same

mdelapenya:
> This won't work for PRs, `whenTrue` and `whenFalse` conditions apply for the same

Thanks for the feedback! I refactored it to simplify the logic, explained in the main body

kuisathaverat (approved):

v1v (approved):

PR #674, opened 2021-01-29
fix: notify build result only once (#673) backport for 7.x

mdelapenya:
 - fix: notify build result only once (#673)


mdelapenya:
jenkins run the tests please

PR #677, opened 2021-01-29
fix: notify build result only once (#673) backport for 6.8.x

mdelapenya:
 - fix: notify build result only once (#673)


mdelapenya:
Errors happen because of the upstream kibana snapshot. Merging.

PR #705, opened 2021-02-08
Metricbeat goal

v1v:
## What does this PR do?

Enable metricbeat script to delegate all the logic to this repo then the consumer only need to call the script

## Why is it important?

Simplify the consumer calls

## Checklist

- ~~[ ] My code follows the style guidelines of this project~~
- ~~[ ] I have made corresponding changes to the documentation~~
- ~~[ ] I have made corresponding change to the default configuration files~~
- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~
- ~~[ ] I have run the Unit tests for the CLI, and they are passing locally~~
- ~~[ ] I have noticed new Go dependencies (run `make notice` in the proper directory)~~

## Author's Checklist

- [ ]

Run the script `.ci/scripts/metricbeat-test.sh`

## Related issues

Relates https://github.com/elastic/beats/pull/23854/

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -325,11 +325,6 @@ def generateFunctionalTestStep(Map args = [:]){
           if(isInstalled(tool: 'docker', flag: '--version')) {
             dockerLogin(secret: "${DOCKER_ELASTIC_SECRET}", registry: "${DOCKER_REGISTRY}")
           }
-          retry(3){
-            dir("${BASE_DIR}"){
-              sh script: """.ci/scripts/install-test-dependencies.sh "${suite}" """, label: "Install test dependencies for ${suite}:${tags}"
-            }
-          }
           filebeat(output: "docker_logs_${suite}_${tags}.log", workdir: "${env.WORKSPACE}"){
             dir("${BASE_DIR}"){
               sh script: """.ci/scripts/functional-test.sh "${suite}" "${tags}" "${stackVersion}" "${METRICBEAT_VERSION}" """, label: "Run functional tests for ${suite}:${tags}"
--- a/.ci/functionalTests.groovy
+++ b/.ci/functionalTests.groovy
@@ -107,11 +107,6 @@ pipeline {
       steps {
         deleteDir()
         unstash 'source'
-        retry(3){
-          dir("${BASE_DIR}"){
-            sh script: """.ci/scripts/install-test-dependencies.sh "${GO_VERSION}" """, label: "Install test dependencies for ${params.FEATURE}"
-          }
-        }
         dir(BASE_DIR){
           sh script: """.ci/scripts/functional-test.sh "${GO_VERSION}" "${params.FEATURE}" """, label: 'Run functional tests'
         }
--- a/.ci/scripts/functional-test.sh
+++ b/.ci/scripts/functional-test.sh
@@ -23,6 +23,9 @@ METRICBEAT_VERSION=${4:-'8.0.0-SNAPSHOT'}
 TARGET_OS=${GOOS:-linux}
 TARGET_ARCH=${GOARCH:-amd64}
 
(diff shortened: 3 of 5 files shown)


.ci/scripts/metricbeat-test.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
```
mdelapenya, line 1: Do you prefer a script (mainly used by the CI) or a Make goal in the e2e directory, so that developers can run them locally? There are already a few examples for fleet in that Makefile.

v1v: Will the retry in the CI something that could be also included in the make goal? See https://github.com/elastic/e2e-testing/pull/705/files#diff-12ed1626622bd381afb059882db321d0014a9e9e394c4ca6ff4731887d912076R20-R28

No strong opinion whether to use a sh or a make goal as long as it's just a simple call that the consumer of this project needs to care about

mdelapenya: Mmm that's right: I did not look at the retry command. Well, if it's simple as it is, then I'm OK with it.

BTW, this script is needed because we need to install the dependencies, and it's done in the Jenkinsfile, right? Wdyt about simplifying the Jenkinsfile to execute the same as in the consumer, with one script? It will simplify the Jenkinsfile (less calls) and the future script will be self-contained, following same strategy as the script in this PR.

v1v: > BTW, this script is needed because we need to install the dependencies, and it's done in the Jenkinsfile, right?

Yes

> Wdyt about simplifying the Jenkinsfile to execute the same as in the consumer, with one script? 

I was tempted to do so, but it might require further surgery, let me give a go

mdelapenya: Let me know if we can collaborate there

.ci/Jenkinsfile
```diff
@@ -325,11 +325,6 @@ def generateFunctionalTestStep(Map args = [:]){
           if(isInstalled(tool: 'docker', flag: '--version')) {
             dockerLogin(secret: "${DOCKER_ELASTIC_SECRET}", registry: "${DOCKER_REGISTRY}")
           }
-          retry(3){
```
v1v, line 328: `.ci/scripts/functional-test.sh` is the one in charge to call this script

.ci/functionalTests.groovy
```diff
@@ -107,11 +107,6 @@ pipeline {
       steps {
         deleteDir()
         unstash 'source'
-        retry(3){
```
v1v, line 110: `.ci/scripts/functional-test.sh` is the one in charge to call this script

.ci/scripts/functional-test.sh
```diff
@@ -23,6 +23,9 @@ METRICBEAT_VERSION=${4:-'8.0.0-SNAPSHOT'}
 TARGET_OS=${GOOS:-linux}
 TARGET_ARCH=${GOARCH:-amd64}
 
+## Install the required dependencies for the given SUITE
+.ci/scripts/install-test-dependencies.sh "${SUITE}"
+
```
v1v, line 28: It calls the script though the one in charge to retry is the `install-test-dependencies.sh` itself

.ci/scripts/metricbeat-test.sh
v1v: Done, I restructure the scripts so this is now a dummy one to call the functional-test.sh with some parameters, we could even potentially remove it, though I like the idea of using this contract

.ci/functionalTests.groovy
mdelapenya: I believe this file can be removed as it's not used anymore: let's do it in another PR

mdelapenya (approved):

PR #731, opened 2021-02-15
chore: use alias for stale version

mdelapenya:
## What does this PR do?
It replaces the hardcoded 7.10.2 version for the stale agent with an alias, in this case for this branch 7.10-SNAPSHOT. For that, we are resolving the alias to the real version with the `GetElasticArtifactVersion` method.

## Why is it important?
The 7.10.2 artifact dissapeared from Elastic's artifactory, so we are consuming the alias instead, and this is causing master and 7.11.x to fail because of that.

## Checklist

- [x] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have run the Unit tests for the CLI, and they are passing locally
- [x] I have run the End-2-End tests for the suite I'm working on, and they are passing locally
- [ ] I have noticed new Go dependencies (run `make notice` in the proper directory)

## Author's Checklist

- [x] 7.10-SNAPSHOT exists

## Related issues

- Closes #730

## Follow-ups
When backporting, we have to take care of the versions and aliases for each branch

PR #738, opened 2021-02-16
chore: use alias for stale version (#731) backport for 7.10.x

mdelapenya:
 - chore: use alias for stale version (#731)


mdelapenya:
@EricDavisX we are disabling here the upgrade tests for 7.10.x

PR #751, opened 2021-02-17
fix: support selecting the platform in docker-compose

mdelapenya:
## What does this PR do?
It adds the `platform` element to the docker-compose files for metricbeat and the elastic-agent, so that we will be able to target AMD64 or ARM64 (last one in the near future). This change forced us to upgrade compose files versions from 2.3 to 2.4, which is the version where `platform` attribute appeared.

Because of that, all files in the execution must use the same version, so we are updating all compose files in consequence. And that means that when synchronising the Integrations from Beats, we must enforce the 2.4 version too.

## Why is it important?
We want to make sure we consume the proper platform. After Beats creating ARM images, we need to make sure we consume the proper platform.

## Checklist

## Author's Checklist

- [ ]

## Related issues

- Relates #748

### DIFF
--- a/cli/cmd/sync.go
+++ b/cli/cmd/sync.go
@@ -221,6 +221,9 @@ func sanitizeComposeFile(composeFilePath string, targetFilePath string) error {
 		return err
 	}
 
+	// sets version to 2.4 for testing purpose
+	c.Version = "2.4"
+
 	for k, srv := range c.Services {
 		// we'll copy all fields but the excluded ones in this struct
 		output := make(map[string]service)
--- a/e2e/_suites/fleet/stand-alone.go
+++ b/e2e/_suites/fleet/stand-alone.go
@@ -95,6 +95,7 @@ func (sats *StandAloneTestSuite) aStandaloneAgentIsDeployed(image string) error 
 
 	profileEnv["elasticAgentContainerName"] = containerName
 	profileEnv["elasticAgentConfigFile"] = sats.AgentConfigFilePath
+	profileEnv["elasticAgentPlatform"] = "linux/amd64"
 	profileEnv["elasticAgentTag"] = agentVersion
 
 	err = serviceManager.AddServicesToCompose(FleetProfileName, []string{ElasticAgentServiceName}, profileEnv)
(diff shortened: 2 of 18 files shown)


e2e/_suites/fleet/stand-alone.go
```diff
@@ -95,6 +95,7 @@ func (sats *StandAloneTestSuite) aStandaloneAgentIsDeployed(image string) error
 
 	profileEnv["elasticAgentContainerName"] = containerName
 	profileEnv["elasticAgentConfigFile"] = sats.AgentConfigFilePath
+	profileEnv["elasticAgentPlatform"] = "linux/amd64"
```
mdelapenya, line 98: This value will be eventually dynamic for ARM/AMD

e2e/_suites/metricbeat/metricbeat_test.go
```diff
@@ -359,6 +359,7 @@ func (mts *MetricbeatTestSuite) runMetricbeatService() error {
 	}
 
 	env["metricbeatDockerNamespace"] = e2e.GetDockerNamespaceEnvVar()
+	env["metricbeatPlatform"] = "linux/amd64"
```
mdelapenya, line 362: This value will be eventually dynamic for ARM/AMD

cli/cmd/sync.go
```diff
@@ -221,6 +221,9 @@ func sanitizeComposeFile(composeFilePath string, targetFilePath string) error {
 		return err
 	}
 
+	// sets version to 2.4 for testing purpose
```
v1v, line 224: Not testing anymore?

```suggestion
```

v1v (approved):

kuisathaverat (approved):

cli/cmd/sync.go
mdelapenya: Well, I mean testing the execution of the tests. I added there to explain why we are setting that version there. Maybe it's not clear enough and it could be deleted

mdelapenya:
The failed test could be considered flaky. Merging

PR #761, opened 2021-02-17
chore: simplify synchronising the integrations

mdelapenya:
## What does this PR do?
It directly uses Go to run the CLI code instead of building the CLI binary for the underlying platform. Therefore we have moved the Make goal from the e2e dir to the CLI, updating the exec command: instead of invoking the binary we are directly running Go source code.

## Why is it important?
It simplifies the process, avoiding to package the binary.

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have run the Unit tests for the CLI, and they are passing locally
- [ ] I have run the End-2-End tests for the suite I'm working on, and they are passing locally
- [ ] I have noticed new Go dependencies (run `make notice` in the proper directory)

## How to test this PR locally

```shell
make -C cli sync-integrations
```

## Related issues

- Closes #740

PR #762, opened 2021-02-17
chore: go version cannot be overriden

### LINKED ISSUE #703: Do not allow go version to be overriden by consumer
GO_VERSION can be overriden, and if another project fetches the e2e tests to run them, then it's very possible that the go version value is modified.

As we are consuming beats binaries, we can be in different Go version.

mdelapenya:
## What does this PR do?
It makes the GO_VERSION variable to be not overridable.

## Why is it important?
This was a problem when a project downloads this one and runs the tests in an embedded manner, as it could be the case the project could use another Go version.

On the other hand, after #761, the build goal is not invoked anymore unless a release is performed, so the impact is very very low.

## Checklist

## Author's Checklist

- [ ]

## Related issues

- Closes #703

### DIFF
--- a/cli/Makefile
+++ b/cli/Makefile
@@ -4,7 +4,7 @@ ROOT_DIR:=$(CURDIR)
 TEST_TIMEOUT?=5m
 
 GO_IMAGE?='golang'
-GO_VERSION?='$(shell cat ../.go-version )'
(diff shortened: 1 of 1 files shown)


v1v (approved):

mdelapenya:
Failing tests potentially flaky. Merging

PR #784, opened 2021-02-18
Fleet goal

v1v:
## What does this PR do?

Enable metricbeat script to delegate all the logic to this repo then the consumer only need to call the script

## Why is it important?

Simplify the consumer calls

## Checklist

- ~~[ ] My code follows the style guidelines of this project~~
- [x] I have commented my code, particularly in hard-to-understand areas
- ~~[ ] I have made corresponding changes to the documentation~~
- ~~[ ] I have made corresponding change to the default configuration files~~
- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~
- ~~[ ] I have run the Unit tests for the CLI, and they are passing locally~~
- [x] I have run the End-2-End tests for the suite I'm working on, and they are passing locally
- ~~[ ] I have noticed new Go dependencies (run `make notice` in the proper directory)~~

## Author's Checklist

- [ ]

## How to test this PR locally

Run the script `.ci/scripts/fleet-test.sh`

## Related issues

Similar too https://github.com/elastic/e2e-testing/pull/705 and required by https://github.com/elastic/beats/pull/24112

PR #791, opened 2021-02-22
Fleet goal (#784) backport for 6.8.x

v1v:
 - Fleet goal (#784)


v1v:
unrelated errors since this script is not called in this pipeline

PR #809, opened 2021-02-24
chore: add logs when OS certs are not installed for the upgrade scenario

mdelapenya:
## What does this PR do?
This PR logs the error happened when the install certificates code fails. Adding this log, we noticed that the test scenario for the installation of the certificates was using an input parameter (`os`) that was not used, so we are removing it in consequence.

After that removal, we noticed that the TAR installer was not complete in the sense that the certs installation function was coupled to Debian only. This is working for the current implementation but could be misleading for future scenarios that would want to upgrade a Centos-based agent. For that reason, in this PR we are supporting the TAR installer package to know about the OS flavour that will be used to install the agent (centos or debian), and with that value, the install certs function will be able to call the proper package manager (`yum` or `apt`).

In this last refactor, we extracted that code to be reusable, as it was used in two different places.

## Why is it important?
We have seen a few times errors on CI about the certificates not being installed, and want to add visibility on those errors (before this change, the error was simply thrown without logs).

## Checklist

## Author's Checklist

- [ ]

```shell
SUITE="fleet" TAGS="fleet_mode_agent && reenroll" TIMEOUT_FACTOR=3 LOG_LEVEL=TRACE DEVELOPER_MODE=true ELASTIC_APM_ACTIVE=false make -C e2e functional-test
```

### DIFF
--- a/e2e/_suites/fleet/fleet.go
+++ b/e2e/_suites/fleet/fleet.go
@@ -187,7 +187,20 @@ func (fts *FleetTestSuite) installCerts(targetOS string) error {
 		return errors.New("no installer found")
 	}
 
-	return installer.InstallCertsFn()
+	err := installer.InstallCertsFn()
+	if err != nil {
+		log.WithFields(log.Fields{
+			"agentVersion":      agentVersion,
+			"agentStaleVersion": agentStaleVersion,
+			"error":             err,
+			"installer":         installer,
+			"targetOS":          targetOS,
+			"version":           fts.Version,
+		}).Error("Could not install the certificates")
+		return err
+	}
+
+	return nil
 }
 
 func (fts *FleetTestSuite) anAgentIsUpgraded(desiredVersion string) error {
--- a/e2e/_suites/fleet/fleet.go
+++ b/e2e/_suites/fleet/fleet.go
@@ -138,7 +138,7 @@ func (fts *FleetTestSuite) contributeSteps(s *godog.ScenarioContext) {
 	s.Step(`^an attempt to enroll a new agent fails$`, fts.anAttemptToEnrollANewAgentFails)
 	s.Step(`^the "([^"]*)" process is "([^"]*)" on the host$`, fts.processStateChangedOnTheHost)
 	s.Step(`^the file system Agent folder is empty$`, fts.theFileSystemAgentFolderIsEmpty)
-	s.Step(`^certs for "([^"]*)" are installed$`, fts.installCerts)
+	s.Step(`^certs are installed$`, fts.installCerts)
 
 	// endpoint steps
 	s.Step(`^the "([^"]*)" integration is "([^"]*)" in the policy$`, fts.theIntegrationIsOperatedInThePolicy)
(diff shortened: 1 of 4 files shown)


e2e/_suites/fleet/fleet.go
```diff
@@ -175,7 +175,7 @@ func (fts *FleetTestSuite) anStaleAgentIsDeployedToFleetWithInstaller(image, ver
 	return fts.anAgentIsDeployedToFleetWithInstaller(image, installerType)
 }
 
-func (fts *FleetTestSuite) installCerts(targetOS string) error {
```
mdelapenya, line 178: Input parameter not used internally. This targetOS is known by the installer that installed the agent

e2e/_suites/fleet/installers.go
```diff
@@ -80,16 +80,18 @@ func (i *DEBPackage) Install(containerName string, token string) error {
 
 // InstallCerts installs the certificates for a DEB package
 func (i *DEBPackage) InstallCerts() error {
-	if err := execCommandInService(i.profile, i.image, i.service, []string{"apt-get", "update"}, false); err != nil {
+	return installCertsForDebian(i.profile, i.image, i.service)
+}
+func installCertsForDebian(profile string, image string, service string) error {
```
mdelapenya, line 85: Extracting to a function so that the TAR installer is able to call this function when it's installed on a Debian machine

e2e/_suites/fleet/installers.go
```diff
@@ -223,19 +225,21 @@ func (i *RPMPackage) Install(containerName string, token string) error {
 
 // InstallCerts installs the certificates for a RPM package
 func (i *RPMPackage) InstallCerts() error {
-	if err := execCommandInService(i.profile, i.image, i.service, []string{"yum", "check-update"}, false); err != nil {
+	return installCertsForCentos(i.profile, i.image, i.service)
+}
+func installCertsForCentos(profile string, image string, service string) error {
```
mdelapenya, line 230: Extracting to a function so that the TAR installer is able to call this function when it's installed on a Centos machine

e2e/_suites/fleet/services.go
```diff
@@ -421,8 +421,8 @@ func newDockerInstaller(ubi8 bool, version string) (ElasticAgentInstaller, error
 
 // newTarInstaller returns an instance of the Debian installer for a specific version
 func newTarInstaller(image string, tag string, version string) (ElasticAgentInstaller, error) {
-	image = image + "-systemd" // we want to consume systemd boxes
-	service := image
+	dockerImage := image + "-systemd" // we want to consume systemd boxes
```
mdelapenya, line 424: This variable will be the one holding the docker image name, that's why we replaced all occurrences with `dockerImage`. On the other hand, the `image` variable will hold the _flavour_: centos or debian.

e2e/_suites/fleet/installers.go
```diff
@@ -287,16 +292,12 @@ func (i *TARPackage) Install(containerName string, token string) error {
...
-	if err := execCommandInService(i.profile, i.image, i.service, []string{"apt", "install", "ca-certificates", "-y"}, false); err != nil {
-		return err
-	}
-	if err := execCommandInService(i.profile, i.image, i.service, []string{"update-ca-certificates", "-f"}, false); err != nil {
-		return err
+	if i.OSFlavour == "centos" {
+		return installCertsForCentos(i.profile, i.image, i.service)
+	} else if i.OSFlavour == "debian" {
+		return installCertsForDebian(i.profile, i.image, i.service)
 	}
```
michalpristas, line 301: can we have a debug message that installation was skipped because of unknown OS

michalpristas (approved):
LGTM

e2e/_suites/fleet/installers.go
mdelapenya: Done in 38d068c

cachedout (approved):

PR #812, opened 2021-02-24
chore: add logs when OS certs are not installed for the upgrade scenario (#809) backport for 6.8.x

mdelapenya:
 - chore: add logs when OS certs are not installed for the upgrade scenario (#809)


mdelapenya:
Failed test could be potentially flaky. Merging

PR #816, opened 2021-02-24
chore: download installers when they are used (#810) backport for 7.11.x

mdelapenya:
Backports the following commits to 7.11.x:
 - chore: download installers when they are used (#810)


mdelapenya:
Failed test could be potentially flaky. Merging

PR #856, opened 2021-03-03
WIP: gather metrics

v1v:


mdelapenya:
As discussed with @v1v, it would make more sense having a dedicated pipeline under Beats CI to gather these metrics than gathering them in a live PR, which will have conflicts on each change in the Jenkinsfile. Changes that come every day with the automatic bump of the runtime dependencies (the stack).

We suggested moving this very same pipeline to a repo controlled by JJBB under Beats CI.

Until that, we can keep this PR open

PR #859, opened 2021-03-08
move code to where it is used

jalvz:
## What does this PR do?

This moves helper functions and variables closer to where they are generally used.

## Why is it important?

It follows standard Go practices and aids IDE auto-completion. Normally, you would not call a function defined in a file `foo_test.go` from `foo.go`; usually is the other way around. Godog compiles fine by linking the right files, but it confuses the hell out of IDEs and me too ;) 

## Checklist

Run tests


mdelapenya:
I've been thinking about this for a long time and never gave it a try, but you make a good point about file names. How would you see removing the `ingest_manager_test.go` to `ingest_manager.go`? The old name was there for historical not-valid-anymore reasons, so I'd see with good eyes to rename the file. In fact, newer code is using the approach you brightly points: using `_test.go` for unit tests.

I'd prefer renaming the file because the ingest_manager file represents the main entrypoint for the test suite: it defines main variables, and is the responsible of creating the downstream test suite structs (fleet and standalone).

Wdyt?

jalvz:
@mdelapenya that sounds good to me. However I tried and it didn't work, it seems that godog will only search for the entry point in a `_test.go` file. So following your suggestion as much as I could, I moved everything to `ingest_manager.go` and kept the bare minimum in the test file. 
Additionally, I added a `TestMain` function to integrate `godog` with `go test`, meaning that bdd tests can be run with `go test` as well. We don't need to change how it is run, but just as an alternative - eg. this should make possible to run and debug tests from the IDE even if the IDE doesn't support godog.   

I know there is a merge conflict, but lmk what do you think generally.

e2e/_suites/fleet/ingest_manager_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+func InitializeIngestManagerTestSuite(ctx *godog.TestSuiteContext) {
+	initializeIngestManagerTestSuite(ctx)
+}
+
+func InitializeIngestManagerTestScenario(ctx *godog.ScenarioContext) {
+	initializeIngestManagerTestScenario(ctx)
+}
+
+func TestMain(m *testing.M) {
```
mdelapenya, line 28: Although I like this approach, I think it's mixing the unit tests from the godog tests.

There is a unit test layer (`make -C e2e unit-test`) that executes only unit tests, not BDD/godog tests, and it seems this change is adding the godog tests to the execution of the unit tests, which uses `[gotestsum](https://github.com/gotestyourself/gotestsum)` as test runner (uses go test plus pigments)

jalvz:
Superseded by #865

Issue #884, opened 2021-03-15
Fix the download packages with read only access


v1v:
## What

Downloaded artifacts are not deleted

```
cannot remove 'workspace/e2e-tests_e2e-testing-mbp_PR-707/pkg/mod/howett.net/plist@v0.0.0-20181124034731-591f970eefbb/plist.go': Permission denied

```

### Why

Workspace are left in some static workers such as the ARM ones and that could cause disk space issues.

### Actions

Use something similar to https://github.com/elastic/beats/pull/24316/files#diff-e6ffa5dc854b843b3ee3c3c28f8eae2f436c2df2b1ca299cca1fa5982e390cf8R624

Reference https://github.com/elastic/beats/issues/24284#issuecomment-789746894

PR #707, opened 2021-02-08
feat: add support for running the tests in ARM

mdelapenya:
## What does this PR do?
It enables the ARM platform for metricbeat and fleet test suites, grabbing ARM workers, and for that reason we are refactoring the way we define the architecture in the e2e code. Instead of hardcoding the `amd64` architecture, we are going to read it from the environment. Golang provides a `GOARCH` environment variable that is automatically set by Go's runtime. With that variable we are going to calculate the artifact name based on the underlying architecture:

- for CI workers running `amd64`, the code will generate an artifact for `amd64`.
- for CI workers running `arm64`, the code will generate an artifact for `arm64`.

Besides that, we had to rewrite the Jenkins pipeline to properly assign a worker in the right platform, to retrieve a non-immutable worker for `arm`.

Finally, we are adding the support to exclude scenarios per platform, leveraging the `.e2e-testing.yaml` file, being possiblee for a scenario to override its parent suite's platforms. To add awareness at this level, we are always removing the scenarios tagged with `@skip:PLATFORM` (amd64, arm64), including those tags in the feature file when needed, so that developers now what is going to be run directly in the source code, not in the CI descriptor.

## Why is it important?
Add support to ARM platform

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have run the Unit tests for the CLI, and they are passing locally
- [x] I have run the Unit tests for the E2E, and they are passing locally
- [x] I have run the End-2-End tests for the suite I'm working on, and they are passing locally
- [ ] I have noticed new Go dependencies (run `make notice` in the proper directory)

## Author's Checklist

- [x] Beats merged the compilation in `arm`.
- [ ] Official Ubi8 Docker images for `arm` are not in place. See https://github.com/elastic/beats/issues/24186. This will cause that any job for the E2E consuming an official release will fail the ubi8 tests for `arm`. The jobs triggered by Beats project (PRs and branches) won't be affected, as they consume the CI artifacts.
- [x] Beats-CI Ubi8 Docker images for `arm` are generated and pushed in our GCP bucket

## Related issues

- Closes #586
- Closes #864 
- Needs https://github.com/elastic/beats/issues/24186

## Screenshots

[...]

PR #885, opened 2021-03-15
[CI] tear down the workspace

v1v:
## What does this PR do?

Delete workspace

## Why is it important?

If we use static/permanent workers then we might suffer disk space issues

## Checklist

## Author's Checklist

- [ ]

## Related issues

Closes https://github.com/elastic/e2e-testing/issues/884

https://github.com/elastic/e2e-testing/pull/707 can benefit from this :)

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -326,10 +326,10 @@ def generateFunctionalTestStep(Map args = [:]){
 
   return {
     node("${workerLabels}") {
-      try {
-        deleteDir()
-        unstash 'source'
-        withGoEnv(version: "${GO_VERSION}"){
+      deleteDir()
+      unstash 'source'
+      withGoEnv(version: "${GO_VERSION}"){
+        try {
           if(isInstalled(tool: 'docker', flag: '--version')) {
             dockerLogin(secret: "${DOCKER_ELASTIC_SECRET}", registry: "${DOCKER_REGISTRY}")
           }
@@ -340,13 +340,11 @@ def generateFunctionalTestStep(Map args = [:]){
               }
             }
           }
+        } finally {
+          junit(allowEmptyResults: true, keepLongStdio: true, testResults: "${BASE_DIR}/outputs/TEST-*.xml")
+          archiveArtifacts allowEmptyArchive: true, artifacts: "${BASE_DIR}/outputs/TEST-*.xml"
+          tearDown(labels: workerLabels)
         }
-      } catch(e) {
-        error(e.toString())
-      } finally {
-        junit(allowEmptyResults: true, keepLongStdio: true, testResults: "${BASE_DIR}/outputs/TEST-*.xml")
-        archiveArtifacts allowEmptyArchive: true, artifacts: "${BASE_DIR}/outputs/TEST-*.xml"
-        tearDown(labels: workerLabels)
       }
     }
   }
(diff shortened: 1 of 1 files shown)


.ci/Jenkinsfile
```diff
@@ -344,7 +345,24 @@ def generateFunctionalTestStep(Map args = [:]){
...
+
+/**
+* Tear down the setup for the static workers.
+*/
+def tearDown(Map args = [:]){
+  catchError(buildResult: 'SUCCESS', stageResult: 'SUCCESS') {
+    dir("${BASE_DIR}"){
+      sh(label: 'Remove the entire module cache', script: 'go clean -modcache', returnStatus: true)
+    }
+    if (isStaticWorker(labels: args.workerLabels)) {
```
v1v, line 363:
```suggestion
    if (isStaticWorker(labels: args.labels)) {
```

.ci/Jenkinsfile
```diff
@@ -344,7 +345,24 @@ def generateFunctionalTestStep(Map args = [:]){
...
+        tearDown(labels: workerLabels)
+      }
+    }
+  }
+}
+
+/**
+* Tear down the setup for the static workers.
+*/
+def tearDown(Map args = [:]){
```
mdelapenya, line 358: Wdyt about having a specific step in the library for cleaning up Go dependencies?

mdelapenya (approved):

.ci/Jenkinsfile
v1v: IMO the build system should be the one ensuring the workspace is in good shape. Then the CI should only caring to delete the workspace if required.

Though, it seems `go` has got some opinionated approach and the build system does not provide those tools, based on that, we could create a specific step in the library to help with.

There are some questions regarding:
- the new step will always run within the `Go` environmental context?
- if running as a post declarative step, will it work?

Do you think we just need to raise an issue regarding this improvement?

mdelapenya: I was/am concerned about having the very same method in different pipelines (Beats and here) and probably in many of the other Go projects that we maintain. 

I agree that the build system should keep the workspace and dependencies in a good shape, but as you know there is nothing as maven/gradle to do so.

Regarding your questions:
- I think the Go environment should be the same, just in case the `clean` method could change between different Go versions
- I've been reading about `go clean`. Do you think it could be a best practice to invoke it before tests run? The same we do `mvn clean test` in the Maven world. If so then we should not think about the post step issue, as the environment would always be right. Even better, we could enforce the clean in the set Go env step 🤔

cachedout (approved):

mdelapenya:
BTW, the step in the pipeline shows this message:

>[2021-03-15T10:28:09.830Z] /var/lib/jenkins/workspace/e2e-tests_e2e-testing-mbp_PR-885/src/github.com/elastic/e2e-testing@tmp/durable-95501832/script.sh: 1: /var/lib/jenkins/workspace/e2e-tests_e2e-testing-mbp_PR-885/src/github.com/elastic/e2e-testing@tmp/durable-95501832/script.sh: go: not found

Is it related?

v1v:
It's now much better with https://github.com/elastic/e2e-testing/pull/885/commits/4630b71dfe412fd410ab60e18c3ac0ab65e1f28d

image

mdelapenya (approved):
LGTM! The failed test seems one of the flaky ones

.ci/Jenkinsfile
```diff
@@ -339,12 +340,27 @@ def generateFunctionalTestStep(Map args = [:]){
               }
             }
           }
-        }
-      } catch(e) {
-        error(e.toString())
-      } finally {
-        junit(allowEmptyResults: true, keepLongStdio: true, testResults: "${BASE_DIR}/outputs/TEST-*.xml")
+        } finally {
```
v1v, line 343: We only need the finally, then error is redundant

Issue #438, opened 2020-11-12
[Fleet] Infra work / new test creation, for coming Fleet Server component 


EricDavisX:
will add more notes as it is clearer... main tracking issue: integrations-dev/issues/389

What we know now:
- Any 'Agent' can be set up as a Fleet Server, so our Docker usage is acceptable, as is of course the standard Centos / Debian / UBI8 / other images
- Stand-alone Agent scenarios won't be modified to use Fleet Server, as they will communicate w ES direct still
... An Agent running as Fleet Server will or will not be still a regular 'Agent' as well?
- We could validate that Stand-Alone Agent cannot be setup as a Fleet Server
- I think we should build the expectation of using Fleet Server into the existing install scripts... need to think on this more.

EricDavisX:
Discussion notes from meeting on setting up Fleet Server:

general steps:

step 1) Agent needs to be running as a daemon

step 2) install (or enroll) command, pass in option --fleet_server with url + uid/pwd
 - no api key usage 
 - which writes into fleet.yml

then the running daemon is told to reexec (on linux and mac it injects into the same pid) - and it reads the yml and brings up the fleet-server subprocess

if Kibana is not running, Fleet-server will wait, polling ES for the policy it should be on and confirming the index exists.  wait for 2 mins, then times out.

Fleet server will then allow enrollment to begin

waits for 'degraded' state, then it enrolls (locally). rewrites fleet.yml - then the bootstrap flag is not included (set to false). and reexecs again (2nd time).

must include the Fleet Server integration in Kibana policy, get that policy id from the enroll command

on Kibana side, just a feature flag setup.  not merged yet, but coming soon. :)
 - will need to visit the /setup endpoint, but what about actually clicking button in Kibana?

e2e-test steps / questions:
 - process name is fleet-server running as a child of agent
 - more tests:  passing in bad PWD
 - tests:  using user with missing permissions

EricDavisX:
FYI - the basic process above should be discussed in: https://github.com/elastic/fleet-server/blob/master/README.md
and it basically amounts to just calling:
sudo ./elastic-agent enroll --enrollment-token {enrollment-token} --fleet-server http://[elastic-user]:[elastic-user-creds]@[fleet-server-url]:9200 --fleet-server-policy {fleet-server-policy-id}

the {fleet-server-policy-id} will be generated by default when Kibana starts up with Fleet in 7.13 and beyond. so it's name / id can be queried like any other policy id.

[...]

PR #900, opened 2021-03-16
feat: add first scenario for Fleet Server

mdelapenya:
## What does this PR do?
This PR adds a feature file for Fleet Server, adding one scenario checking that when an agent is deployed with Fleet Server mode, then Fleet Server is enabled.

To allow that, we had to do some refactors, explained below:

- We enabled Fleet Server in Kibana config: `xpack.fleet.agents.fleetServerEnabled: true`
- We added a stronger manner to capture the Default Policy. In the past, we simply got the first child item in the policies list, as we were sure there was only one policy. Now with Fleet Server, a new policy is created, and the way to distinguish both is filtering by the `is_default` and `is_default_fleet_server` attributes in the policy.
- We added a struct to handle Fleet Config, for both modes: Fleet mode and Fleet server mode. This struct will hold the configuration to handle the connection to Elasticsearch and the policy used to enroll the fleet server agent. A pointer to this config will be returned each time an agent is deployed to fleet, and the internals of this config will be different for Fleet Server, as it will request the default Fleet Server policy from Kibana, and will store its ID in the struct. This field will be used to identify what flags will be used when enrolling/installing the agent. For that reason there is a `flags()` method returning an array of flags to use in the install/enroll commands. Finally, this structure will be passed to the different command to be reused, specially the `EnrollmentToken`.
- We abstracted the `deployAgentWithInstaller` method, so that it's reusable for existing scenarios, and the new one for fleet server.

It's important to notice that the final step in the new scenario is not finished, it's returning a `Pending` error. This is because, at the moment of sending this PR, it's not clear to me how to checck that Fleet Server has been enabled/deployed successfully.

## Why is it important?
It will bring the first scenario for fleet server, also improving code base health after refactors.

## Checklist

## Author's Checklist

- [ ] @blakerouse could you help out with the assertions to perform here in the scenario?

```shell
SUITE="fleet" TAGS="fleet_server && centos" TIMEOUT_FACTOR=3 LOG_LEVEL=TRACE DEVELOPER_MODE=true make -C e2e functional-test
```

## Related issues

- Relates #438


e2e/_suites/fleet/features/fleet_server.feature
```diff
@@ -0,0 +1,19 @@
+@fleet_server
+Feature: Fleet Server
+  Scenarios for Fleet Server, where an Elasticseach and a Kibana instances are already provisioned,
+  so that the Agent is able to communicate with them
+
+@start-fleet-server
+Scenario Outline: Deploying the <os> fleet-server agent
+  When a "<os>" agent is deployed to Fleet with "tar" installer in fleet-server mode
+  Then Fleet server is enabled
```
mdelapenya, line 9: @EricDavisX  @blakerouse I wrote this `Then` clause, but I'd like to know if there is an assertion that should be done here. Something like: `the elastic-agent process is started`, or `the Fleet app in Kibana shows "FooBar" in the Fleet page`, or "elasticsearch contains THIS doc in THAT index".

Could you help me here in writing the right expected behaviour?

e2e/_suites/fleet/fleet_server.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func (fts *FleetTestSuite) anAgentIsDeployedToFleetWithInstallerInFleetMode(image string, installerType string) error {
+	fts.ElasticAgentStopped = true
+	return fts.anAgentIsDeployedToFleetWithInstallerAndFleetServer(image, installerType, true)
+}
+
+func (fts *FleetTestSuite) fleetServerIsEnabled() error {
+	log.Debug("Fleet server is enabled")
+	return godog.ErrPending
```
mdelapenya, line 73: Pending until we find the right assertion

e2e/_suites/fleet/installers.go
```diff
@@ -374,8 +375,8 @@ func (i *TARPackage) Preinstall() error {
 
 	// simplify layout
 	cmds := [][]string{
-		[]string{"rm", "-fr", "/elastic-agent"},
-		[]string{"mv", fmt.Sprintf("/%s-%s-%s-%s", i.artifact, i.version, i.OS, i.arch), "/elastic-agent"},
+		{"rm", "-fr", "/elastic-agent"},
```
mdelapenya, line 378: Remove some leftovers: the type is automatically inferred by the Go compiler

EricDavisX:
This is wonderful Manu.  Thank you.  I will review expectations with the team if we don't hear feedback.

e2e/_suites/fleet/features/fleet_server.feature
blakerouse: I am not too familiar with the structure of these files so I might not be fully understanding the context.

I believe your asking if its possible to known if the Fleet Server is running correct. The simplest way is to check that the Agent is reported Healthy in Kibana. That might seem to simple but the **only** way for the Agent running a Fleet Server to show in Kibana as healthy is if it can communicate to its local Fleet Server and that Fleet Server can write to elasticsearch.

mdelapenya: Never mind about the code structure yet, I'm still interested in the behavior of the product without considering internal details/implementations.

With that in mind:

- what ES query should we write to verify that?
- is it enough to check the healthy status for the promoted-to-server agent?

cachedout (approved):

cachedout (approved):

e2e/_suites/fleet/features/fleet_server.feature
```diff
@@ -0,0 +1,19 @@
+@fleet_server
+Feature: Fleet Server
+  Scenarios for Fleet Server, where an Elasticseach and a Kibana instances are already provisioned,
+  so that the Agent is able to communicate with them
+
+@start-fleet-server
+Scenario Outline: Deploying the <os> fleet-server agent enables Fleet Server
```
jalvz, line 7: This is surely because of my lack of understanding of how fleet-server works, but I don't find this phrasing very clear: there is no such a thing as "fleet-server agent", right? There is an Elastic Agent that can start a Fleet Server process: shouldn't this be in the scenario definition?

Since "Fleet" is such a ubiquitous term, it might be worth to be more explicit every time is it used: Is it Fleet API, Fleet UI, etc.
Related: since this spec is meant to be semi-formal, and we put a lot of effort in standarisation elsewhere (linting, etc), It would be good to define components canonically. Eg., always "Elastic Agent", and not a combination of "Elastic Agent", "Agent", "the agent", etc.

e2e/_suites/fleet/fleet.go
```diff
@@ -1259,24 +1268,38 @@ func deployAgentToFleet(installer ElasticAgentInstaller, containerName string, t
...
 	}
 
 	err = installer.PreInstallFn()
 	if err != nil {
-		return err
+		return nil, err
 	}
 
-	err = installer.InstallFn(containerName, token)
+	var fleetConfig *FleetConfig
```
jalvz, line 1279: Didn't actually tried, but I think it would be a bit simpler to have just 1 kind of config and 1 way of creating it, eg: `func NewAgentConfig(token, fleetServerMode bool) (*AgentConfig, error)` 

Credentials, url and port can also be hardcoded in the only place they are used.

jalvz (commented):
This is looking good! 
What is needed to land it?

e2e/_suites/fleet/features/fleet_server.feature
mdelapenya: **I strongly agree with this. 💯**

What about sending the refactors in a follow-up? Let's create an issue to standardise the names across Fleet test suite. Maybe @EricDavisX can help with the wording.

mdelapenya: Related to the name, yes. It could be rephrased. Let me add a suggestion.

mdelapenya, line 7:
```suggestion
Scenario Outline: Deploying an <os> Elastic Agent that starts Fleet Server
```

e2e/_suites/fleet/features/fleet_server.feature
```diff
@@ -0,0 +1,19 @@
...
+  When a "<os>" agent is deployed to Fleet with "tar" installer in fleet-server mode
```
mdelapenya, line 8: @jalvz do you think this step must be rephrased too? IIRC, I decided to create a new one instead of:

```gherkin
When a "<os>" agent is deployed to Fleet with "tar" installer
   And the agent start Fleet Server
```
To avoid invoking the `install` method twice.

e2e/_suites/fleet/fleet.go
mdelapenya: 👍

mdelapenya: Implemented in https://github.com/elastic/e2e-testing/pull/900/commits/f5a9f46f8153aaf9595a4821f1a50951c4452a39

mdelapenya:
> This is looking good!
> What is needed to land it?

There isn error in the `agent in the "online" status`, as it never reaches that state. Maybe we need to perform another assertion

blakerouse (approved):

e2e/_suites/fleet/features/fleet_server.feature
jalvz: sgtm!

jalvz:
> There is an error in the agent in the "online" status, as it never reaches that state

I spent quite some time today trying to understand why it doesn't work. I stumbled upon the installers code and got a bit lost. 

If I understand correctly (please correct me if not), some tests are validating different installation methods of Elastic Agent: why are those here? It seems to me that packaging should be unit tested and not e2e, and I expect that to be fairly well covered in libbeat: is that not the case?

Secondly: even if we maintain the install tests, why the other tests (enrolling, ingestion etc.) go trough custom installers? Since it all runs on Docker, shouldn't we just run the stock images as they are? 

I'm probably missing something about multiple platform support, but well, that's what the questions are for :) 

I hacked a `when` function to start Agent in Fleet Server mode and it worked. It's far from production ready and probably more complicated that what it should be, but for the sake of it here it is. Keep the same feature file in this PR, just hook the `when` step to the following:

```go
func (fts *FleetTestSuite) StartElasticAgentWithFleetServerMode(_ string) error {
	docker, err := client.NewClientWithOpts(client.FromEnv)
	if err != nil {
		return errors.Wrap(err, "Error creating Docker client")
	}
	ctx := context.Background()
	events, err := docker.ImagePull(ctx, "docker.elastic.co/beats/elastic-agent:8.0.0-SNAPSHOT", types.ImagePullOptions{})
	if err != nil {
		return errors.Wrap(err, "Error pulling Elastic Agent image")
	}

	d := json.NewDecoder(events)

	type Event struct {
		Status         string `json:"status"`
		Error          string `json:"error"`
		Progress       string `json:"progress"`
		ProgressDetail struct {
			Current int `json:"current"`
			Total   int `json:"total"`
		} `json:"progressDetail"`
	}

	var event *Event
	for {
		if err := d.Decode(&event); err != nil {
			if err == io.EOF {
				break
			} else {
				return errors.Wrap(err, "Error decoding image download status")
			}
		}
	}

	if event != nil {
		if strings.Contains(event.Status, "Downloaded") {
			fmt.Println("Downloaded Elastic Agent image")
		}

		if strings.Contains(event.Status, "Image is up to date") {
			fmt.Println("Elastic Agent image up to date")
		}
	}

	images, err := docker.ImageList(context.Background(), types.ImageListOptions{})
	if err != nil {
		return errors.Wrap(err, "Error retrieving Docker images")
	}
	var imageID string
	var created int64
	for _, im := range images {
		for _, tag := range im.RepoTags {
			if strings.Contains(tag, "docker.elastic.co/beats/elastic-agent:8.0.0-SNAPSHOT") {
				if im.Created > created {
					imageID = im.ID
					created = im.Created
				}
			}
		}
	}
	fmt.Println("Found Elastic Agent image: " + imageID)

	agentImageDetails, _, err := docker.ImageInspectWithRaw(context.Background(), imageID)
	if err != nil {
		return errors.Wrap(err, "Error inspecting Docker image")
	}

	cfg := agentImageDetails.Config
	cfg.Image = "docker.elastic.co/beats/elastic-agent:8.0.0-SNAPSHOT"
	cfg.Env = append(cfg.Env,
		"FLEET_SERVER_ENABLE=1", "FLEET_SERVER_INSECURE_HTTP=1", "KIBANA_FLEET_SETUP=1", "FLEET_SERVER_HOST=0.0.0.0",
		"ELASTICSEARCH_PASSWORD=changeme", "ELASTICSEARCH_USERNAME=elastic", "ELASTICSEARCH_HOST=http://elasticsearch:9200",
		"KIBANA_PASSWORD=changeme", "KIBANA_USERNAME=elastic", "KIBANA_HOST=http://kibana:5601")
	cfg.ExposedPorts = map[nat.Port]struct{}{"8220": {}}
	container, err := docker.ContainerCreate(context.Background(),
		cfg,
		&container.HostConfig{
			PortBindings: map[nat.Port][]nat.PortBinding{
				"8220": {{HostPort: "8220"}},
			},
		},
		&network.NetworkingConfig{},
		"fleet_server",
	)
	if err != nil {
		return errors.Wrap(err, "Error creating container")
	}
	fmt.Println("Container created: " + container.ID[:12])

	networks, err := docker.NetworkList(context.Background(), types.NetworkListOptions{})
	if err != nil {
		return errors.Wrap(err, "Error listing Docker networks")
	}
	var networkID string
	for _, n := range networks {
		if n.Name == "fleet_default" {
			networkID = n.ID
			break
		}
	}
	err = docker.NetworkConnect(context.Background(), networkID, container.ID, &network.EndpointSettings{})
	if err != nil {
		return errors.Wrap(err, "Error connecting to the fleet_default network")
	}

	// Need to stop and remove manually
	err = docker.ContainerStart(context.Background(), container.ID, types.ContainerStartOptions{})
	if err != nil {
		return errors.Wrap(err, "Error starting container")
	}
	fmt.Println("Container started")


	hostname, err := steps.GetContainerHostname("fleet_server")
	if err != nil {
		return err
	}
	fts.Hostname = hostname

	return nil
}
```

mdelapenya:
Let me try to address your questions :)

> I spent quite some time today trying to understand why it doesn't work. I stumbled upon the installers code and got a bit lost.
> 
> If I understand correctly (please correct me if not), some tests are validating different installation methods of Elastic Agent: why are those here? It seems to me that packaging should be unit tested and not e2e, and I expect that to be fairly well covered in libbeat: is that not the case?

My thoughts on the installation tests been moved to the e2e level of the testing pyramid is related to the need a live host to perform the installation. That could be perfectly done at the libbeat codebase (which I do not have experience with it), but it would need the build system to have the ability to start/stop "hosts" (VMs, containers, etc) to perform the install. I'm aware of the existence of the [Beats Tester project](https://github.com/elastic/beats-tester), which uses Ansible to actually "install" a Beat into a VM. I could see with good eyes both projects to converge into one, although I see differences in how both systems "install" the Beat:

- Beats Tester: it uses package managers to install the Beat from a local binary, and makes sure the installation directory layout is created (https://github.com/elastic/beats-tester/blob/master/roles/test-install/tasks/main.yml). It uses a declarative manner for this verification, using Ansible roles for it. This is great, because it should be idempotent installing the role again and again. BTW, it does not cover the Elastic Agent (https://github.com/elastic/beats-tester/blob/master/packages.yml)
- E2E Testing: t uses package managers to install the Beat from a local binary, and exercises certain elastic-agent's commands, such as "install" and "enroll". So, from my point of view, this approach is more functional as it is verifying features and creating assertions on the expected outcomes: there are backend processes started in the system, an API call was invoked in Kibana, the agent is enrolled on Fleet, there are documents in the index, etc, etc.

Not sure if the latter should be done at the unit level in libbeat.
 
> Secondly: even if we maintain the install tests, why the other tests (enrolling, ingestion etc.) go trough custom installers? Since it all runs on Docker, shouldn't we just run the stock images as they are?
> 
> I'm probably missing something about multiple platform support, but well, that's what the questions are for :)

There are two modes, and please correct me @EricDavisX if I'm wrong: **stand-alone** and **fleet**. While the stand-alone mode uses the stock images (`docker.elastic.co/beats/elastic-agent:8.0.0-SNAPSHOT`), which is automatically auto-enrolled in Fleet when started, the fleet mode works differently in the way that represents a valid use case: _As a sysadmin, I have my centos machine and want to install the agent and enroll it_. For that reason this fleet mode uses a plain Centos container and perform the installation of the artifacts (TAR, RPM, DEB) first.

mdelapenya:
>I hacked a when function to start Agent in Fleet Server mode and it worked. It's far from production ready and probably more complicated that what it should be, but for the sake of it here it is. Keep the same feature file in this PR, just hook the when step to the following.

About the "it worked" part: I used your code and indeed it enrolls the agent. I wonder why the code we use to verify if it's enrolled fails.

We hit this API endpoint: `http://localhost:5601/api/fleet/agents?page=1&perPage=20&showInactive=false` to get the enrolled agents, and the agent-under-test is never shown in the response, so we cannot get the `agentID`. This `agentID` is important because it's the field used to get the agent "online" status.

I wonder why using the following fleet-server flags during the `install` command causes the agent not to make it to the list:

```go
func (cfg FleetConfig) flags() []string {
	baseFlags := []string{"--force", "--insecure", "--enrollment-token=" + cfg.EnrollmentToken}

	if cfg.ServerPolicyID != "" {
		return append(baseFlags, "--fleet-server", fmt.Sprintf("http://%s@%s:%d", cfg.ElasticsearchCredentials, cfg.ElasticsearchURI, cfg.ElasticsearchPort), "--fleet-server-policy", cfg.ServerPolicyID)
	}

	return append(baseFlags, "--kibana-url", "http://kibana:5601")
}
```

jalvz:
Thanks for explaining.

> from my point of view, this approach is more functional as it is verifying features and creating assertions on the expected outcomes

Ok, my only comment then is that maybe it puts too much focus on the Elastic Agent and too little on the Integrations. If that is intentional, it is fine. Going forward I think it would be good keeping what there is, but shift focus towards the components that users interact with (ie. packages, data streams, dashboards, etc). 

> the fleet mode works differently in the way that represents a valid use case: As a sysadmin, I have my centos machine and want to install the agent and enroll it

Got it. I didn't look at the standalone code (yet), but personally I would lean towards using stock images for pretty much everything. The scenario you describe is still important and needs coverage, but I would exercise it based on official documentation (so perform the steps that external users are told to do). In other words, wait until Fleet Server is stable and closer to the finish line. Just my 2c, ofc.

(I don't know either why that code doesn't work, but even if it worked I wouldn't be surprised if it breaks later.)

jalvz:
Btw, I think that the code used to verify if it is healthy does work. What doesn't work well the way in which Elastic Agent is ran. `docker logs` doesn't show anything (as kind of expected - but it would be nice if it did), `top -c` only shows `/opt/Elastic/Agent/elastic-agent` (without any arguments), and logs show no attempt to try to enroll at all.

EricDavisX:
I appreciate the communication to push this forward!  I agree we should test as close to what user's will do, and intend long term to use the actual Doc'ed steps.  For now it may help to stand-up some support to test it with any 'hacky' means we can, to confirm progress is stable against the long-term goal (of having lots of tests).

We'll want to have tests for the stand-alone Docker container, and for the .tar.gz and .rpm / .deb Agents, and soon enough for .zip Windows agents too (it is coming soon in April I think).

mdelapenya:
> Ok, my only comment then is that maybe it puts too much focus on the Elastic Agent and too little on the Integrations. If that is intentional, it is fine. Going forward I think it would be good keeping what there is, but shift focus towards the components that users interact with (ie. packages, data streams, dashboards, etc).
 
Indeed. We started with the install scenarios because at that time that was what we had 🤷‍♀️ and now that we have more hands working on this (@adam-stokes joined the team two weeks ago, and you), we can work on integrations more and more.

mdelapenya:
> Btw, I think that the code used to verify if it is healthy does work. What doesn't work quite well is the way in which Elastic Agent is ran / installed (I think)

@michalpristas @blakerouse any guess why running the install command with fleet-server flags is not causing the enrollment in this use case?

jalvz:
@blakerouse Is `./elastic-agent -e -v install --force --insecure --enrollment-token=<TOKEN> --fleet-server-insecure-http --fleet-server-host=0.0.0.0 --fleet-server-policy=<policy> -k http://elastic:changeme@localhost:5601` the right way to install and enroll Elastic Agent with Fleet Server mode? 
If not, what is the right way?
(Also: what is `--url` for? Kibana/Elasticsearch/other...?)

jalvz:
doh didn't see the message from 3 mins ago, sorry for double ping

mdelapenya:
> @blakerouse Is `./elastic-agent -e -v install --force --insecure --enrollment-token=<TOKEN> --fleet-server-insecure-http --fleet-server-host=http://0.0.0.0 --fleet-server-policy=<policy> -k http://elastic:changeme@kibana:5601` the right way to install and enroll Elastic Agent with Fleet Server mode?
> If not, what is the right way?
> (Also: what is `--url` for? Kibana/Elasticsearch/other...?)

We used this thread as reference: https://github.com/elastic/e2e-testing/issues/438#issuecomment-794107349 maybe is it obsolete?

mdelapenya:
Interesting 🤔. I'm running the following command with this error:

```shell
[root@ddfd8c1c519b elastic-agent]$ /elastic-agent/elastic-agent install -e -v --force --insecure --enrollment-token=VFMxeWIzZ0JvbDc5RTFUUVQ1b2Y6THVndlJMUXRTUXlSQzVOMFFWVTE3QQ== --fleet-server-insecure-http --fleet-server http://elastic:changeme@elasticsearch:9200  --fleet-server-policy df11f2d0-8e52-11eb-877c-4397835e8a98 --kibana-url http://elastic:changeme@kibana:5601
The Elastic Agent is currently in BETA and should not be used in production

2021-03-26T16:53:17.931Z        INFO    [composable]    composable/controller.go:44     EXPERIMENTAL - Inputs with variables are currently experimental and should not be used in production
2021-03-26T16:53:18.036Z        INFO    [composable.providers.docker]   docker/docker.go:43     Docker provider skipped, unable to connect: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
2021-03-26T16:53:18.138Z        INFO    capabilities/capabilities.go:59 capabilities file not found in /elastic-agent/capabilities.yml
Error: failed to communicate with elastic-agent daemon; is elastic-agent running?
2021-03-26T16:53:19.099Z        INFO    [composable]    composable/controller.go:44     EXPERIMENTAL - Inputs with variables are currently experimental and should not be used in production
2021-03-26T16:53:19.202Z        INFO    [composable.providers.docker]   docker/docker.go:43     Docker provider skipped, unable to connect: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
2021-03-26T16:53:19.304Z        INFO    capabilities/capabilities.go:59 capabilities file not found in /elastic-agent/capabilities.yml
2021-03-26T16:53:19.452Z        INFO    [composable]    composable/controller.go:44     EXPERIMENTAL - Inputs with variables are currently experimental and should not be used in production
2021-03-26T16:53:19.554Z        INFO    [composable.providers.docker]   docker/docker.go:43     Docker provider skipped, unable to connect: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
2021-03-26T16:53:19.655Z        INFO    capabilities/capabilities.go:59 capabilities file not found in /elastic-agent/capabilities.yml
Error: enroll command failed with exit code: 1
```

First, it seems there is no `capabilities.yml` file in the layout.

```shell
[root@ddfd8c1c519b elastic-agent]$ find /elastic-agent -name "*.yml"
./fleet.yml
./elastic-agent.yml
./elastic-agent.reference.yml
```

Last, a super tiny thing: the Error says `enroll command failed`, when it used the `install` command.

jalvz:
What do we need here to move this forward?

adam-stokes:
From what I can tell we should be using the `--url` flag instead of `--fleet-server*` for enrolling

EricDavisX:
my understanding is that the Docker usage should be ready on the product code side, and we can finish the work on e2e-testing side and get it merged in.  Just need time and maybe some luck (I say that because I have not personally seen Fleet Server working).  I updated https://github.com/elastic/e2e-testing/issues/438 with my notes to the best of my knowledge.  Hope to have more soon.

mdelapenya:
@EricDavisX @adam-stokes @jalvz This PR is failing because the way we are bootstrapping the initial elastic-agent is wrong. We would need the canonical command that does that:

```gherkin
When I want to bootstrap the fleet server in a host
Then I run the FOO command with FLAGS and VALUES
```

As a follow-up, once this is merged and passing, we'd like to add scenarios for another agent using the bootstrapped fleet-server. Something like:

```gherkin
Given a fleet-server already bootstrapped
When I enroll a second agent in Fleet using the fleet-server
Then the agent is "online" in Fleet
```

adam-stokes:
According to the documentation we should be starting up the environment, copying over the `elastic-agent.yml` to the directory where we run `elastic-agent install` from:

https://www.elastic.co/guide/en/fleet/7.12/run-elastic-agent-standalone.html

That should do the bootstrapping and allow us to continue.  We will also have to workaround https://github.com/elastic/beats/issues/24950 until that is fixed as well

blakerouse:
@adam-stokes You should not need a custom `elastic-agent.yml` at all when running in Fleet mode. Elastic Agent will not even read it, it will just copy it to make a backup and write a new one.

e2e/_suites/fleet/features/fleet_server.feature
EricDavisX: The check can be the same as for all Agents, that it is listed as 'healthy' in the Agents API list call.  A secondary check could be to assess if the Fleet Server process is running on the host, as noted in this example:

```
edavis-mbp:elastic-agent-8.0.0-SNAPSHOT-darwin-x86_64-infra-build edavis$ ps ax | grep elastic
SNAPSHOT-darwin-x86_64/fleet-server --agent-mode -E logging.level=info -E http.enabled=true -E http.host=unix:///Library/Elastic/Agent/data/tmp/default/fleet-server/fleet-server.sock -E logging.json=true -E logging.ecs=true -E logging.files.path=/Library/Elastic/Agent/data/elastic-agent-53d75c/logs/default -E logging.files.name=fleet-server-json.log -E logging.files.keepfiles=7 -E logging.files.permission=0640 -E logging.files.interval=1h -E path.data=/Library/Elastic/Agent/data/elastic-agent-53d75c/run/default/fleet-server--8.0.0-SNAPSHOT
```
This example is from macOS, but the the process name is the same.

e2e/_suites/fleet/features/fleet_server.feature
EricDavisX: Another note about this pr vs a separate one...  we have a new feature file which is helpful, but to me, it is mostly helpful in confirming we have the code right here, with a straightforward usage.  We'll need to adapt (now or in a separate PR) all of the:
  Given a "<os>" agent is deployed to Fleet with "tar" installer
to
  When a "<os>" agent is deployed to Fleet with "tar" installer in fleet-server mode
...because any non-Fleet server usage (in Fleet) will not be supported.  The only non-Fleet-Server usage will be as a stand-alone Agent mode.

So, all the following feature files will need review and update, at least in some way (listing them out explicitly):
e2e/_suites/fleet/features/agent_endpoint_integration.feature
e2e/_suites/fleet/features/backend_processes.feature
e2e/_suites/fleet/features/linux_integration.feature
e2e/_suites/fleet/features/fleet_mode_agent.feature

The stand_alone_agent file is the exception:
e2e/_suites/fleet/features/stand_alone_agent.feature
 - we have opportunity there to improve the stand-alone test to include Docker usage with Fleet Server.  And it may be a great and easy way to spin up a 2nd Agent to connect to the first (the first Agent also running the Fleet Server, and the 2nd being a 'normal' Agent)

So, knowing this now, I'm not sure if we want to keep this first PR small and merge it so we have some passing scenario knowing the others will fail, or if we'll want to push forward with slightly larger impact across more files after we see it working.

EricDavisX:
there is a lot to unpack here, I wanted to finally chime in more with some thoughts.  This comment from you @mdelapenya to @jalvz I wanted to clarify some bits:

```
Not sure if the latter should be done at the unit level in libbeat.
```
I think the libbeat code is not the same as the Agent - and we can't test all or much of Agent by hitting libbeat itself or any libbeat / beat (only) installer.

```
Secondly: even if we maintain the install tests, why the other tests (enrolling, ingestion etc.) go trough custom installers? Since it all runs on Docker, shouldn't we just run the stock images as they are?

I'm probably missing something about multiple platform support, but well, that's what the questions are for :)

There are two modes, and please correct me @EricDavisX if I'm wrong: stand-alone and fleet. While the stand-alone mode uses the stock images (docker.elastic.co/beats/elastic-agent:8.0.0-SNAPSHOT), which is automatically auto-enrolled in Fleet when started, the fleet mode works differently in the way that represents a valid use case: As a sysadmin, I have my centos machine and want to install the agent and enroll it. For that reason this fleet mode uses a plain Centos container and perform the installation of the artifacts (TAR, RPM, DEB) first.
```
Indeed there are many different artifacts all of which we want to test.  Docker is just ONE way to run the stand-alone Agent, but any artifact can be used with a config that does NOT include Fleet usage and it will be a 'stand-alone' mode Agent.   We are using the Docker container as our stand-alone test, because it was easy to do so (and we just happened to code it up first, as such).  We could expand the Docker Container tests to NOT be stand-alone mode and this would be a nice enhancement, but we have so many priorities, it has not gotten high in the list yet.

Also wanted to call out my prior comment that all Agents (except Stand-alone mode Agents) will need to connect to a Fleet Server to send data.  Coming ASAP that will be the requirement, so we'll have to consider making each test (except for Stand-alone mode if easier to leave it alone for now) to be a Fleet-Server using test.  The implication there too is that the 'real' test for Fleet Server is to have a second container or Agent connect to a separate Fleet Server (Agent) to include that communication part of the architecture (which was formerly in Kibana but is now on the edge, on the host, in Fleet Server, running as part of Agent).

EricDavisX:
> @EricDavisX @adam-stokes @jalvz This PR is failing because the way we are bootstrapping the initial elastic-agent is wrong. We would need an example of the canonical command that does that:
> 
> ```
> When I want to bootstrap the fleet server in a host
> Then I run the FOO command with FLAGS and VALUES
> ```
> 
> As a follow-up, once this is merged and passing, we'd like to add scenarios for another agent using the bootstrapped fleet-server. Something like:
> 
> ```
> Given a fleet-server already bootstrapped
> When I enroll a second agent in Fleet using the fleet-server
> Then the agent is "online" in Fleet
> ```

I have not seen it working end to end yet but we're close and I think just waiting on a new build from Infa, so by the time this is read on Wednesday Apr 14 it should work!  Unless there are more bugs.  So, the code I know to bootstrap the Agent, including Fleet Server process, using 'install' command is:
./elastic-agent install -f --fleet-server https://elastic:{{elastic_password}}@{{es_url}}:9200"

Then, with a healthy Fleet Server (Agent) running, the code to install a non-Fleet Server Agents to use it is:
./elastic-agent install -f --url=https://{{fleet-server-host}}:8220 --enrollment-token={{ enroll_token }}"
 - the enrollment token should relate to a policy with Fleet Server integration.

NOTE, if we are using insecure communications between the Agent(s) and Fleet Server we can remove the 's' in the https and use the below parameter at the end of the command:
--fleet-server-insecure-http
^ the above goes in the Fleet Server bootstrap line. And then the relating line for Agent install, you use: 
--insecure
Using https and security is the default.

Also - there is a new Fleet level param to be set in a Kibana API, we should already be using this, but it can be extended to include the Fleet Server URL (planning ahead for whichever host we want to use - if we don't do this, no other host can connect):
PUT to "https://{{ kibana_domain_name }}/api/fleet/settings"
 with body:
{"fleet_server_hosts": ["https://{{fleet-server-host}}:8220"]}
This is the same API where we are currently setting the Kibana URLs, so it may be changing.  Also changing is the parameter '--fleet-server' maybe changing to '--fleet-server-es' to make it more clear that it is actually the es connection (that fleet server uses)

blakerouse:
> * the enrollment token should relate to a policy with Fleet Server integration.

The enrollment token for the Elastic Agent not running Fleet Server should be for the Default policy, not the Fleet Server default policy.

EricDavisX:
Ah, thank you Blake!

mdelapenya:
I've seen this error to fail:
> ./elastic-agent install -f --fleet-server [https://elastic:{{elastic_password}}@{{es_url}}:9200](https://elastic:%7B%7Belastic_password%7D%7D@%7B%7Bes_url%7D%7D:9200)"

```
Error: unknown flag: --fleet-server
Usage:
  elastic-agent install [flags]

Flags:
  -p, --ca-sha256 string                    Comma separated list of certificate authorities hash pins used for certificate verifications
  -a, --certificate-authorities string      Comma separated list of root certificate for server verifications
  -t, --enrollment-token string             Enrollment token to use to enroll Agent into Fleet
      --fleet-server-cert string            Certificate to use for exposed Fleet Server HTTPS endpoint
      --fleet-server-cert-key string        Private key to use for exposed Fleet Server HTTPS endpoint
      --fleet-server-es string              Start and run a Fleet Server along side this Elastic Agent connecting to the provided elasticsearch
      --fleet-server-es-ca string           Path to certificate authority to use with communicate with elasticsearch
      --fleet-server-host string            Fleet Server HTTP binding host (overrides the policy)
      --fleet-server-insecure-http          Expose Fleet Server over HTTP (not recommended; insecure)
      --fleet-server-policy string          Start and run a Fleet Server on this specific policy
      --fleet-server-port uint16            Fleet Server HTTP binding port (overrides the policy)
      --fleet-server-service-token string   Service token to use for communication with elasticsearch
  -f, --force                               Force overwrite the current and do not prompt for confirmation
  -h, --help                                help for install
  -i, --insecure                            Allow insecure connection to Kibana
  -k, --kibana-url string                   URL of Kibana to enroll Agent into Fleet
      --staging string                      Configures agent to download artifacts from a staging build
      --url string                          URL to enroll Agent into Fleet

Global Flags:
  -c, --c string                     Configuration file, relative to path.config (default "elastic-agent.yml")
  -d, --d string                     Enable certain debug selectors
  -e, --e                            Log to stderr and disable syslog/file output
      --environment environmentVar   set environment being ran in (default default)
      --path.config string           Config path is the directory Agent looks for its config file (default "/elastic-agent")
      --path.home string             Agent root path (default "/elastic-agent")
      --path.logs string             Logs path contains Agent log output (default "/elastic-agent")
  -v, --v                            Log at INFO level
```
I think you mean `--fleet-server-host` and `--fleet-server-port`, right? I see `--fleet-server-service-token`, is this required to bootstrap the fleet server agent?

mdelapenya:
Also running this command starts the agent but not bootstrapped:

```shell
/elastic-agent/elastic-agent install \
   --force --fleet-server-insecure-http --fleet-server-host elasticsearch --fleet-server-port 9200
```

EricDavisX:
The last build had a check-in to change the parameter from fleet-server to fleet-server-es.  that much I know.  I am trying to gather communication about pending / coming changes but it is a challenging effort as we didn't have Infra builds to validate with until today.

blakerouse:
@mdelapenya @EricDavisX  is correct it is now `--fleet-server-es` versus previous `--fleet-server`.

cli/config/compose/profiles/fleet/configurations/kibana.config.yml
```diff
@@ -15,5 +15,6 @@ xpack.fleet.enabled: true
 xpack.fleet.registryUrl: http://package-registry:8080
 xpack.fleet.agents.enabled: true
 xpack.fleet.agents.elasticsearch.host: http://elasticsearch:9200
+xpack.fleet.agents.fleetServerEnabled: true
 xpack.fleet.agents.kibana.host: http://kibana:5601
```
ruflin, line 19: You might need `xpack.fleet.agents.fleet_server.hosts: [http://?:5601]` instead here with the most recent builds. Probably worth rerunning.

PR #925, opened 2021-03-24
fix: bump stale agent version to 7.11.2

mdelapenya:
## What does this PR do?
It bumps stale agent version from 7.11.1 to 7.11.2

## Why is it important?
7.11.1 dissapeared from https://artifacts-api.elastic.co/v1/search/7.11.1/elastic-agent

## Checklist


cachedout (approved):

EricDavisX:
Thanks so much Manu.  If we ever wanted to review deeper, I think the 'shipped' versions should be findable longer, if we can identify the better artifactory url to pull from, like this:
https://artifacts-api.elastic.co/v1/versions/7.11/builds/latest

Using this would always pull 'the latest' from the 7.11.x line.  And further, to get the semVer in the prior url, I still think we could find a way to do string to number magic and back to string to make it dynamic of major release minus 1.  Just takes time I know tho.

PR #968, opened 2021-03-30
ci: use withNode step to avoid reusing workers

v1v:
## What does this PR do?

Use the `withNode` step (https://github.com/elastic/apm-pipeline-library/blob/master/vars/withNode.groovy) 

## Why is it important?

Add self-resilience when the provisioner allocates an existing worker instead assigning a new one.

```
Step the file system Agent folder is empty
```

image

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -286,7 +286,7 @@ def generateStep(Map args = [:]){
   def oss = args.get('oss')
   def platform = args.get('platform')
   return {
-    withNode(labels: 'ubuntu-18.04 && immutable && docker', sleepMax: 20, forceWorspace: true){
+    withNode(labels: 'ubuntu-18.04 && immutable && docker', sleepMax: 20, forceWorkspace: true){
       try {
         deleteDir()
         unstash 'source'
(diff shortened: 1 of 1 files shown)


.ci/Jenkinsfile
```diff
@@ -295,8 +295,6 @@ def generateStep(Map args = [:]){
             sh script: 'make build', label: 'Create releases'
           }
         }
-      } catch(e) {
-        error(e.toString())
```
v1v, line 299: There is no need to catch the error so we reduce two step errors: the original one and the error step.

kuisathaverat (approved):

.ci/Jenkinsfile
```diff
@@ -325,7 +323,7 @@ def generateFunctionalTestStep(Map args = [:]){
   def workerLabels = "${platform} && immutable && docker"
 
   return {
-    node("${workerLabels}") {
+    withNode(labels: "${workerLabels}", sleepMax: 20, forceWorspace: true){
```
v1v, line 326:
```suggestion
    withNode(labels: "${workerLabels}", sleepMax: 20, forceWorkspace: true){
```

kuisathaverat (approved):

PR #983, opened 2021-04-01
Remove the agent config file parameters for stand alone

### LINKED ISSUE #981: device or resource busy when attempting to load elastic-agent config in docker
In some instances like stand-alone testing and running from the cli we hit this error:

```
Performing setup of Fleet in Kibana
The Elastic Agent is currently in BETA and should not be used in production

Error: could not save enrollment information: could not backup /usr/share/elastic-agent/elastic-agent.yml: rename /usr/share/elastic-agent/elastic-agent.yml /usr/share/elastic-agent/elastic-agent.yml.2021-04-01T12-38-10.4257.bak: device or resource busy
Error: enrollment failed: exit status 1
```

If we remove that volume mount everything works as intended, however, this makes it a little difficult to include the default system integration

adam-stokes:

## What does this PR do?

This removes the volume mounting of elastic-agent.yml for fleet server stand alone. The side effect is that we will have to think about how we want to approach fleet mode outside of that. Currently in stand alone that config file is not useful and it causes device and resource errors when mounting inside of docker when operations within the elastic-agent occur.

I think for the other use cases we can tests integrations etc more dynamically than having elastic-agent.yml mounted as a volume. Thoughts?

## Why is it important?

Testing standalone fleet server will fail if this configuration and volume exists

## Checklist

## Related issues
- Closes #981

### DIFF
--- a/e2e/_suites/fleet/stand-alone.go
+++ b/e2e/_suites/fleet/stand-alone.go
@@ -7,7 +7,6 @@ package main
 import (
 	"context"
 	"fmt"
-	"os"
 	"strings"
 	"time"
 
@@ -22,10 +21,9 @@ import (
 
 // StandAloneTestSuite represents the scenarios for Stand-alone-mode
 type StandAloneTestSuite struct {
-	AgentConfigFilePath string
-	Cleanup             bool
-	Hostname            string
-	Image               string
+	Cleanup  bool
+	Hostname string
+	Image    string
 	// date controls for queries
 	AgentStoppedDate             time.Time
 	RuntimeDependenciesStartDate time.Time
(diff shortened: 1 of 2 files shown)


cachedout:
I'm not totally sure I follow the explanation of the original problem. Perhaps it would be good to file an issue with a more detailed explanation of the problem so that it's easier to put this PR in the proper context?

e2e/_suites/fleet/stand-alone.go
```diff
@@ -102,11 +85,6 @@ func (sats *StandAloneTestSuite) startAgent(image string, env map[string]string)
 		dockerImageTag += "-amd64"
 	}
 
-	configurationFilePath, err := steps.FetchBeatConfiguration(true, "elastic-agent", "elastic-agent.docker.yml")
```
mdelapenya, line 105: I think if we convert the file int a static one, instead of dynamic, we will lose the ability to use the elastic-agent configuration file from a PR, in the case it's modified.

mdelapenya:
I'd say it's not ok yet to remove the file from the e2e tests. If the original use case is to use what it's bundled into the image, we could move to an approach where we _dockerCopy_ the file into the container **only for PRs**. Wdyt?

adam-stokes:
> I'd say it's not ok yet to remove the file from the e2e tests. If the original use case is to use what it's bundled into the image, we could move to an approach where we _dockerCopy_ the file into the container **only for PRs**. Wdyt?

I think this only effects running fleet in stand alone mode. For all other scenarios we can use the volumes mount directive. I still don't know why docker gives us a resource busy error in stand alone mode when the `elastic-agent.yml` is modified though. But not using the volumes and doing a docker copy would workaround this issue, I just don't know if this is the right approach we should take.

Another idea would be to have 2 separate profiles, one for fleet and another for fleet in standalone mode.

e2e/_suites/fleet/stand-alone.go
mdelapenya: I rollback my previous commit with #981

mdelapenya: Can we check other usages off the FetchBeatConfiguration method? If it's not needed in the metricbeat suite, then we could completely remove it

jalvz:
> I think for the other use cases we can tests integrations etc more dynamically than having elastic-agent.yml mounted as a volume. Thoughts?

I like that, I generally think passing arguments is more flexible that mounting config files, + we don't need it for standalone.

jalvz:
fyi I think you need to merge master to make tests pass.

mdelapenya:
I've tested this PR locally and I think it's the right think to do, as explained in https://github.com/elastic/e2e-testing/issues/981#issuecomment-814214259.

I'm going to merge it. Thanks for your work here!

PR #992, opened 2021-04-06
consolidate kibana api to internal project structure

adam-stokes:
Add all relevant kibana api calls into the golang project standard of
`internal/kibana`.

Signed-off-by: Adam Stokes <51892+adam-stokes@users.noreply.github.com>

## What does this PR do?

This PR is the start of some refactoring to get the project in a more standardized layout. It adds all the related kibana api calls into `internal/kibana` and also has additional api calls added to abstract out more of the work that is being handled in the testsuites. Currently, this is only an additive PR to get the bits in place for a further refactor of the testsuites and cli.

## Why is it important?

It is nice to have all the relevant api calls under a single location that is easily identifiable

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
~~- [ ] I have made corresponding change to the default configuration files~~
~~- [ ] I have added tests that prove my fix is effective or that my feature works~~
~~- [ ] I have run the Unit tests for the CLI, and they are passing locally~~
- [x] I have run the End-2-End tests for the suite I'm working on, and they are passing locally
~~- [ ] I have noticed new Go dependencies (run `make notice` in the proper directory)~~

## How to test this PR locally

Nothing to test at this moment as there are several subsequent PR's that will utilize this new api in the scenario testsuites

## Related issues

- Relates #959 

## Follow-ups

Once this PR is approved and merge I have several additional changes that make use of this code in the testsuites.

This is part of a larger refactor https://github.com/elastic/e2e-testing/compare/feat-kibana-refactor that gives a better idea of where these new additions will be utilized and which of the original code is moved/removed. Most of the code related to this PR lives in `cli/services/kibana.go` and additional code lives in `e2e/_suites/fleet/fleet.go` `e2e/_suites/fleet/integrations.go`. In the referenced `feat-kibana-refactor` branch you can get a better idea of how that looks and get a feel for the direction I am taking.

PR #1008, opened 2021-04-09
v2 refactor

adam-stokes:

## What does this PR do?

This PR refactors a lot of the internal code into more organized packages under `internal` top level directory.

## Why is it important?

As we grow the project it makes it a bit easier to reason about what code is doing what and provides a central location to find function references and definitions.

## Checklist

## Author's Checklist

- [x] Tag current master as `1.99.0` prior to merging
- [ ]  Verify current e2e tests pass

The standard godog test instructions apply here:

```
cd e2e/_suites/fleet
godog -t '@fleed_mode_agent'
```

## Related issues

- Supersedes #992


e2e/_suites/fleet/features/backend_processes.feature
```diff
@@ -7,169 +7,169 @@ Scenario Outline: Deploying the <os> agent
   Given a "<os>" agent is deployed to Fleet with "tar" installer
```
mdelapenya, line 7: I'd prefer not touching these files in this PR, to make it easier the review

e2e/_suites/helm/helm_charts_test.go
```diff
@@ -81,13 +83,12 @@ func setupSuite() {
...
 	stackVersion = shell.GetEnv("STACK_VERSION", stackVersion)
-	stackVersion = e2e.GetElasticArtifactVersion(stackVersion)
+	stackVersion = utils.GetElasticArtifactVersion(stackVersion)
 
-	h, err := k8s.HelmFactory(helmVersion)
+	_, err := helm.Factory(helmVersion)
 	if err != nil {
 		log.Fatalf("Helm could not be initialised: %v", err)
 	}
-	helm = h
```
mdelapenya, line 90:
```suggestion
h, err := helm.Factory(helmVersion)
if err != nil {
	log.Fatalf("Helm could not be initialised: %v", err)
}
helmManager = h
```
I think this would resolve the panic caused by not initialising the helm manager.

adam-stokes:
@mdelapenya with the new fleet server updates, a lot of the initial steps code here doesn't work any longer so I am unable to finish testing this branch without including the work from your #900 in addition to even more refactoring for the agents themselves. How would you like to handle this? I feel like we should merge this refactor and then build out the fleet server code on top and the additional agent testing after that. I think there is still significant work required to fix the agent code to connect to the new fleet server anyway, wdyt?

e2e/_suites/fleet/fleet.go
```diff
@@ -157,7 +162,7 @@ func (fts *FleetTestSuite) contributeSteps(s *godog.ScenarioContext) {
 	s.Step(`^the "([^"]*)" datasource is shown in the policy as added$`, fts.thePolicyShowsTheDatasourceAdded)
 	s.Step(`^the host name is shown in the Administration view in the Security App as "([^"]*)"$`, fts.theHostNameIsShownInTheAdminViewInTheSecurityApp)
 	s.Step(`^the host name is not shown in the Administration view in the Security App$`, fts.theHostNameIsNotShownInTheAdminViewInTheSecurityApp)
-	s.Step(`^an Endpoint is successfully deployed with a "([^"]*)" Agent using "([^"]*)" installer$`, fts.anEndpointIsSuccessfullyDeployedWithAgentAndInstalller)
+	s.Step(`^an "([^"]*)" is successfully deployed with a "([^"]*)" Agent using "([^"]*)" installer$`, fts.anEndpointIsSuccessfullyDeployedWithAgentAndInstalller)
```
mdelapenya, line 165: Because we abstracted the integration name (converting "Endpoint" into a variable), I'd rename the method too:

```diff
- anEndpointIsSuccessfullyDeployedWithAgentAndInstalller
+ anIntegrationIsSuccessfullyDeployedWithAgentAndInstalller
```

internal/installer/apt.go
```diff
@@ -0,0 +1,131 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
mdelapenya, line 1: Wdyt about renaming the file to `deb` (the binary extension)? We are already doing the same for rpm and tar. Otherwise we are doing an abstraction about the package manager (apt) where for RPM we are not using yum

mdelapenya (commented):
I left some naming and location comments. 

It's a massive refactor, although the commits structure helped me in the review, thanks for that.

I also noticed some changes in behaviour, like splitting methods (like `getIntegrationLatestVersion`). From review standpoint, I'd have preferred a more straightforward refactor for moving things to the new package, although I see the value brought by this new separation of concerns.

LGTM after we fix the failing test scenarios, which seems related to running all tests in a row

adam-stokes:
@mdelapenya I think CI is using an older version of kibana for testing now, because when I try to test locally I can't get the tests to run without utilizing the new fleet server and updated api calls. How can I use the version of kibana/elastic that is referenced in CI, locally?

mdelapenya:
> @mdelapenya I think CI is using an older version of kibana for testing now, because when I try to test locally I can't get the tests to run without utilizing the new fleet server and updated api calls. How can I use the version of kibana/elastic that is referenced in CI, locally?

IIRC it fetches docker.elastic.co/kibana/kibana:8.0.0-SNAPSHOT by default. Check the ci logs when pulling them image, maybe you can get the hash pulled by docker compose.

adam-stokes:
> IIRC it fetches docker.elastic.co/kibana/kibana:8.0.0-SNAPSHOT by default. Check the ci logs when pulling them image, maybe you can get the hash pulled by docker compose.

Ah it looks like kibana was updated and now the same errors I'm seeing locally are showing up in CI. Not sure how to tests this PR any further without doing the work for Fleet Server :(

mdelapenya:
> > IIRC it fetches docker.elastic.co/kibana/kibana:8.0.0-SNAPSHOT by default. Check the ci logs when pulling them image, maybe you can get the hash pulled by docker compose.
> 
> Ah it looks like kibana was updated and now the same errors I'm seeing locally are showing up in CI. Not sure how to tests this PR any further without doing the work for Fleet Server :(

There is this log in the artifacts (https://beats-ci.elastic.co/job/e2e-tests/job/e2e-testing-mbp/job/master/727/artifact/docker_logs_fleet_agent_endpoint_integration.log): 

>[fleet_kibana_1][docker.elastic.co/kibana/kibana:8.0.0-SNAPSHOT][92ea04a8160be3afdd0c164e27df2e4a9ff6ade7d45878c0e094081f4fc541e1][2021-04-19T05:43:07.305Z]  FATAL  Error: [config validation of [xpack.fleet].agents.kibana]: definition for this key is missing

adam-stokes:
> [fleet_kibana_1][docker.elastic.co/kibana/kibana:8.0.0-SNAPSHOT][92ea04a8160be3afdd0c164e27df2e4a9ff6ade7d45878c0e094081f4fc541e1][2021-04-19T05:43:07.305Z] FATAL Error: [config validation of [xpack.fleet].agents.kibana]: definition for this key is missing

I think this is because we have to use the new config options:

```
xpack.fleet.agents.fleet_server.hosts:
  - http://kibana:5601
```

adam-stokes:
@mdelapenya I merged master with fleet server and got it worked fully in this branch. I've also switched over the tests to use the fleet server bootstrap since the tests can not accommodate for a fleet agent and a regular agent connected to the fleet. Good news is the backend processes tests are passing where they were failing, however, a couple other tests are now failing due to enrollment/re-enrollment.

e2e/_suites/fleet/fleet.go
```diff
@@ -41,6 +41,7 @@ type FleetTestSuite struct {
 	Installers          map[string]installer.ElasticAgentInstaller
 	Integration         kibana.IntegrationPackage // the installed integration
 	Policy              kibana.Policy
+	FleetPolicy         kibana.Policy
```
mdelapenya, line 44: Good catch! At first place, the fleet config struct calculated the fleet policy on creation, and it was removed

e2e/_suites/fleet/fleet.go
```diff
@@ -307,15 +316,15 @@ func (fts *FleetTestSuite) anAgentIsDeployedToFleetWithInstallerAndFleetServer(i
 	containerName := fmt.Sprintf("%s_%s_%s_%d", profile, fts.Image+"-systemd", serviceName, 1) // name of the container
 
 	// enroll the agent with a new token
-	enrollmentKey, err := fts.kibanaClient.CreateEnrollmentAPIKey(fts.Policy)
+	enrollmentKey, err := fts.kibanaClient.CreateEnrollmentAPIKey(fts.FleetPolicy)
```
mdelapenya, line 319: Mmm, IIRC @blakerouse mentioned the token must be provided by the default policy, not the default fleet server policy.

adam-stokes: I think for the agent that is bootstrapping fleet server it was supposed to use the fleet server policy and all other agents use the default policy, @blakerouse is that correct?

adam-stokes: I verified with @EricDavisX that this should indeed point to the fleet server policy when bootstrapping an agent with fleet server

mdelapenya: Cool then. I took a note and added a comment in the src pointing to that. This should be changed then

adam-stokes:
Another bug hit: https://github.com/elastic/fleet-server/issues/245 putting here to track

internal/kibana/agents.go
```diff
@@ -100,70 +101,81 @@ func (c *Client) GetAgentStatusByHostname(hostname string) (string, error) {
 
 // GetAgentEvents get events of agent
 func (c *Client) GetAgentEvents(applicationName string, agentID string, packagePolicyID string, updatedAt string) error {
-	statusCode, respBody, err := c.get(fmt.Sprintf("%s/agents/%s/events", FleetAPI, agentID))
-
+	query := map[string]interface{}{
```
mdelapenya, line 104: I'm OK with this change, although not sure if we still want to test the kibana APIs for that. Are we sure the API disappeared instead of changed? @EricDavisX @nchaulet could you please confirm?

e2e/_suites/fleet/fleet.go
```diff
@@ -105,7 +105,8 @@ func (fts *FleetTestSuite) afterScenario() {
 		}).Error("The package policies could not be found")
 	}
 	for _, pkgPolicy := range packagePolicies {
-		if pkgPolicy.PolicyID == fts.FleetPolicy.ID {
+		// Do not remove the fleet server package integration otherwise fleet server fails to bootstrap
+		if !strings.Contains(pkgPolicy.Name, "fleet_server") && pkgPolicy.PolicyID == fts.FleetPolicy.ID {
```
mdelapenya, line 109: Good catch!

internal/kibana/agents.go
nchaulet: Yes that APIs disappear, there is no agents events saved object on our side.

mdelapenya: @adam-stokes what I'm not sure is if we want to make this change in this refactor, as we pinned the kibana version until we fix the rest of the enroll scenarios 🤔

mdelapenya: If that API disappeared, maybe we can conclude that the scenario steps related to it could disappear too. @EricDavisX are your thoughts on this? Should we keep steps related to agent events after the removal of the API? Apart from the ES query that Adam brightly added, is there any other way to get those events?

adam-stokes: > @adam-stokes what I'm not sure is if we want to make this change in this refactor, as we pinned the kibana version until we fix the rest of the enroll scenarios 🤔

This change was required for the agent endpoint test suite to pass when adding the endpoint policy

nchaulet: No there is no other way to get these logs otherwise than querying ES or going to the logs UI. 
I think testing that Agent correctly send logs could make sense in e2e tests but it's  probably a new scenario step.

internal/kibana/agents.go
```diff
@@ -217,15 +217,12 @@ func (c *Client) ListAgents() ([]Agent, error) {
...
+func (c *Client) UnEnrollAgent(hostname string) error {
 	agentID, err := c.GetAgentIDByHostname(hostname)
 	if err != nil {
 		return err
 	}
-	reqBody := `{"force": false}`
-	if force {
-		reqBody = `{"force": true}`
-	}
+	reqBody := `{"revoke": true}`
```
mdelapenya, line 225: Neat!!

PR #1016, opened 2021-04-13
fix: delay checking stale agent version until it's used

### LINKED ISSUE #1015: Do not fail the entire suite when the stale agent test cannot find the stale agent version
The code is logging.Fatal when a runtime dependency is not satisfied, and the stale agent version is considered a runtime dep. But it causes to fail the entire suite instead of failing the single upgrade scenario.

mdelapenya:
## What does this PR do?
It moves the initialisation of the stale agent version from the setup suite phase (before any scenario) to the only step it's used (deploying an stale agent to      Fleet).

Besides that, we detected that the namespace for the custom kibana image was not properly set, so we are making sure the custom version starts with "pr" (it's a PR)
to use the proper Docker namespace.

Finally, we are changing the signature of the method to check if a version is an alias, returning the error so that the caller handles it in whatever manner it needs, instead of handling it in the internal method.

## Why is it important?
It will avoid failing all scenarios if the stale agent version is not found. Instead it will fail the step. Therefore setup methods will fail the setup with `log.Fatal`, and steps will simply return the error, failing that scenario only.

## Checklist

```shell
# stale agent, only
SUITE="fleet" TAGS="fleet_mode_agent && nightly" TIMEOUT_FACTOR=3 LOG_LEVEL=TRACE BEATS_USE_CI_SNAPSHOT=false DEVELOPER_MODE=true make -C e2e functional-test
# all suite for centos
SUITE="fleet" TAGS="fleet_mode_agent && centos" TIMEOUT_FACTOR=3 LOG_LEVEL=TRACE BEATS_USE_CI_SNAPSHOT=false DEVELOPER_MODE=true make -C e2e functional-test
```

## Related issues

- Fixes #1015

### DIFF
--- a/e2e/_suites/fleet/ingest_manager_test.go
+++ b/e2e/_suites/fleet/ingest_manager_test.go
@@ -7,7 +7,6 @@ package main
 import (
 	"context"
 	"os"
-	"strings"
 	"time"
 
 	"github.com/cucumber/godog"
@@ -36,15 +35,6 @@ func setUpSuite() {
 
 	timeoutFactor = shell.GetEnvInteger("TIMEOUT_FACTOR", timeoutFactor)
 	agentVersion = shell.GetEnv("BEAT_VERSION", agentVersionBase)
-
-	agentStaleVersion = shell.GetEnv("ELASTIC_AGENT_STALE_VERSION", agentStaleVersion)
-	// check if stale version is an alias
-	agentStaleVersion = e2e.GetElasticArtifactVersion(agentStaleVersion)
-
-	useCISnapshots := shell.GetEnvBool("BEATS_USE_CI_SNAPSHOTS")
-	if useCISnapshots && !strings.HasSuffix(agentStaleVersion, "-SNAPSHOT") {
-		agentStaleVersion += "-SNAPSHOT"
-	}
 
 	// check if version is an alias
 	agentVersion = e2e.GetElasticArtifactVersion(agentVersion)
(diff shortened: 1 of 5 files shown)


e2e/_suites/fleet/ingest_manager_test.go
```diff
@@ -32,31 +32,43 @@ func setUpSuite() {
 	}
 
 	// check if base version is an alias
-	agentVersionBase = e2e.GetElasticArtifactVersion(agentVersionBase)
+	v, err := e2e.GetElasticArtifactVersion(agentVersionBase)
```
cachedout, line 35: I'm confused by this line. Where is `agentVersionBase` coming from now that you've deleted it from https://github.com/elastic/e2e-testing/pull/1016/files#diff-3e08fc63f4ba9deffcc23be1d3bdfd0ef89646940c34dd47d06ab797ab6fa673L35 ? Am I missing something obvious?

mdelapenya: The `agentVersionBase` variable is initialised in https://github.com/elastic/e2e-testing/blob/07f6c4d1c44c64ca15a9f18431a969cf8675ced5/e2e/_suites/fleet/world.go#L29. We have a `world.go` file with the initial values for all constants, and during setup we potentially override them with the env vars provided by CI or the developer.

cachedout: Ah, so it's a global. Interesting. Thanks!

mdelapenya: @adam-stokes this is something we can improve for the future: how initial configs are created. I know [viper](https://github.com/spf13/viper) has support for initialising the project reading from the environment (and many others). Maybe this is something we can explore to simplify setting up the project. Thoughts?

cachedout (approved):

PR #1030, opened 2021-04-13
fix: use an agent when building kibana

mdelapenya:
## What does this PR do?
It uses an ubuntu 20 agent to build the job.

## Why is it important?
It fixes the following error on CI:

>ERROR: Attempted to execute a step that requires a node context while ‘agent none’ was specified. Be sure to specify your own ‘node { ... }’ blocks when using ‘agent none’.

## Checklist

### DIFF
--- a/.ci/e2eKibana.groovy
+++ b/.ci/e2eKibana.groovy
@@ -3,7 +3,7 @@
 @Library('apm@current') _
 
 pipeline {
-  agent { label 'ubuntu-20' }
+  agent none
   environment {
     REPO = 'kibana'
     BASE_DIR = "src/github.com/elastic/${env.REPO}"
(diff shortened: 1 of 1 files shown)


.ci/e2eKibana.groovy
```diff
@@ -3,7 +3,7 @@
 @Library('apm@current') _
 
 pipeline {
-  agent none
+  agent { label 'ubuntu-20' }
```
v1v, line 6: I'd say to leave none and use the agent in the stage itself

mdelapenya: I added a commit on top with your suggestion, thanks!

v1v (approved):

PR #1045, opened 2021-04-15
fix: typo in parameters

mdelapenya:
## What does this PR do?
It adds a comma separating parameters for the generic webhook trigger

## Why is it important?
My fault...

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have run the Unit tests for the CLI, and they are passing locally
- [ ] I have run the End-2-End tests for the suite I'm working on, and they are passing locally
- [ ] I have noticed new Go dependencies (run `make notice` in the proper directory)

## Related issues

- Caused by #1044

PR #1044, opened 2021-04-15
chore: add debug info for the payload

mdelapenya:
## What does this PR do?
It renames variable not to pollute variable namespace, adding log traces for getting the payload

## Why is it important?
It seems the `pull_request` field is not present in the payload when a comment is added to a PR, and this field is used to calculate the target branch and the current SHA of the PR

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have run the Unit tests for the CLI, and they are passing locally
- [ ] I have run the End-2-End tests for the suite I'm working on, and they are passing locally
- [ ] I have noticed new Go dependencies (run `make notice` in the proper directory)

## Author's Checklist

- [ ]

## Related issues

-

PR #1043, opened 2021-04-14
chore: add debug traces for the webhook payload

mdelapenya:
## What does this PR do?
It updates the buid cause string including HEAD_SHA, BASE_REF and PR_ID

I'd like to verify that the `pull_request` JSON object is returned in the weebhook payload.

## Why is it important?
We found the original error:

>hudson.AbortException: No item named e2e-tests/e2e-testing-mbp/null.x found

So this groovy block always evals to true, appending '.x' to the branch name, in this case `null`:
```groovy
def branchName = "${env.GT_BASE_REF}"
if (env.GT_BASE_REF != "master") {
  branchName = "${env.GT_BASE_REF}.x"
}
```

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have run the Unit tests for the CLI, and they are passing locally
- [ ] I have run the End-2-End tests for the suite I'm working on, and they are passing locally
- [ ] I have noticed new Go dependencies (run `make notice` in the proper directory)

PR #1049, opened 2021-04-19
fix: do not use GT_REPO variable, use fixed repo name instead

mdelapenya:
## What does this PR do?
It creates a global env var for repo name (`elastic/kibana`), which will be used instead of the usages of GT_REPO.

## Why is it important?
We noticed that, although the webhook event was properly captured by `regexpFilterText: '$GT_REPO$GT_BODY'`, at the moment we used it later in the pipeline, **for a manual trigger**, the `GT_REPO` variable was `null`. I.e.:

```groovy
def pullRequest = githubApiCall(token: token, url: "https://api.github.com/repos/${env.GT_REPO}/pulls/${prID}")
```

See https://beats-ci.elastic.co/blue/organizations/jenkins/e2e-tests%2Fe2e-testing-kibana-fleet/detail/e2e-testing-kibana-fleet/19/pipeline/

>githubApiCall: The REST API call https://api.github.com/repos/null/pulls/97041 return the message : java.lang.Exception: httpRequest: Failure connecting to the service https://api.github.com/repos/null/pulls/97041 : httpRequest: Failure connecting to the service https://api.github.com/repos/null/pulls/97041 : 

## Checklist

## Related issues

- Relates #1046 #1045 #1044 #1043 #1042 #1037

### DIFF
--- a/.ci/e2eKibana.groovy
+++ b/.ci/e2eKibana.groovy
@@ -29,8 +29,7 @@ pipeline {
       [key: 'GT_REPO', value: '$.repository.full_name'],
       [key: 'GT_PR', value: '$.issue.number'],
       [key: 'GT_BODY', value: '$.comment.body'],
-      [key: 'GT_COMMENT_ID', value: '$.comment.id'],
-      [key: 'GT_PAYLOAD', value: '$']
+      [key: 'GT_COMMENT_ID', value: '$.comment.id']
      ],
     genericHeaderVariables: [
      [key: 'x-github-event', regexpFilter: 'comment']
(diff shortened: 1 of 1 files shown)


.ci/e2eKibana.groovy
```diff
@@ -7,6 +7,7 @@ pipeline {
   environment {
     REPO = 'kibana'
     BASE_DIR = "src/github.com/elastic/${env.REPO}"
+    ELASTIC_REPO = "elastic/${env.REPO}"
```
v1v, line 10: If you use the the gitCheckout then ORG_NAME should be there and point to `elastic`:
- https://github.com/elastic/apm-pipeline-library/blob/8f48394fe31ae516841e25ea45b3faa94d147e96/vars/githubEnv.groovy#L79-L80
In case it helps you to move away from the hardcoded elastic

v1v (approved):

.ci/e2eKibana.groovy
mdelapenya: We are internally using `gitCheckout` in the buildKibana step, although it could be the case the image is already built and we skip that part

PR #1064, opened 2021-04-21
Kubernetes autodiscover suite

jsoriano:
## What does this PR do?

It adds a new suite to test Kubernetes autodiscover features.

It reproduces https://github.com/elastic/beats/issues/22718, for what we didn't have tests before.

It allows to define the scenarios using the gherkin language for feature files. Scenarios use templates that are mostly kubernetes manifests. A simple system of options is included to allow to select configuration blocks in these templates.

The objective is to make it easy to add more cases only by adding them to the feature files and providing new templates.

It uses kubectl to interact with a kubernetes cluster, if no cluster is available, it starts one with kind. Each scenario is executed in a different namespace, namespace is destroyed after the scenario is run to clean all created resources. 

It currently works only with loads in a single namespace, in a single node, this is enough to test many autodiscover cases, but it could be extended in the future to cover multi-node or multi-namespace cases.

It currently tests beats only, but it is not tied to beats in any way, it could be extended in the future to be used with agent.

## Why is it important?

Kubernetes autodiscover depends on the states of several components with many corner cases, and the transitions of these states. This complicates its implementation. And it is difficult to maintain a complete test suite.

The lack of a full featured test suite has caused the introduction of regressions in the past.

## Checklist

## Author's Checklist

- [x] Ensure that this is executed in CI jobs
- [x] Use proper beats images when running for PRs
- [x] `shell.Execute` is too verbose in case of error, consider alternatives.
- [x] ~~Consider reusing more features available in this repo.~~ TBD in follow ups.
- [x] Add some docs.
 
- Install `kind` and `kubectl`.
- Run `godog` from the `e2e/_suites/kubernetes-autodiscover` directory.

## Related issues

- Tests https://github.com/elastic/beats/pull/24742 

## Logs

With `BEAT_VERSION=7.12.0 godog` it is possible to reproduce https://github.com/elastic/beats/issues/22718:
```
  Scenario: Short-living cronjob                                                       # features/beats.feature:30
    Given a cluster is available                                                       # autodiscover_test.go:335 -> InitializeScenario.func3
    And configuration for "filebeat" has "hints enabled"                               # autodiscover_test.go:75 -> *podsManager
    And "filebeat" is running                                                          # autodiscover_test.go:130 -> *podsManager
    When "a short-living cronjob" is deployed                                          # autodiscover_test.go:116 -> *podsManager
    And "60s" have passed                                                              # autodiscover_test.go:336 -> InitializeScenario.func4
    Then "filebeat" collects events with "kubernetes.container.name:cronjob-container" # autodiscover_test.go:146 -> *podsManager
    timeout waiting for events with kubernetes.container.name:cronjob-container
```
And confirm that it passes with current master after https://github.com/elastic/beats/pull/24742.
```
  Scenario: Short-living cronjob                                                       # features/beats.feature:28
    Given a cluster is available                                                       # autodiscover_test.go:335 -> InitializeScenario.func3
    And configuration for "filebeat" has "hints enabled"                               # autodiscover_test.go:75 -> *podsManager
    And "filebeat" is running                                                          # autodiscover_test.go:130 -> *podsManager
    When "a short-living cronjob" is deployed                                          # autodiscover_test.go:116 -> *podsManager
    And "10s" have passed                                                              # autodiscover_test.go:336 -> InitializeScenario.func4
    Then "filebeat" collects events with "kubernetes.container.name:cronjob-container" # autodiscover_test.go:146 -> *podsManager

```


e2e/_suites/kubernetes-autodiscover/features/beats.feature
```diff
@@ -0,0 +1,52 @@
+@beats
+Feature: Beats
+  Use Kubernetes autodiscover features in Beats to monitor pods
+
+Scenario: Pod is started
+  Given a cluster is available
+    And configuration for "filebeat" has "hints enabled"
+    And "filebeat" is running
```
mdelapenya, line 8: wdyt about having one single step for that? Having multiple ANDs is a smell that we could be "scripting". Ideally zero or one AND is acceptable, although if properly discussed it can be accepted

```suggestion
    And "filebeat" is running with a configuration that has "hints enabled"
```

e2e/_suites/kubernetes-autodiscover/features/beats.feature
```diff
@@ -0,0 +1,52 @@
+@beats
+Feature: Beats
+  Use Kubernetes autodiscover features in Beats to monitor pods
+
+Scenario: Pod is started
+  Given a cluster is available
```
mdelapenya, line 6: We could remove the `Given` clause and set it as part of the setUp, performed in the beforeScenario test hook. If it does not matter to create/destroy the cluster among scenarios, we could use the beforeSuite hook

mdelapenya: Mmm I see, this is only a check that the cluster is up, right? Then I'm OK with this one

ChrsMark:
cc-ing @mtojek  who might be interested into this too

cli/shell/shell.go
```diff
@@ -17,7 +18,7 @@ import (
 )
 
 // CheckInstalledSoftware checks that the required software is present
-func CheckInstalledSoftware(binaries []string) {
+func CheckInstalledSoftware(binaries ...string) {
```
mdelapenya, line 21: 👍

e2e/_suites/kubernetes-autodiscover/kubernetes_test.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func (c *kubernetesCluster) initialize(ctx context.Context) error {
+	err := c.isAvailable(ctx)
+	if err == nil {
+		return nil
+	}
+
+	log.Info("Kubernetes cluster not available, will start one using kind")
+	shell.CheckInstalledSoftware("kind")
```
mdelapenya, line 121: You could do this just once, in the setup of the test suite: https://github.com/elastic/e2e-testing/blob/master/e2e/_suites/helm/helm_charts_test.go#L661

e2e/_suites/kubernetes-autodiscover/kubernetes_test.go
```diff
@@ -0,0 +1,160 @@
...
+		_, err := c.Run(ctx, "get", "serviceaccount", "default")
+		if err == nil {
+			return nil
+		}
+		select {
+		case <-timeoutCtx.Done():
+			return fmt.Errorf("namespace was created but it is not ready after %s: %w", timeout, err)
+		default:
+		}
+		time.Sleep(time.Second)
```
mdelapenya, line 67: We have functions to waitFor using backoff. See https://github.com/elastic/e2e-testing/search?q=GetExponentialBackOff

e2e/_suites/kubernetes-autodiscover/features/beats.feature
mdelapenya: I think we could simply remove the `Given` clause and assert it in the `beforeSuite` or `beforeScenario` hook. From a high level perspective, I would expect the cluster to be available as a strong requirement to run the tests. We usually create and start runtime (test-time) dependencies in the proper `beforeSuite` hook, exiting if they cannot be satisfied.

e2e/_suites/kubernetes-autodiscover/features/beats.feature
```diff
@@ -0,0 +1,52 @@
...
+    And configuration for "filebeat" has "hints enabled"
+    And "filebeat" is running
+   When "a pod" is deployed
+   Then "filebeat" collects events with "kubernetes.pod.name:a-pod"
+
+Scenario: Pod is deleted
+  Given a cluster is available
+    And configuration for "filebeat" has "hints enabled"
+    And "filebeat" is running
```
mdelapenya, line 15: Ditto:
```suggestion
    And "filebeat" is running with a configuration that has "hints enabled"
```

e2e/_suites/kubernetes-autodiscover/features/beats.feature
```diff
@@ -0,0 +1,52 @@
...
+    And "a pod" is deployed
+    And "filebeat" collects events with "kubernetes.pod.name:a-pod"
+   When "a pod" is deleted
+   Then "filebeat" stops collecting events
+
+Scenario: Pod is failing
+  Given a cluster is available
+    And configuration for "filebeat" has "hints enabled"
+    And "filebeat" is running
```
mdelapenya, line 24: Ditto:
```suggestion
    And "filebeat" is running with a configuration that has "hints enabled"
```

e2e/_suites/kubernetes-autodiscover/features/beats.feature
```diff
@@ -0,0 +1,52 @@
...
+   When "a failing pod" is deployed
+   Then "filebeat" collects events with "kubernetes.pod.name:a-failing-pod"
+
+# This scenario explicitly waits for 60 seconds before doing checks
+# to be sure that at least one job has been executed.
+Scenario: Short-living cronjob
+  Given a cluster is available
+    And configuration for "filebeat" has "hints enabled"
+    And "filebeat" is running
```
mdelapenya, line 33: Ditto:
```suggestion
    And "filebeat" is running with a configuration that has "hints enabled"
```

e2e/_suites/kubernetes-autodiscover/features/beats.feature
```diff
@@ -0,0 +1,52 @@
...
+   When "a short-living cronjob" is deployed
+    And "60s" have passed
+   Then "filebeat" collects events with "kubernetes.container.name:cronjob-container"
+
+Scenario: Metrics hints with named ports
+  Given a cluster is available
+    And configuration for "metricbeat" has "hints enabled"
+    And configuration for "a pod" has "metrics annotations with named port"
+    And "metricbeat" is running
```
mdelapenya, line 42:
```suggestion
    And "metricbeat" is running with a configuration that has "hints enabled"
    And configuration for "a pod" has "metrics annotations with named port"
```

e2e/_suites/kubernetes-autodiscover/features/beats.feature
```diff
@@ -0,0 +1,52 @@
...
+   When "a pod" is deployed
+   Then "metricbeat" collects events with "kubernetes.pod.name:a-pod"
+
+Scenario: Monitor hints with named ports
+  Given a cluster is available
+    And configuration for "heartbeat" has "hints enabled"
+    And configuration for "a service" has "monitor annotations with named port"
+    And "heartbeat" is running
```
mdelapenya, line 50:
```suggestion
    And "heartbeat" is running with a configuration that has "hints enabled"
    And configuration for "a service" has "monitor annotations with named port"
```

e2e/_suites/kubernetes-autodiscover/features/beats.feature
jsoriano: This is a good point, I added it at the beginning to setup the cluster but it actually doesn't contribute anything to the scenario, at least by now, scenario is setup in the `BeforeScenario` hook as you propose. I will remove this step.

e2e/_suites/kubernetes-autodiscover/features/beats.feature
```diff
@@ -0,0 +1,52 @@
+@beats
```
mdelapenya, line 1: Let's add a tag that selects the entire test suite directly. I would rename the base dir to `beats-kubernetes`, acting as a namespace for any kubernetes related tests.
```suggestion
@autodiscover
```

mdelapenya:
This PR is great @jsoriano! Thanks for the contribution 👏 

I added a few comments related to Gherkin style, in order to standardise the creation of the scenarios. I did not change the step functions, so please take a look at them when reviewing my comments for the feature file.

I only miss adding the tests to the CI. You can add the new stage here: https://github.com/elastic/e2e-testing/blob/master/.ci/.e2e-tests.yaml, adding a new suite entry (including changes in https://github.com/elastic/e2e-testing/pull/1064/files#r617639126):

```yaml
  - suite: "beats-kubernetes"
    platforms:
      - "ubuntu-18.04"
    scenarios:
      - name: "k8s autodiscover"
        tags: "autodiscover"
```

In a follow up PR, once this is merged, I'll add APM spans and traces for the suite, as we are capturing that data for other test suites.

mdelapenya:
@adam-stokes I see code here that could be extracted to the new project layout in #1008. I.e. the kubernetes controller

e2e/_suites/kubernetes-autodiscover/features/beats.feature
jsoriano: I am on the fence here, but I will let you advise me :slightly_smiling_face: 

There are two main reasons why I did it this way:
* I expect different ways of deploying resources, this is why there are different "is running" and "is deployed" steps, the first one would be for pods that I would expect to be running before continuing with the next step, and the second one for resources in general (including pods that are never going to reach the running state, as failing or short living ones).
* I expect the configurations for some of these resources to have any number of "options" to chose different configuration blocks.

The alternatives I saw for this would be to have more specific steps or more specific templates.

More specific steps would be similar to what you propose, there would be steps such as:
* `"a pod" is running`
* `"a pod" is deployed`
* `"a pod" is running with a configuration that has "hints for metrics"`
* `"a pod" is deployed with a configuration that has "hints for metrics"`

We would need to implement each one of the different combinations, and we would be limited to zero or one options. If we want to include templates with more than one variation we would need to implement some additional step, or make the templates aware of more complex options. But I am not sure if we need that, so maybe this is not such a problem.

Maybe another option is to make the desired state a parameter, so we only have two steps that would allow to use one or none options:

* `"a pod" is "running"`
* `"a pod" is "deployed"`
* `"a pod" is "running" with a configuration that has "hints for metrics"`
* `"a pod" is "deployed" with a configuration that has "hints for metrics"`

One extra point in favour of that is that we don't need to keep state as I am keeping now in the `configurations` map. But this state is also used in the deletion steps, we would need to ensure that for the same pod we are using the same options in the run/deploy and delete steps.

With more specific templates we would have:
* `"a pod" is running`
* `"a pod" is deployed`
* `"a pod with hints for metrics" is running`
* `"a pod with hints for metrics" is deployed`

That would require to have many more templates.

e2e/_suites/kubernetes-autodiscover/features/beats.feature
jsoriano: Ok, I will review the organization of this. In principle I only wanted to test kubernetes autodiscover in this suite (and not other kubernetes features as metricbeat modules), with beats by now, but possibly with agent too in the near future.

Regarding the feature files, if I expect to have different scenarios for metricbeat, filebeat, hearbeat and agent, would they better go in the same feature file, or would it be prefered to have a different feature file and different tags for each one of them?

e2e/_suites/kubernetes-autodiscover/features/beats.feature
mdelapenya: Ok, now I get your point. What do you think about having more scenarios, with just the required steps: one `When` and one `Then`, and probably one `And` with each of them?

e2e/_suites/kubernetes-autodiscover/features/beats.feature
mdelapenya: I'd separate them by feature file, unless they are totally related. OTOH, we can start simple with one feature file, and then split it if/when needed

e2e/_suites/kubernetes-autodiscover/features/beats.feature
jsoriano: I have implemented your proposal at the end, reusing most of the underlying code and removing this `configurations` state. It actually simplifies things. Let me know what you think: 8ac0978bfb776c3f71f82af83978c6d6f0f79c37

e2e/_suites/kubernetes-autodiscover/features/beats.feature
```diff
@@ -0,0 +1,38 @@
...
+
+Scenario: Pod is started
+  Given "filebeat" is running with "hints enabled"
+   When "a pod" is deployed
+   Then "filebeat" collects events with "kubernetes.pod.name:a-pod"
+
+Scenario: Pod is deleted
+  Given "filebeat" is running with "hints enabled"
+    And "a pod" is deployed
+    And "filebeat" collects events with "kubernetes.pod.name:a-pod"
```
mdelapenya, line 13: Does the order of these steps change the expected behaviour?

e2e/_suites/kubernetes-autodiscover/kubernetes_test.go
jsoriano: I only wanted to make this check if kind is going to be used. It is only used if there is no configured cluster.

e2e/_suites/kubernetes-autodiscover/features/beats.feature
mdelapenya: The first one for what happens to filebeat if it starts collecting events before a pod is deployed, and the second one, exactly as you proposed.

jsoriano: Yes, order matters here, I want to be sure that the beat was collecting events before the pod was deleted, and that after deleting it, the beat stops collecting events.

(While writing this I see this scenario doesn't make any sense for filebeat, a deleted pod is not going to generate logs that could be collected by filebeat. But it makes sense for metricbeat that will generate events with errors if it tries to continue monitoring a deleted pod).

adam-stokes:
@jsoriano Thanks for the PR, just a heads up we merged a refactor in that moved a lot of the modules into the `internal` directory. A few of your references to `cli/shell/shell.go` will need to be updated. I'd also like to extract out any general code reuse into the `internal` directory as well where it makes sense.

Thanks again for the contribution!

ChrsMark:
@jsoriano thank you for adding this, it will save us a lot of time! 
Since I see that you use `kubectl` mainly I wonder if you considered using go-client or sth similar to interact with the resources. For instance at https://github.com/elastic/elastic-package/pull/323 Marcin used a helm library to make sure that resources are ready before starting the test.

adam-stokes:
> @jsoriano thank you for adding this, it will save us a lot of time!
> Since I see that you use `kubectl` mainly I wonder if you considered using go-client or sth similar to interact with the resources. For instance at [elastic/elastic-package#323](https://github.com/elastic/elastic-package/pull/323) Marcin used a helm library to make sure that resources are ready before starting the test.

Hey @ChrsMark ! I actually used bits from elastic-package when I did my refactor, do you think it would be worth extracting this code out of elastic-package internal into a `/pkg`? I'd rather just link to the code rather than copy and paste it :) If you look in this repo under `internal/kibana` you'll see a lot of the same code used there as well. But having a shared library of sorts would be awesome. @mtojek wdyt?

mtojek:
I can tell you, @adam-stokes , that there was already a discussion around this issue :) See: https://github.com/elastic/integrations/issues/800 . I think it just waits for the right moment. Unfortunately we didn't have time to look into this yet, we need more hands on board to proceed with client extraction.

e2e/_suites/kubernetes-autodiscover/kubernetes_test.go
jsoriano: For loops refactored to use these backoffs: 1cee53297409c5634259727f80ac621144d1d5c9

It looks a bit less go native to me, but the code of the operations look cleaner now, so let's go with it. Thanks for the tip.

e2e/_suites/kubernetes-autodiscover/features/beats.feature
jsoriano: Separated by feature files: 6954b9c26136fa492d0d87eda4d6943ddcf878a9, one for each beat. I have also rephrased the titles of the scenarios to tell more about what it tests, and not about how it does it.

jsoriano:
>  Since I see that you use `kubectl` mainly I wonder if you considered using go-client or sth similar to interact with the resources. For instance at [elastic/elastic-package#323](https://github.com/elastic/elastic-package/pull/323) Marcin used a helm library to make sure that resources are ready before starting the test.

The truth is that I didn't consider so much to use go-client or any other library for this. The reason was that I was trying to automate more or less the steps that I follow to manually test these cases, and for this I use kind and kubectl (I am only replacing vim + commented-out lines with templates with "option blocks" :slightly_smiling_face: ). kind and kubectl are already available in the CI builds of this repo, so it was straight forward. Other point in favour of kubectl for this is that what I needed more here was `kubectl apply/delete`, and this looks less convenient to me with go-client.

I am open in any case to reimplement this in the future to use `go-client` or any other library, or to reuse the kubectl helpers that are already in this repo and I am not using. But I think we can go on with this by now.

And +1 too to have a common library for this kind of use cases, when we have time for this.

Also `kubectl cp` was pretty handy to get the events from the output file, and this may be more complicated with `go-client` (or it also has helpers for that?) Though I see this is a bit of a hack (that works!) and may need to be reimplemented to use a centralized output as ES or redis if we want to add multi-node scenarios at some moment.

e2e/_suites/kubernetes-autodiscover/testdata/kind.yml
```diff
@@ -0,0 +1,4 @@
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+featureGates:
```
mdelapenya, line 3: >[2021-04-22T18:44:05.639Z] time="2021-04-22T18:44:04Z" level=error msg="Error executing command" args="[create cluster --name kind-590267e1-989e-49dd-ab45-e97edaf4a313 --config testdata/kind.yml --kubeconfig /tmp/test-335526962/kubeconfig]" baseDir=. command=kind error="exit status 1" stderr="Creating cluster \"kind-590267e1-989e-49dd-ab45-e97edaf4a313\" ...\nERROR: failed to create cluster: unable to decode config: yaml: unmarshal errors:\n  line 3: field featureGates not found in type v1alpha4.Cluster\n"

This error is raised on CI

jsoriano: Umm, this works for me:tm: with kind 0.10.0, what version of kind is used in CI?

e2e/_suites/kubernetes-autodiscover/autodiscover_test.go
```diff
@@ -0,0 +1,437 @@
...
+					return true
+				}
+			}
+			return false
+		},
+		"beats_namespace": func() string {
+			return utils.GetDockerNamespaceEnvVar("beats")
+		},
+		"beats_version": func() string {
+			return shell.GetEnv("GITHUB_CHECK_SHA1", shell.GetEnv("BEAT_VERSION", defaultBeatVersion))
```
jsoriano, line 52: I added this in https://github.com/elastic/e2e-testing/pull/1064/commits/dcd98735d70ec77a7194ac459c4b30bfde2a878c to support using the versions from beats PRs, would this be enough? I guess docker inside kind needs to be authenticated too? :thinking:

jsoriano: @mdelapenya sorry, I wanted to ping you in this comment ^ :slightly_smiling_face:

e2e/_suites/kubernetes-autodiscover/testdata/kind.yml
mdelapenya: https://github.com/elastic/e2e-testing/blob/249fd97a14b97fe7e51731651b2335b57df1da33/.ci/Jenkinsfile#L53

0.10.0

jsoriano: Ok, it seems that each suite has its own script to install dependencies, I was not providing one and the machine had kind 0.7.0 installed. Added a script to install the expected versions of kind and kubectl.

.ci/Jenkinsfile
```diff
@@ -52,6 +52,8 @@ pipeline {
     string(name: 'HELM_VERSION', defaultValue: '3.5.2', description: 'SemVer version of Helm to be used.')
     string(name: 'HELM_KIND_VERSION', defaultValue: '0.10.0', description: 'SemVer version of Kind to be used.')
     string(name: 'HELM_KUBERNETES_VERSION', defaultValue: '1.18.2', description: 'SemVer version of Kubernetes to be used.')
+    string(name: 'KUBERNETES_AUTODISCOVER_KIND_VERSION', defaultValue: '0.10.0', description: 'SemVer version of Kind to be used.')
```
mdelapenya, line 55: wdyt about reusing existing Jenkins params? We could remove those two new lines, and in your script, use the values.

If you see it convenient, you could rename the existing Jenkins params & env vars (https://github.com/elastic/e2e-testing/pull/1064/commits/6f8cd442840d28acd5b942e720efa5bc71555ad7#diff-ef8b2f5e4d21d81c0eba4754dea3a3dfc2c67d26284f195165b650cf4aff7beaR79-R81) with a reusable namespaced name, something like `KIND_VERSION`. wdyt?

jsoriano: Done in a64429347c8317b4b616f51f09ad51b7d6a53c67

Are `HELM_*` variables used somewhere else to invoke these jobs? Should I keep them for backwards compatibility?

jsoriano:
Added some docs and opening for review.

Some pending things would be to confirm that these jobs work with snapshot images created by the packaging job of beats PRs, and to share/reuse more code related to the kubernetes setups. I am ok with doing these changes before merging, but I also think that they can be done later. In any case I won't have much time to dedicate to this in the following 2-3 weeks.

For the integration with beats CI jobs, I guess that it needs to be explicitly enabled there, so we can continue with this if we find issues when enabling this integration.

adam-stokes (approved):
Super cool 🥳 Thanks for the contribution!

e2e/_suites/kubernetes-autodiscover/kubernetes_test.go
```diff
@@ -135,11 +135,17 @@ func (c *kubernetesCluster) initialize(ctx context.Context) error {
...
 
-	output, err := shell.Execute(ctx, ".", "kind", "create", "cluster",
+	args := []string{
+		"create", "cluster",
 		"--name", name,
 		"--config", "testdata/kind.yml",
 		"--kubeconfig", c.kubeconfig,
-	)
+	}
+	if version, ok := os.LookupEnv("KUBERNETES_VERSION"); ok && version != "" {
```
mdelapenya, line 144: I guess this is needed to install the k8s version no matter whether the worker has it or not, right? Does it makes sense to install it instead of failing?

e2e/_suites/kubernetes-autodiscover/README.md
```diff
@@ -0,0 +1,174 @@
...
+used to provide a local cluster.
+
+The tests will follow this general high-level approach:
+
+1. It uses `kubectl` to interact with a Kubernetes cluster.
+1. If there is no configured Kubernetes cluster in kubectl, a new one
+   is deployed using `kind`. If a cluster is created, it is also removed after
+   the suite is executed.
+1. Execute BDD steps representing each scenario. Each scenario is executed in a
+   different Kubernetes namespace, so it is easier to cleanup and avoid one
```
mdelapenya, line 32: Great seeing each test suite namespaced!

e2e/_suites/kubernetes-autodiscover/README.md
```diff
@@ -0,0 +1,174 @@
...
+Several of the available steps can be parameterized with template names, these
+names can be written as the name of the template without the extension. Spaces
+in these names are replaced with hyphens.
+
+There are steps intended to define a desired state for the resources in the
+template, such as the following ones:
+* `"filebeat" is running` deploys the template `filebeat.yml.tmpl` and waits for
+  filebeat pods to be running. This step expects some pod to be labeled with
+  `k8s-app:filebeat`.
+* `"a service" is deployed deploys the resources in the template
```
mdelapenya:
```suggestion
* `"a service" is deployed` deploys the resources in the template
```

e2e/_suites/kubernetes-autodiscover/kubernetes_test.go
```diff
@@ -0,0 +1,174 @@
...
+			log.Warnf("Failed to delete kind cluster %s", c.kindName)
+		}
+		c.kindName = ""
+	}
+	if c.tmpDir != "" {
+		err := os.RemoveAll(c.tmpDir)
+		if err != nil {
+			log.Warnf("Failed to remove temporary directory %s", c.tmpDir)
+		}
+	}
```
mdelapenya, line 173: Wdyt about adding this info to logs? reading Jenkins logs I cannot see if the test suite was finished (OTOH they scenarios are present in the Tests UI)
```suggestion
	}
	log.Infof("kind cluster %s was deleted", c.kindName)
```

e2e/_suites/kubernetes-autodiscover/features/heartbeat.feature
```diff
@@ -0,0 +1,25 @@
...
+   When "redis" is deployed with "monitor annotations with named port"
+   Then "heartbeat" collects events with "kubernetes.pod.name:redis"
+    And "heartbeat" collects events with "monitor.status:up"
+
+Scenario: Monitor service availability using hints
+  Given "heartbeat" is running with "hints enabled for services"
+   When "redis service" is deployed with "monitor annotations"
+   Then "heartbeat" collects events with "kubernetes.service.name:redis"
+    And "heartbeat" collects events with "monitor.status:down"
+   When "redis" is running
```
mdelapenya, line 18: Is it possible to split this into two separate scenarios? It's not common to have a `When` after a `Then`. It makes readers to think in a scripted (complex) scenario rather than in a straight forward one.

For me this is the only blocker. Other than that, the PR is GOLD 🥇

e2e/_suites/kubernetes-autodiscover/kubernetes_test.go
jsoriano: Each version of kind uses a version of kubernetes by default, it is ok not to set this option. With this option we are adding support for selecting a different version. I can rephrase of remove the log message below if it is confusing.

e2e/_suites/kubernetes-autodiscover/kubernetes_test.go
```diff
@@ -0,0 +1,174 @@
+package main
```
mdelapenya, line 1: I think you can remove the `_test` from this file, as it's not a test by itself but a library used in tests. Does it make sense from Go native standpoint?

e2e/_suites/kubernetes-autodiscover/kubernetes_test.go
mdelapenya: I'm OK with enriching the log explaining the purpose :)

.ci/.e2e-tests.yaml
```diff
@@ -62,3 +62,9 @@ SUITES:
         tags: "integrations && redisenterprise"
       - name: "vSphere"
         tags: "integrations && vsphere"
+  - suite: "kubernetes-autodiscover"
+    platforms:
+      - "ubuntu-18.04"
+    scenarios:
+      - name: "kubernetes autodiscover with beats"
+        tags: "kubernetes && autodiscover && beats"
```
mdelapenya, line 70: As a suggestion trying to speed up the build, you can run the 3 feature files in parallel in three different workers. Simply explicitly tag each feature file with its own annotation (i.e. `@filebeat`, `@heartbeat`, `@metricbeat`), and create a new `- suite` section in this YAML file.

e2e/_suites/kubernetes-autodiscover/features/heartbeat.feature
```diff
@@ -13,12 +13,19 @@ Scenario: Monitor pod availability using hints with named ports
...
     And "heartbeat" collects events with "monitor.status:up"
     And "heartbeat" does not collect events with "monitor.status:down" during "20s"
 
+# A service without backend pods should be reported as down.
+Scenario: Monitor service unavailability using hints
+  Given "heartbeat" is running with "hints enabled for services"
+   When "redis service" is deployed with "monitor annotations"
+   Then "heartbeat" collects events with "kubernetes.service.name:redis"
+    And "heartbeat" collects events with "monitor.status:down"
+    And "heartbeat" does not collect events with "monitor.status:up" during "20s"
```
mdelapenya, line 27: Thanks for this change!

I have another question: are those three conditions (Then > And > And) required all together? Or can they be split into different scenarios? It could be the case where the first condition fails, and then we won't receive feedback about the other two.

e2e/_suites/kubernetes-autodiscover/kubernetes_test.go
jsoriano: I actually added `_test` to both files because I thought it was the convenience with godog :slightly_smiling_face: In principle code that is only used by tests can live in `_test` files, but no strong opinion here. Suffix removed.

.ci/.e2e-tests.yaml
jsoriano: Done.

e2e/_suites/kubernetes-autodiscover/features/heartbeat.feature
jsoriano: Umm, I would prefer to leave them together, for the same scenario we expect all the conditions to happen.

It would also work if they were in three different scenarios, but it would look a bit weird to me to have three identical scenarios where only the expectations change.

Also, maybe we want to add support for more complex conditions, these checks would be more accurate if we could do this:
```
   Then "heartbeat" collects events with "kubernetes.service.name:redis and monitor.status:down"
    And "heartbeat" does not collect events with "kubernetes.service.name:redis and monitor.status:up" during "20s"
```

mdelapenya: Ok, let's keep it like this and we can refactor that in a follow-up PR.

As a reference, this is possible:
```gherkin
Scenario: Monitor service unavailability using hints
  Given "heartbeat" is running with "hints enabled for services"
   When "redis service" is deployed with "monitor annotations"
   Then "heartbeat" collects events with
   | key    | value    |
   | config | <config> |
    And "heartbeat" does not collect events with "monitor.status:up" during "20s"
Examples:
    | config                        | 
    | kubernetes.service.name:redis |
    | monitor.status:down           |
```
but it would require changing all steps using that method to accept a data table:

```go
func collectsEventsWith(config string, table *messages.PickleStepArgument_PickleTable) error {
	return godog.ErrPending
}
```

mdelapenya (approved):
Great job! LGTM 👍 

image

.ci/.e2e-tests.yaml
```diff
@@ -66,5 +66,9 @@ SUITES:
     platforms:
       - "ubuntu-18.04"
     scenarios:
-      - name: "kubernetes autodiscover with beats"
-        tags: "kubernetes && autodiscover && beats"
+      - name: "kubernetes autodiscover with filebeat"
```
mdelapenya, line 69: This change passed from 16 minutes in one worker to 9 in three workers

adam-stokes:
@mdelapenya It's mentioning gherkinci lint is failing here but I can't figure out what is failing. Other than that this PR is ready to be merged

mdelapenya:
> @mdelapenya It's mentioning gherkinci lint is failing here but I can't figure out what is failing. Other than that this PR is ready to be merged

Yep, it's possible to reproduce the errors with `make -C e2e lint` (name is misleading, we should change to `lint-gherkin`)

jsoriano:
Linting hopefully fixed.

Some rules seem a bit counter-intuitive to me, such as limiting the name of the feature, forcing the use of "Background" if there are common "Givens" in a feature, or allowing a single "When", what forces to move things to the "Given", or the "Then". But probably I need to work a little more with Gherkin to think more this way :slightly_smiling_face:

e2e/_suites/kubernetes-autodiscover/features/filebeat.feature
```diff
@@ -1,38 +1,35 @@
...
    Then "filebeat" collects events with "kubernetes.pod.name:a-failing-pod"
 
 # This scenario explicitly waits for 60 seconds before doing checks
 # to be sure that at least one job has been executed.
 Scenario: Logs collection from short-living cronjobs
-  Given "filebeat" is running with "hints enabled"
-   When "a short-living cronjob" is deployed
-    And "60s" have passed
+  Given "a short-living cronjob" is deployed
+   When "60s" have passed
```
mdelapenya, line 21: I'd keep it like this, because the `When` clause is considered the real test, being the `Given` the requirements/preparation for the scenario. The `Given` is optional, `When`, mandatory.

```suggestion
  When "a short-living cronjob" is deployed
   And "60s" have passed
```

e2e/_suites/kubernetes-autodiscover/features/filebeat.feature
```diff
@@ -1,38 +1,35 @@
-@kubernetes
-@autodiscover
+@kubernetes-autodiscover
 @filebeat
-Feature: Log collection from Kubernetes pods.
+Feature: Filebeat
   Use Kubernetes autodiscover features in Filebeat to collect logs.
 
-Scenario: Logs collection from running pod
+Background:
```
mdelapenya, line 6: About the background, I prefer having scenarios being self-content, with no need to check background clauses at the top of the file, although in this particular case the feature file is small.

In the case the feature file grows, we can skip the Background lint rule, but let's keep it as is by the moment

e2e/_suites/kubernetes-autodiscover/features/filebeat.feature
jsoriano: I agree with this, this is what I had it before, but `gherkin-lint` doesn't think the same:
```
AvoidScripting - Multiple Actions
  e2e/_suites/kubernetes-autodiscover/features/filebeat.feature (19): Filebeat.Logs collection from short-living cronjobs
```

e2e/_suites/kubernetes-autodiscover/features/filebeat.feature
jsoriano: Same thing, I would also prefer not to use `Background`, but the linter asks for it:
```
UseBackground - Step 'Given "filebeat" is running with "hints enabled"' should be part of background
  e2e/_suites/kubernetes-autodiscover/features/filebeat.feature (4): Log collection from Kubernetes pods.
```

jsoriano:
Would it be possible/make sense to disable `AvoidScripting` and `UseBackground` in the linter if we don't agree with it?

mdelapenya:
> Would it be possible/make sense to disable `AvoidScripting` and `UseBackground` in the linter if we don't agree with it?

I agree in the `UseBackground` one, although I think the `UseScripting` rule helps scenario writers to focus in more straightforward scenarios

e2e/_suites/kubernetes-autodiscover/features/filebeat.feature
jsoriano: Let's go back to what the linter wants, I don't see it so bad and this way we don't need to add an exception.

jsoriano:
> > Would it be possible/make sense to disable `AvoidScripting` and `UseBackground` in the linter if we don't agree with it?
> 
> I agree in the `UseBackground` one, although I think the `UseScripting` rule helps scenario writers to focus in more straightforward scenarios

Ok, I have added `UseBacground` to the list of exceptions and reverted the related changes. I have also fixed the `UseScripting` case. I would prefer to have the `And` under the `When`, but not so strongly to need to add an exception to the linter.

mdelapenya:
I'd say that the lint is telling us that the `60s have passed` sounds weird from a specification stand point. Is it required to wait a number of secs to collect events?

> I know we have something similar in Metricbeat, which we have considered to replace for long time.

And about the number, why 60, or 40? Should we update that number in slower/faster machines? It seems to me that this is an implementation detail that must be hidden from spec. You could use the `TIMEOUT_FACTOR` variable to define this waiting period internally, representing a multiplier of 3 to a default timeout you set in your codebase.

Nevertheless, if the waiting period is a strong requirement, wdyt about this refactor for that block:

```gherkin
  When "a short-living cronjob" is deployed
   And "60s" have passed
```
to
```gherkin
  When "a short-living cronjob" is deployed with "60s" of waiting
```

mdelapenya:
> Ok, I have added `UseBackground` to the list of exceptions and reverted the related changes.

The rule must be excluded at the pre-commit level: https://github.com/elastic/e2e-testing/blob/master/.pre-commit-config.yaml#L48

The one you updated is a helper for development purpose to easily reproduce what pre-commit does.

jsoriano:
> I'd say that the lint is telling us that the `60s have passed` sounds weird from a specification stand point. Is it required to wait a number of secs to collect events?

Yes, we have to wait 60s to avoid flakiness. We are checking here that we are collecting logs from a cronjob, the minimum scheduling available here is one job per minute, and Kubernetes only guarantees that one job is executed every 60 seconds, so since the job is configured till it is executed the first time can effectively pass up to 60 seconds.

One alternative I considered was to use the "is running" version of the rule, but the problem that we needed to handle in beats is that short living pods are a bit special and never reach the running state, they can go directly from waiting/pending to succeed/deleted or even to be deleted. We handle these things in Beats, and I wouldn't want to copy this logic into the tests too, so I thought that it would be better to model what a user would expect: the job has to run every 60 seconds, so after this time I should have the logs from the first pod.
In this case this wait models what the user would do. This is also why I think that having `When "60s" have passed` maybe not so bad.

I also considered to don't declare the wait explicitly, the next step `"filebeat" collects events with "kubernetes.container.name:cronjob-container"` has a bigger timeout, so the tests currently pass the same. But if at some moment we decide to reduce the timeout of the `collects events`, we may introduce unrelated flakiness.

> And about the number, why 60, or 40? Should we update that number in slower/faster machines?

It is not related to the performance of the machine, but about the kubernetes controller for cronjobs.

> Nevertheless, if the waiting period is a strong requirement, wdyt about this refactor for that block:
> 
> ```
>   When "a short-living cronjob" is deployed
>    And "60s" have passed
> ```
> 
> to
> 
> ```
>   When "a short-living cronjob" is deployed with "60s" of waiting
> ```

Yep, I also considered to have specific steps for cronjobs, but all of them sounded to me as "the cronjob is deployed **and** _the first period_ has passed", or "the cronjob is deployed **and** it has executed at least once", so it sounded to me that better do it as two separate steps.
This may be a matter of style, so if you think that it'd be better to have a single more specific step, I can do it.

e2e/_suites/kubernetes-autodiscover/features/filebeat.feature
```diff
@@ -0,0 +1,37 @@
...
+   When "a pod" is deployed
+   Then "filebeat" collects events with "kubernetes.pod.name:a-pod"
+
+Scenario: Logs collection from failing pod
+  Given "filebeat" is running with "hints enabled"
+   When "a failing pod" is deployed
+   Then "filebeat" collects events with "kubernetes.pod.name:a-failing-pod"
+
+# This scenario explicitly waits for 60 seconds before doing checks
+# to be sure that at least one job has been executed.
```
jsoriano, line 17: Btw, @mdelapenya, I had added this comment explaining why there is a explicit wait here, because it also looked funny to me. Is this the best way to comment this? Does gherkin support comments at the step level?

mdelapenya (approved):
Thank you for your patience in this review @jsoriano. I like it a lot how the new test suite is specified, being the scenarios and steps very definite.

Good job!

mdelapenya:
@jsoriano please check what branches should we backport this code. Please add the `v7.X.0` labels and mergify will do the rest (You can take a look at mergify config file for rules)

jsoriano:
Thanks for the reviews and suggestions! This has been a great exercise :slightly_smiling_face:

PR #1067, opened 2021-04-21
fix: run unit tests after refactor

mdelapenya:
## What does this PR do?
It moves test resources from the original location before the refactor to the new location under the internal (we could use the root dir too. wdyt?)

Besides that it updates the automation to run the unit tests, both locally with Make and in the CI.

## Why is it important?
We lost the unit tests (see https://beats-ci.elastic.co/blue/organizations/jenkins/e2e-tests%2Fe2e-testing-mbp/detail/PR-1008/44/pipeline/106)

## Checklist

- [x] I have run the Unit tests for the CLI/Internal, and they are passing locally

```shell
make unit-test
```

## Related issues

- Caused by #1008


adam-stokes:
> It moves test resources from the original location before the refactor to the new location under the internal (we could use the root dir too. wdyt?)

I think this makes sense, the _testresources is only useful for our internal testing of the code?

mdelapenya:
> I think this makes sense, the _testresources is only useful for our internal testing of the code?

Yes, it's only used in the `internal` package, although it could be eventually useful for unit test in the cli one. OTOH, we could be moving code from cli to internal

adam-stokes (approved):

PR #1078, opened 2021-04-22
feat: bootstrap fleet-server for the deployment of regular elastic-agents

mdelapenya:
## What does this PR do?
It converts existing deployments of regular agents to use a bootstrapped fleet-server. For that, we:

- cloned the centos/debian services into fleet-server centos/debian ones. This is needed to simplify the generation of namespaces, service names and container names.
- extracted common logic to calculate the container and service name. This method will internally use the installer struct to get the image and service from there. For context, the container name follows the `profile-image-service-index` pattern.
- turned back to use the step to deploy a regular agent instead of a fleet-server one (the latter bootstrapped fleet-server). This was added in the previous refactor to make test pass after recent kibana breaking changes when enrolling agents using fleet-server. This new step is always bootstrapping a fleet-server, waiting for it to be online first.
- added a reference to the fleet server hostname in the fleet struct, so that it's empty at the beginning of each scenario, and populated right after a bootstrap happens. With this approach, the deployment will create the Fleet configuration using that hostname, and depending on that will create the right flags for installing/enrolling the agent with an already bootstrapped fleet server (if the fleet-server hostname is not empty), or bootstrapping fleet-server (if the hostname is empty).
- fixed the flags used to communicate with fleet-server. As we bootstrap it with HTTPS, we need to use `https` in the connection URL.
- fixed how the enrollment token was created. For agents using an already bootstrapped fleet-server, the token must becreated from the default fleet policy, not the fleet server one.
- performed minor refactors, such as renaming a variable for consistency and readability

## Why is it important?
It adds fleet-server support, keeping up with current development status for Fleet project.

## Checklist

```shell
$ SUITE="fleet" TAGS="fleet_mode_agent && centos" TIMEOUT_FACTOR=3 LOG_LEVEL=TRACE BEATS_USE_CI_SNAPSHOT=false DEVELOPER_MODE=false make -C e2e functional-test
```

## Related issues

-


cachedout (approved):

e2e/_suites/fleet/stand-alone.go
```diff
@@ -87,7 +87,7 @@ func (sats *StandAloneTestSuite) theStandaloneAgentIsListedInFleetWithStatus(des
 	return nil
 }
 
-func (sats *StandAloneTestSuite) aStandaloneAgentIsDeployedWithFleetServerMode(image string) error {
+func (sats *StandAloneTestSuite) bootstrapFleetServerFromAStandaloneAgent(image string) error {
```
cachedout, line 90: Nit: This has a capitalization mismatch. In `StandAloneTestSuite` the word `Alone` is capitalized by in `bootstrapFleetServerFromAStandaloneAgent`, it is lowercase.

cachedout (approved):

e2e/_suites/fleet/stand-alone.go
mdelapenya: What is the correct spell for standalone? I think the wrong one is the `StandAloneTestSuite` right?

kuisathaverat (approved):

adam-stokes (approved):

mdelapenya:
@adam-stokes I'm tempted to remove the fleet-server feature file, and move the bootstrap of the fleet server agent to the `beforeSuite` hook, so that the fleet server is always running (only bootstrapped once). This will reduce the build time, as we do not have to bootstrap-and-wait for the server to be online. Thoughts?

e2e/_suites/fleet/fleet.go
```diff
@@ -350,16 +363,31 @@ func (fts *FleetTestSuite) anAgentIsDeployedToFleetWithInstallerAndFleetServer(i
 	return err
 }
 
+// getContainerName returns the current container name for the service:
+// we are using the Docker client instead of docker-compose because it does not support
+// returning the output of a command: it simply returns error level
```
adam-stokes, line 368: This comment doesn't apply to this function any longer

e2e/_suites/fleet/fleet.go
```diff
@@ -350,16 +363,31 @@ func (fts *FleetTestSuite) anAgentIsDeployedToFleetWithInstallerAndFleetServer(i
...
+func (fts *FleetTestSuite) getContainerName(i installer.ElasticAgentInstaller, index int) string {
+	return fmt.Sprintf("%s_%s_%s_%d", i.Profile, i.Image, common.ElasticAgentServiceName, 1)
```
adam-stokes, line 370: `index` is not used here, I think we want to replace `1` with `index`

e2e/_suites/fleet/fleet.go
```diff
@@ -1035,11 +1047,9 @@ func (fts *FleetTestSuite) anAttemptToEnrollANewAgentFails() error {
 
 	agentInstaller := fts.getInstaller()
 
-	profile := agentInstaller.Profile // name of the runtime dependencies compose file
-
-	containerName := fmt.Sprintf("%s_%s_%s_%d", profile, fts.Image+"-systemd", common.ElasticAgentServiceName, 2) // name of the new container
+	containerName := fts.getContainerName(agentInstaller, 2) // name of the new container
```
adam-stokes, line 1050: The index is not being utilized here so getContainerName is always returning `_1`

adam-stokes (changes requested):

adam-stokes:
> @adam-stokes I'm tempted to remove the fleet-server feature file, and move the bootstrap of the fleet server agent to the `beforeSuite` hook, so that the fleet server is always running (only bootstrapped once). This will reduce the build time, as we do not have to bootstrap-and-wait for the server to be online. Thoughts?

Yea I think this is a good idea +1

adam-stokes (approved):

e2e/_suites/fleet/fleet.go
```diff
@@ -367,7 +367,7 @@ func (fts *FleetTestSuite) anAgentIsDeployedToFleetWithInstallerAndFleetServer(i
 // we are using the Docker client instead of docker-compose because it does not support
 // returning the output of a command: it simply returns error level
 func (fts *FleetTestSuite) getContainerName(i installer.ElasticAgentInstaller, index int) string {
-	return fmt.Sprintf("%s_%s_%s_%d", i.Profile, i.Image, common.ElasticAgentServiceName, 1)
+	return fmt.Sprintf("%s_%s_%s_%d", i.Profile, i.Image, common.ElasticAgentServiceName, index)
```
mdelapenya, line 370: Good catch!

mdelapenya:
jenkins run the test please

PR #1084, opened 2021-04-23
chore: initialise stale agent only when used

mdelapenya:
## What does this PR do?
It removes the initialisation of the stale agent in the fleet test suite, as it's already checked in the specific step for the stale agent. We delayed fetching the stale version in #1016

## Why is it important?
It fails fast when the version is not present, and for the stale agent, which is only executed on nightlies, we don't want to fail fast.

## Checklist

## Related issues

- Caused by #1008
- Relates #1016


mdelapenya:
CI passing, merging to fix builds

PR #1097, opened 2021-04-27
fix: branch_specifier is needed

mdelapenya:
## What does this PR do?
It reverts dddba263159f3fa8b30bd7f973d7e75ead6ce22b, as the branch_specifier is required, although failing the first time the build is executed (params are not there).

We are changing the schedued trigger to daily

## Checklist

## Author's Checklist

- [ ] JJBB

## Related issues

- Caused by #1093

### DIFF
--- a/.ci/build-docker-images.groovy
+++ b/.ci/build-docker-images.groovy
@@ -25,7 +25,7 @@ pipeline {
     quietPeriod(10)
   }
   parameters {
-    string(name: 'branch_specifier', defaultValue: 'master', description: 'It would not be defined on the first build, see JENKINS-41929.')
+    string(name: 'BRANCH_SPECIFIER', defaultValue: 'master', description: 'It would not be defined on the first build, see JENKINS-41929.')
   }
   triggers {
     cron 'H H(4-5) * * 1-5'
@@ -35,7 +35,7 @@ pipeline {
       environment {
         // Parameters will be empty for the very first build, setting an environment variable
         // with the same name will workaround the issue. see JENKINS-41929
-        BRANCH_SPECIFIER = "${params?.branch_specifier}"
+        BRANCH_SPECIFIER = "${params?.BRANCH_SPECIFIER}"
       }
       steps {
         deleteDir()
--- a/.ci/build-docker-images.groovy
+++ b/.ci/build-docker-images.groovy
@@ -25,7 +25,7 @@ pipeline {
     quietPeriod(10)
   }
   triggers {
-    cron '@daily'
+    cron 'H H(4-5) * * 1-5'
   }
   stages {
     stage('Checkout') {
--- a/.ci/build-docker-images.groovy
+++ b/.ci/build-docker-images.groovy
@@ -25,7 +25,7 @@ pipeline {
     quietPeriod(10)
   }
   triggers {
-     cron '@daily'
+    cron '@daily'
   }
   stages {
     stage('Checkout') {
--- a/.ci/build-docker-images.groovy
+++ b/.ci/build-docker-images.groovy
@@ -24,15 +24,23 @@ pipeline {
     rateLimitBuilds(throttle: [count: 60, durationName: 'hour', userBoost: true])
     quietPeriod(10)
   }
+  parameters {
+    string(name: 'branch_specifier', defaultValue: 'master', description: 'It would not be defined on the first build, see JENKINS-41929.')
+  }
   triggers {
     cron 'H H(4-5) * * 1-5'
   }
   stages {
     stage('Checkout') {
+      environment {
+        // Parameters will be empty for the very first build, setting an environment variable
+        // with the same name will workaround the issue. see JENKINS-41929
+        BRANCH_SPECIFIER = "${params?.branch_specifier}"
+      }
       steps {
         deleteDir()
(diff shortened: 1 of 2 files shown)


.ci/build-docker-images.groovy
```diff
@@ -25,7 +25,7 @@ pipeline {
     quietPeriod(10)
   }
   triggers {
-    cron 'H H(0-5) * * 1-5'
+    cron '@daily'
```
kuisathaverat, line 28: do not use daily, it will run on weekends, and 24h after the last execution so the time is not fixed.

mdelapenya: Ok, I used what we added in the apm-pipeline-library for building images: https://github.com/elastic/apm-pipeline-library/blob/master/.ci/jobs/beats-docker-images-pipeline.yml#L27

What is the recommended cron trigger? Maybe weekly/monthly is enough

.ci/jobs/build-docker-images.yml
```diff
@@ -17,6 +22,6 @@
             credentials-id: f6c7695a-671e-4f4f-a331-acdce44ff9ba
             reference-repo: /var/lib/jenkins/.git-references/e2e-testing.git
             branches:
-              - master
+              - $branch_specifier
```
v1v, line 25: Do you need to change -> https://github.com/elastic/e2e-testing/blob/master/.ci/build-docker-images.groovy#L35?

OTOH, I don't think you need this parameters but https://github.com/elastic/apm-pipeline-library/blob/64546ee45b1c64b11d6138053401062d51c7e36f/resources/JenkinsfileTemplate.groovy#L160-L163

In other words, if the issue is related to the first build, then you need to set the parameters with a default value and create and env variable. This is the workaround with the 🐔 🥚 

-

.ci/jobs/build-docker-images.yml
```diff
@@ -17,6 +22,6 @@
...
     triggers:
-      - timed: 'H H(0-5) * * 1-5'
+      - timed: '@daily'
```
v1v, line 27: If it helps you can remove this and use the one in the pipeline, though the very first build might be triggered manually to apply the changes from the pipeline. Again another 🐔 🥚

.ci/jobs/build-docker-images.yml
mdelapenya: You mean removing the parameter from JJBB and adding the parameter block in the groovy file?

.ci/build-docker-images.groovy
kuisathaverat: 'H H(0-5) * * 1-5' Monday to Friday anytime between 00:00 AM to 05:00 AM this is fine, the key is to set a range (H(0-5)) of hours where the build can run, Jenkins will manage to not overload the instance in that range.

mdelapenya: Ok, this is what we use for the nightly scheduled pipelines. Will use what we have in there

cachedout (approved):

.ci/jobs/build-docker-images.yml
v1v: yes

.ci/build-docker-images.groovy
```diff
@@ -24,15 +24,23 @@ pipeline {
     rateLimitBuilds(throttle: [count: 60, durationName: 'hour', userBoost: true])
     quietPeriod(10)
   }
+  parameters {
+    string(name: 'branch_specifier', defaultValue: 'master', description: 'It would not be defined on the first build, see JENKINS-41929.')
+  }
   triggers {
-    cron 'H H(0-5) * * 1-5'
+    cron 'H H(4-5) * * 1-5'
```
mdelapenya, line 31: Use same cron than in the nightly builds

.ci/build-docker-images.groovy
```diff
@@ -24,15 +24,23 @@ pipeline {
...
   }
   stages {
     stage('Checkout') {
+      environment {
+        // Parameters will be empty for the very first build, setting an environment variable
+        // with the same name will workaround the issue. see JENKINS-41929
+        BRANCH_SPECIFIER = "${params?.branch_specifier}"
```
v1v, line 38:
```suggestion
        BRANCH_SPECIFIER = "${params?.BRANCH_SPECIFIER}"
```

.ci/build-docker-images.groovy
```diff
@@ -24,15 +24,23 @@ pipeline {
     rateLimitBuilds(throttle: [count: 60, durationName: 'hour', userBoost: true])
     quietPeriod(10)
   }
+  parameters {
+    string(name: 'branch_specifier', defaultValue: 'master', description: 'It would not be defined on the first build, see JENKINS-41929.')
```
v1v, line 28:
```suggestion
    string(name: 'BRANCH_SPECIFIER', defaultValue: 'master', description: 'It would not be defined on the first build, see JENKINS-41929.')
```

v1v (approved):
NIT: normalise the names

mdelapenya:
> NIT: normalise the names

Done 😊  I used lowercase for params and uppercase for env vars on purpose, as I thought there was a naming convenition. If it's not there, we can add it to the shared library docs.

PR #1110, opened 2021-04-28
Kubernetes Deployment

adam-stokes:

Adds ability to deploy Elasticsearch, Kibana, Fleet Server (bootstrapped) into a Kubernetes cluster

## What does this PR do?

Adds necessary manifests to deploy a full EKF stack with optional ingress support for local development and testing

## Why is it important?

Would like to utilize this deployment tactic to run our e2e test suites against local and remote kubernetes clusters

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,6 +8,8 @@ repos:
         exclude: ^notice/overrides.json
     -   id: check-merge-conflict
     -   id: check-yaml
+        exclude: >
+          (?x)^(cli/config/kubernetes.*)$
     -   id: check-xml
     -   id: end-of-file-fixer
         exclude: >
--- a/cli/config/kubernetes/README.md
+++ b/cli/config/kubernetes/README.md
@@ -13,7 +13,11 @@ kind create cluster
 kubectl apply -k base
 ```
 
-## Local Ingress
+This will allow you to visit `http://localhost:5601` using username: `elastic` password: `changeme` to login
+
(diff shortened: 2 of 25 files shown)


cachedout:
Have we considered just [using ECK ](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-quickstart.html)instead of rolling our own solution?

.pre-commit-config.yaml
```diff
@@ -8,6 +8,8 @@ repos:
         exclude: ^notice/overrides.json
     -   id: check-merge-conflict
     -   id: check-yaml
+        exclude: >
```
kuisathaverat, line 11: Why we exclude them from the linting?

adam-stokes: I couldn't figure out how to tell yaml-lint to allow multiple documents ina single file.  Is there an option to set to allow that?

adam-stokes:
> Have we considered just [using ECK ](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-quickstart.html)instead of rolling our own solution?

I did consider it, it uses the k8s operator pattern and is a little too opinionated on deployment. For our use cases it was a bit overkill on resources.

mdelapenya (approved):
I think is good to go! We can merge it as is, or once @kuisathaverat's comments are addressed, as it does not impact in the existing test suite: it _only_ brings a new and promising deployment model for the runtime dependencies. 🚀

PR #1116, opened 2021-04-30
Auto bootstrap fleet during initialize scenario

adam-stokes:

## What does this PR do?

This makes fleet server automatically available in bootstrapped mode so that additional agents just need to enroll in the default policy for tests.

## Why is it important?

Speeds up the test scenarios by having a working environment to test the additional elastic-agents

## Checklist


cachedout (approved):

mdelapenya (approved):
LGTM!

mdelapenya:
One consideration to highlight is that after moving the bootstrap of the fleet-server to the startup of the runtime dependencies we are losing the ability to test the bootstrap itself. We agreed in removing that support, as we consider the bootstrap of the fleet server a pre-condition to be satisfied before running any test. At the moment fleet-server fails to bootstrap, for any reason, we will fail the entire test suite.

PR #1118, opened 2021-05-04
chore: initialise timeout factor next to the declaration

mdelapenya:
## What does this PR do?
It initialises the timeout factor in the common package, avaoiding the multiple initialisation performed in all test suites.

Besides that, we removed a local version of the timeout factor variable for the Helm test suite, using the default one.

## Why is it important?
DRY and KISS

## Checklist

## Related issues

- Suggested by @jsoriano in https://github.com/elastic/e2e-testing/pull/1115#discussion_r625185882

### DIFF
--- a/e2e/_suites/fleet/ingest_manager_test.go
+++ b/e2e/_suites/fleet/ingest_manager_test.go
@@ -48,7 +48,6 @@ func setUpSuite() {
 	}
 	common.AgentVersionBase = v
 
-	common.TimeoutFactor = shell.GetEnvInteger("TIMEOUT_FACTOR", common.TimeoutFactor)
 	common.AgentVersion = shell.GetEnv("BEAT_VERSION", common.AgentVersionBase)
 
 	// check if version is an alias
--- a/e2e/_suites/kubernetes-autodiscover/autodiscover_test.go
+++ b/e2e/_suites/kubernetes-autodiscover/autodiscover_test.go
@@ -37,14 +37,6 @@ const defaultBeatVersion = "8.0.0-SNAPSHOT"
 var defaultEventsWaitTimeout = 60 * time.Second
 var defaultDeployWaitTimeout = 60 * time.Second
 
-func init() {
-	// initialise timeout factor
-	common.TimeoutFactor = shell.GetEnvInteger("TIMEOUT_FACTOR", common.TimeoutFactor)
-
-	defaultEventsWaitTimeout = defaultEventsWaitTimeout * time.Duration(common.TimeoutFactor)
-	defaultDeployWaitTimeout = defaultDeployWaitTimeout * time.Duration(common.TimeoutFactor)
-}
-
 type podsManager struct {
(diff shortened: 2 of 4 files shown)


e2e/_suites/kubernetes-autodiscover/autodiscover_test.go
```diff
@@ -472,6 +464,9 @@ func InitializeTestSuite(ctx *godog.TestSuiteContext) {
 		// init logger
 		config.Init()
 
+		defaultEventsWaitTimeout = defaultEventsWaitTimeout * time.Duration(common.TimeoutFactor)
```
mdelapenya, line 467: Because Golang does not warrantee the initialisation order of the init functions, we must move this to the setupSuite, which we are sure will happen after that.

e2e/_suites/fleet/ingest_manager_test.go
```diff
@@ -48,7 +48,6 @@ func setUpSuite() {
 	}
 	common.AgentVersionBase = v
 
-	common.TimeoutFactor = shell.GetEnvInteger("TIMEOUT_FACTOR", common.TimeoutFactor)
 	common.AgentVersion = shell.GetEnv("BEAT_VERSION", common.AgentVersionBase)
```
jsoriano, line 51: We could do the same with any other variable defined in `common` that needs some initialization, such as this `common.AgentVersion`.

e2e/_suites/kubernetes-autodiscover/autodiscover_test.go
jsoriano: There are some guarantees in order of init executions. In this case I think that `common.TimeoutFactor` should have been initialized before `InitializeTestSuite` is called. Wasn't it?

internal/common/retry.go
```diff
@@ -8,12 +8,17 @@ import (
...
+	"github.com/elastic/e2e-testing/internal/shell"
 )
 
 // TimeoutFactor a multiplier for the max timeout when doing backoff retries.
 // It can be overriden by TIMEOUT_FACTOR env var
 var TimeoutFactor = 3
 
+func init() {
+	TimeoutFactor = shell.GetEnvInteger("TIMEOUT_FACTOR", TimeoutFactor)
+}
```
jsoriano, line 20: Now that I see this together. Even better, this should work too, so `init()` is not needed.
```
const defaultTimeoutFactor = 3
var TimeoutFactor = shell.GetEnvInteger("TIMEOUT_FACTOR", defaultTimeoutFactor)
```

e2e/_suites/kubernetes-autodiscover/autodiscover_test.go
mdelapenya: Yes, in this case we moved the defaultEventsXXX timeouts to the InitializeTestSuite for that reason, so it's guaranteed that the default value is read from the environment.

PR #1123, opened 2021-05-04
Pull fresh docker images before suite

jalvz:
## What does this PR do?

Pull fresh Docker images before the suite runs, with the possibility to opt out. If Docker images are current, this takes a couple of seconds, which is marginal considering how long the whole suite takes (even just one scenario).

This makes sure that a test only uses the most recent images and not stale ones.

The clean-docker make target is still kept for convenience, although it probably can be removed in a follow up PR.


adam-stokes (approved):

e2e/_suites/fleet/ingest_manager_test.go
```diff
@@ -215,3 +223,31 @@ func InitializeIngestManagerTestSuite(ctx *godog.TestSuiteContext) {
 		}
 	})
 }
+
+func pullFreshImages(stackVersion, agentVersion, kibanaVersion string) error {
```
mdelapenya, line 227: Wdyt about moving an abstraction of the PullFreshImages method to the internal docker package?

e2e/_suites/fleet/ingest_manager_test.go
```diff
@@ -135,6 +138,11 @@ func InitializeIngestManagerTestSuite(ctx *godog.TestSuiteContext) {
 			"stackVersion":  common.StackVersion,
 		}
 
+		if !shell.GetEnvBool("SKIP_PULL") {
```
mdelapenya, line 141: There is a section for `Environment variables affecting the build`: https://github.com/elastic/e2e-testing/tree/master/e2e#environment-variables-affecting-the-build Can we add info about this new var there?

adam-stokes (changes requested):
I like @mdelapenya idea of moving `pullFreshImages` into `internal/docker`, once that's done this PR should be good to go

internal/docker/docker.go
```diff
@@ -436,3 +437,32 @@ func getDockerClient() *client.Client {
 
 	return instance
 }
+
+// PullImages pulls images
+func PullImages(stackVersion, agentVersion, kibanaVersion string) error {
```
mdelapenya, line 442: One little one: because this method could be reused across different test suites, i.e. metricbeat, we could pass the images array as parameter.

Again, it's not a blocker and it could be done in a follow-up PR.

mdelapenya (approved):

adam-stokes (approved):

PR #1130, opened 2021-05-04
Support multiple deployment backends

adam-stokes:
## What does this PR do?

This abstracts out the deployment of the environment to support different providers. Currently, this supports deploying to both docker and kubernetes.

## Why is it important?

Allows us to expand our testing landscape easily by incorporating different environments.

## Checklist

## Follow-ups

Note, that the kubernetes provider will successfully deploy an environment suitable for testing. However, the installer code for getting an elastic-agent provisioned and registered is still heavily tied to Docker. That work will be done in a subsequent PR to allow us to perform the installation regardless of what environment is being used.


cachedout (approved):

internal/docker/docker.go
```diff
@@ -199,8 +199,7 @@ func InspectContainer(name string) (*types.ContainerJSON, error) {
 	ctx := context.Background()
 
 	labelFilters := filters.NewArgs()
-	labelFilters.Add("label", "service.owner=co.elastic.observability")
-	labelFilters.Add("label", "service.container.name="+name)
+	labelFilters.Add("label", "com.docker.compose.service="+name)
```
mdelapenya, line 202: While working on the docker copy stuff, I noticed this was wrong from the beginning. Instead, we should use:
```suggestion
	labelFilters.Add("name", name)
```

e2e/_suites/fleet/fleet.go
```diff
@@ -335,7 +333,7 @@ func (fts *FleetTestSuite) anAgentIsDeployedToFleetWithInstallerAndFleetServer(i
 	fts.CurrentTokenID = enrollmentKey.ID
 
 	var fleetConfig *kibana.FleetConfig
-	fleetConfig, err = deployAgentToFleet(agentInstaller, containerName, fts.CurrentToken)
+	fleetConfig, err = deployAgentToFleet(agentInstaller, fts.deployer, containerName, fts.CurrentToken)
```
mdelapenya, line 336: This is super NIT: as we commented online yesterday, the hierarchy tells us that the deployer has more importance than the installer. Wdyt about moving the deployer param to the first place in the method signature?

e2e/_suites/fleet/fleet.go
```diff
@@ -46,21 +47,18 @@ type FleetTestSuite struct {
 	FleetServerPolicy   kibana.Policy
 	Version             string // current elastic-agent version
 	kibanaClient        *kibana.Client
-
+	deployer            deploy.Deployment
```
mdelapenya, line 50: If this represents a deployment, let's name the variable as `deployment`. Or maybe we should rename the struct tp `Deployer` 🤔

mdelapenya (approved):
This PR is the seed for the new deployment model, good job!

Added a few non-blocking comments, but LGTM!

e2e/_suites/fleet/fleet.go
adam-stokes: Yea once I get the installer code in place, I think we'll be calling install commands from the deployer anyway and that agentInstaller param will go away

e2e/_suites/fleet/fleet.go
adam-stokes: Let me work with the naming of things in a subsequent PR when I have more of the groundwork done for the installer portion

PR #1136, opened 2021-05-05
feat: use Docker copy to transfer binaries to containers

### LINKED ISSUE #407: [Fleet] instead of using volumes to transfer Agent file, can we maybe copy it after the container is in place?
Logged after discussion with Manu.  This will better support repeat execution in 'DEV' mode

mdelapenya:
## What does this PR do?
Instead of mounting volumes for the binaries, which made it more complex to setup a single agent, we are copying the downloaded file into the container.

How docker CLI works for copying TAR files is that it automatically extracts the TAR file, and for that reason we are removing the PreInstall step we had to perform exactly that operation.

To verify the new implementation, we have added unit tests that runs a busybox container, copying regular and tar files into the container.

Finally, we added a few exceptions in precommit to avoid the hooks to modify the test TAR file with EOFs

## Why is it important?
This abstraction will help with the deployment refactors we are taking into the framework

## Checklist

```shell
make unit-test
```

## Related issues

- Closes #407 

## Follow-ups
Once we are happy with this implementation, we should replace the existing volumes we have for configuration files that apply only on test time, making it possible to use the CLI part to run services without any coupling to the previous mounted files.


internal/docker/docker.go
```diff
@@ -58,6 +91,72 @@ func CheckProcessStateOnTheHost(containerName string, process string, state stri
...
+			"container": containerName,
+			"error":     err,
+			"src":       srcPath,
+			"parentDir": parentDir,
+		}).Error("Could not open file to deploy")
+		return err
+	}
+	defer file.Close()
+
+	// TODO: detect the file has TAR headers
```
adam-stokes, line 129: Ive used this github.com/gabriel-vasile/mimetype with a lot of success for this sort of thing

adam-stokes (approved):

cachedout (approved):

internal/docker/docker.go
mdelapenya: Let's do this in a follow-up PR

PR #1146, opened 2021-05-06
fix: bump stale agent to 7.13-SNAPSHOT

mdelapenya:
## What does this PR do?
It uses one minor version for the stale agent

## Why is it important?
For master branch, using 8.0.0-SNAPSHOT, it's incorrect to use 7.11-SNAPSHOT as stale agent, as that version has disappeared from the artifactory.

## Checklist

## Author's Checklist

- [ ] @michalpristas, can you confirm that the stale version is correct for master?

## Follow-ups
Thinkg about upgrade processes, what if we want to test the upgrade of an old version? Using the artifact repository won't be enough, as versions disappears after 2 minor versions of the stack.

Maybe this is something to ask infra, about allowing accessing those old versions from another repository.


mdelapenya:
I'm triggering the CI job running the nightly tests, as the upgrade test is run in that schedule

michalpristas (approved):
looks good

Issue #1108, opened 2021-04-28
Improve resiliency of testing suit


jalvz:
`make test` should be as independent as possible and pass/fail on its own terms, not in function of machine state, other commands, etc.

- Pull fresh Docker images without needed a separate command.
- Clean up .op folder after each run (to avoid keeping garbage around eg. when switching branches)
- Get docker-compose files for metricbeat directly from the network instead of forcing a whole Beats checkout

mdelapenya:
I think this feature it's almost implemented:
- [x] compose files are refreshed on each test suite
- [x] each test suite is pulling its own set of images (configurable with the SKIP_PULL variable)
- [ ] metricbeat's compose files is not implemented yet.

mdelapenya:
For the latest, I think it would come with a a big refactor:

- the feature file does not make a differentiation on x-pack Vs non-x-pack modules (well, it's done at the annotation/tag level, but that value is not passed from the feature file to the Go code). Therefore, we cannot dynamically fetch the compose file for the integration from the Beats repo directly
- we could add the xpack field to the outline examples in the feature file, and pass it to the scenario Go code. Then we could download the compose file (and related files, after sanitising them) and use it to spin up the integration service

mdelapenya:
I think we can close this one, as metricbeat test suite was removed in favour of the fleet one

Please reopen if needed

PR #1152, opened 2021-05-06
Use the correct beats repo

jalvz:
## What does this PR do?

This PR improves how Metricbeat compose files are synced:

- The test suite should be as independent as possible, and not depend on additional commands to run before. Therefore, the beats repo is fetched every time the test suite with the possibility to opt out.

- The repo is cloned / fetched for the current branch always. Before one would need to remember to delete and re-clone the repo when switching branches, or risk having the wrong compose files.

- It will only clone / fetch the latest commit, not the entire history. This speeds up the operation considerably. 

- The functionality, now that is used in a different place, is moved out of the `cmd` package.

## Checklist

## How to test this changes

Run some scenario with different versions, check that `.op/git/beats` checks out the right branch.

Related to https://github.com/elastic/e2e-testing/issues/1108


e2e/_suites/fleet/ingest_manager_test.go
```diff
@@ -132,6 +136,24 @@ func InitializeIngestManagerTestSuite(ctx *godog.TestSuiteContext) {
 			docker.PullImages(images)
 		}
 
+		if !shell.GetEnvBool("SKIP_FETCH") {
```
mdelapenya, line 139: I think Fleet test suite does not need the metricbeat compose files. Do you think this applies to the metricbeat test suite instead?

internal/git/git.go
```diff
@@ -90,7 +90,7 @@ func (b projectBuilder) withUser(user string) projectBuilder {
 var ProjectBuilder = builder.Register(projectBuilder{}, Project{}).(projectBuilder)
 
 // Clone allows cloning an array of repositories simultaneously
-func Clone(repositories ...Project) {
+func Clone(repositories ...Project) error {
```
mdelapenya, line 93: 👍

internal/compose/sync.go
```diff
@@ -0,0 +1,270 @@
+package compose
```
mdelapenya, line 1: Although the goal of this code is to fetch compose files from Beats' integrations, I'd not put this sync code under the `compose` package, which is managing the deployment of services using Docker Compose. I think it could fit more under the `state` or even `sanitizer` packages, but not having a strong opinion. What other package would you use?

mdelapenya:
Closing, as metricbeat tests are not supported in this test suite anymore

PR #1153, opened 2021-05-10
feat: check how many processes of a process are running in the host

### LINKED ISSUE #218: [Fleet] Parametrize and functionalize the check for the number of beats processes (currently only validates that 1 is running but 2 are the default)
By default, there are two instances for the metricbeat process, one for monitoring the processes

cc/ @EricDavisX

mdelapenya:
## What does this PR do?
It creates a new step that abstract the logic to get the number of processes in a desired state, calling this new step in the existing one (with 1 as number of ocurrences).

We had to refactor the existing code to wait for processes, as it only checked that a process existed in the host using `pgrep -n -l`, which only checks for the existence of the latest process for a process.

With this new approach, we are switching to `pgrep -d , $THE_PROCESS`, which separates each pid for the process with the comma delimiter. Having that list of pids for a process we execute `ps -q $PID -o state --no-headers` (linux only) to get the current state of that pid, to discard zombie processes (we detected that the elastic-agent on Centos creates 3 processes, 2 of them in the zombie state cc/ @michalpristas).

Finally, we refactored the existing method to get the output of executing a command in a container to avoid receiving extra bytes at the beginning of the output. In the previous code we had a list of bytes to remove, but now we simply use a Docker client helper utility (stdcopy) that splits stdout from stderr in the container, making possible to retrieve just the output. See https://stackoverflow.com/a/57132902 for further information.

## Why is it important?
It creates a more comprehensive check about the background processes that the elastic-agent spawns in the host, as we want to make sure the number of processes is correct.

## Checklist

Tests for the backend processes:
```
SUITE="fleet" TAGS="backend_processes" TIMEOUT_FACTOR=3 LOG_LEVEL=TRACE BEATS_USE_CI_SNAPSHOT=false DEVELOPER_MODE=false make -C e2e functional-test
```

#### Considerations for the stand-alone mode using the Docker image
We noticed that the elastic-agent running in the docker image in stand-alone mode spawns only 1 filebeat process, instead of 2 as using the TAR/DEB/RPM installers. Maybe @michalpristas can help out here.
```
SUITE="fleet" TAGS="stand_alone_agent" TIMEOUT_FACTOR=3 LOG_LEVEL=TRACE BEATS_USE_CI_SNAPSHOT=false DEVELOPER_MODE=false make -C e2e functional-test
```

## Related issues

- Closes #218

### DIFF
--- a/e2e/_suites/fleet/features/backend_processes.feature
+++ b/e2e/_suites/fleet/features/backend_processes.feature
@@ -6,8 +6,8 @@ Feature: Backend Processes
 Scenario Outline: Deploying the <os> agent
   Given a "<os>" agent is deployed to Fleet with "tar" installer
   When the "elastic-agent" process is in the "started" state on the host
-  Then the "filebeat" process is in the "started" state on the host
-    And the "metricbeat" process is in the "started" state on the host
+  Then there are "2" instances of the "filebeat" process in the "started" state
+    And there are "2" instances of the "metricbeat" process in the "started" state
 
 @centos
 Examples: Centos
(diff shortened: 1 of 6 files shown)


mdelapenya:
As commented with @michalpristas online, the Docker image has a default configuration in which it collects metrics but not logs, but monitors everything, hence there are 2 metricbeat processes and 1 filebeat process.

adam-stokes (approved):

mdelapenya:
@adam-stokes I'm going to send a follow-up PR adding this to the deployer struct, so that we can start decoupling stuff preparing the multiplatform support

PR #1158, opened 2021-05-11
chore(mergify): delete upstream branches on merge

mdelapenya:
## What does this PR do?
It tells mergify to remove the upstream branches when merging a PR from the upstream repository.

## Why is it important?
To keep the repository clean (and satisfy my obsessive compulsive disorder about outdated branches)

## Checklist

## Author's Checklist

- [ ] Checked Mergify docs

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -49,3 +49,8 @@ pull_request_rules:
         branches:
           - "7.12.x"
         title: "[{{ destination_branch }}](backport #{{ number }}) {{ title }}"
+  - name: delete head branch after merge
+    conditions:
+      - merged
+    actions:
+      delete_head_branch:


cachedout (approved):

.mergify.yml
```diff
@@ -49,3 +49,8 @@ pull_request_rules:
         branches:
           - "7.12.x"
         title: "[{{ destination_branch }}](backport #{{ number }}) {{ title }}"
+  - name: delete head branch after merge
+    conditions:
+      - merged
+    actions:
+      delete_head_branch:
```
v1v, line 56: I guess, the PRs will never be created from any upstream branches, right? If so, +1

v1v: Yes, I think that's something `mergify` does under the hood for the backport action.

I only configured for the specific bump automation we have in place in other repos.

Issue #1080, opened 2021-04-23
Support reading environment variables from ".env"


mdelapenya:
There are plenty of libraries reading configs from env vars from a well-known location (i.e. the `.env` dir), adding a chain of priority for multiple files (production, testing, local, etc)

We could leverage this kind of tools to simplify the execution of the tests.

mdelapenya:
Yep, I mentioned it here: https://github.com/elastic/e2e-testing/pull/1016#discussion_r612310214. We can give it a try

mdelapenya:
Implemented using [godotenv](https://github.com/joho/godotenv) in #2490

PR #1159, opened 2021-05-11
feat: simplify the initialisation of versions

mdelapenya:
## What does this PR do?

It moves initialisation code around, locating it next to the declaration of the variables holding the references, in the `defaults.go` file. To achive that, we:

- created a file for the stack version, which will be used by shell scripts. They will read that file, so the version value s declared once.
- moved the retry.go file from the `common` package to the `utils` one, to avoid cyclic references.
- initialised `StackVersion, AgentVersionBase and AgentVersion` once.
- renamed the AgentVersionBase to BeatVersionBase, updating references in Metricbeat test suite.
- renamed the AgentVersion to BeatVersion, updating references in Matricbeat test suite.

Not directly related to the initialisation itself, but to the version calculation, we created a cache for the requests to the Elastic's artifact endpoint, to calculate versions. This cache is implemented with a map of string to string, being the URL the key, and the version the cached value. Because we read these values once at the beginning of the test suite, it's virtually impossible to get invalid values for the same URLs, as the endpoint response changes in a very distinct schedule than the execution of this test suite.

## Why is it important?
Simplify the configuration and setting the foundations to use a library for environment configuration.

## Checklist

Running the tests

## Related issues

- Relates #1080

## Follows-up
- Convert the `.stack-version` file into a configuration file, with all variables used by the test suite.


cachedout (approved):

.ci/scripts/clean-docker.sh
```diff
@@ -9,7 +9,7 @@ set -euxo pipefail
 # Build and test the app using the install and test make goals.
 #
 
-readonly VERSION="8.0.0-SNAPSHOT"
+readonly VERSION="$(cat $(pwd)/.stack-version)"
```
kuisathaverat, line 12: This implementation depends on the folder you launch the script, it must be the root of the repo otherwise do not work. There is an alternative but the `.stack-version` should be moved to the `.ci/scripts` folder that I am not sure is a good idea.

```
srcdir=$(dirname "$0")
test -z "$srcdir" && srcdir=.
readonly VERSION="$(cat ${dirname}/.stack-version)"
```

kuisathaverat (approved):

PR #1162, opened 2021-05-12
feat: support flavours in services, specially in the elastic-agent

### LINKED ISSUE #1160: Support for multiple flavours of a service
We are supporting multiple services for the elastic-agent:
- centos-systemd
- centos
- cloud
- debian-systemd
- debian

And each of them means having a separate dir in the services layout.

1) We could remove the vanilla centos & debian services, are they are not used nor going to be used.
2) We are hardcoding the `centos-systemd` name for the service, which is causing strong coupling with the installer structs, and service names. Simplifying the layout we could simply rename the service (and eventually the container name) for `elastic-agent`, period.

cc/ @adam-stokes

mdelapenya:
## What does this PR do?
This PR comes with a big refactor of the internals of the installer & deployer structs:

- moved the `compose.go` file from the `compose` package to the `deploy` one.
- created a ServiceRequest struct, which will require a Name and a Flavour (optional)
- the flavour concept will be represented by a subdir in the service dir, so that the dir layout will be simplified. I.e., the elastic-agent will have subdirs per flavours (cloud, centos, debian...). Inside those child dirs, a `docker-compose.yml` file will represent the service in that particular flavour.
- refactored all service methods in the providers to pass and use ServiceRequests instead of Strings:
  - for kubernetes provider, it would be a matter of using the `request.Name` (@adam-stokes, would need help with the kubernetes implementation to support flavours in k8s descriptors)
  - for compose provider, it will pass the struct to the most internal method (`executeCompose`), and that one will be the responsible of unwrapping the struct, adding a path separator between the service and its flavour: i.e. `elastic-agent/centos`
- removed unused services (centos and debian) as they were not used at all.
- moved Docker client operations to the deploy package, to avoid cyclic dependencies. We will need another abstraction to represent the Docker client operations, as it's clear what is a deployment and what is an operation in the deployment. Maybe a Client struct for each provider will help out in differenciate it: something like `deployer.Client.Operation` :thinking:
- added a new step in Fleet test suite to infer the installer type by the underlying OS image. If centos is selected, then the installer will be 'rpm', 'deb' for debian. This new step will call the existing ones.
- made sure that all service requests for the elastic-agent is using the flavour: for Fleet test suite, it will read it from the installer struct. For Standalone test suite, where it can be `cloud` or the regular one, it will be passed to the methods.

## Why is it important?
The most important goal is to simplify project layout, not only in the file system but in the structs used to represent the different installers. Once achieved we can extend the flavours in an easy manner. Apart from that, the code now it's more structured in the way of the input parameters used to create services: instead of strings, the user will have to create a service request on purpose.

## Checklist

## Author's Checklist

- [ ] @adam-stokes we need to work on the k8s part of the deployment

Running the tests for Fleet (fleet_agent_mode and backend_processes)

## Related issues

- Closes #1160 

## Follow-ups
There is room for improvement in the installer struct. I did not want to continue refactoring it to avoid a long lived branch. We can do it in follow-ups


mdelapenya:
Ups, did not rebased first. Will send the merge resolving conflicts

adam-stokes:
This looks awesome, I did want to point out that some of this code will conflict with the work im doing on getting the installers cleaned up. Specifically in the areas of executing and adding files to the environment as I pass in the deployers ExecIn and Addfiles functions into the calling Mount function when loading the installers

https://github.com/elastic/e2e-testing/compare/feat-installer-rework

cli/cmd/deploy.go
```diff
@@ -63,12 +63,16 @@ func buildDeployServiceCommand(srv string) *cobra.Command {
 		Short: `Deploys a ` + srv + ` service`,
 		Long:  `Deploys a ` + srv + ` service, adding it to a running profile, identified by its name`,
 		Run: func(cmd *cobra.Command, args []string) {
-			serviceManager := compose.NewServiceManager()
```
adam-stokes, line 66: Rather than calling `deploy.NewServiceManager` we could initialize it like we do in the fleet tests that way you just use `deployer.Add` instead and it'll do the right thing no matter the provider (docker or k8s)

e2e/_suites/fleet/fleet.go
```diff
@@ -348,7 +365,7 @@ func (fts *FleetTestSuite) anAgentIsDeployedToFleetWithInstallerAndFleetServer(i
 	}
 
 	// get container hostname once
-	hostname, err := docker.GetContainerHostname(containerName)
```
adam-stokes, line 351: In my installer rework code, we don't need to query the hostname as they is done by inspecting the service further abstracting out connection information. We want to be able to just perform actions on a "service" and not have to be so explicit about how to access it as that is handled by the deployment and installer abstractions

e2e/_suites/fleet/fleet.go
```diff
@@ -361,7 +378,7 @@ func (fts *FleetTestSuite) anAgentIsDeployedToFleetWithInstallerAndFleetServer(i
 // we are using the Docker client instead of docker-compose because it does not support
 // returning the output of a command: it simply returns error level
 func (fts *FleetTestSuite) getContainerName(i installer.ElasticAgentInstaller, index int) string {
-	return fmt.Sprintf("%s_%s_%s_%d", i.Profile, i.Image, common.ElasticAgentServiceName, index)
```
adam-stokes, line 364: Same thing here, in the other branch we use an `Inspect` function to populate the service metadata so that we can just reference that rather than trying to piece together the container name as these would differ between docker and k8s

e2e/_suites/fleet/fleet.go
```diff
@@ -391,17 +408,20 @@ func (fts *FleetTestSuite) processStateChangedOnTheHost(process string, state st
 
 	serviceName := agentInstaller.Service // name of the service
 
+	profileService := deploy.NewServiceRequest(profile)
+	imageService := deploy.NewServiceRequest(common.ElasticAgentServiceName).WithFlavour(agentInstaller.Image)
+
 	if state == "started" {
-		return installer.SystemctlRun(profile, agentInstaller.Image, serviceName, "start")
```
adam-stokes, line 395: In my branch these are actually abstracted out to the installer service as `Start` and `Stop` and is specific to the service we are installing (this could differ with other services that may not use systemctl)

internal/deploy/compose.go
```diff
@@ -48,17 +48,27 @@ func (sm *DockerServiceManager) AddServicesToCompose(ctx context.Context, profil
...
 
 	persistedEnv := state.Recover(profile.Name+"-profile", config.Op.Workspace)
 	for k, v := range env {
 		persistedEnv[k] = v
 	}
 
-	err := executeCompose(true, newServices, []string{"up", "-d"}, persistedEnv)
+	cmds := []string{"up", "-d"}
+	if len(scaleCmds) > 0 {
+		cmds = append(cmds, "--scale")
```
adam-stokes, line 67: Nice!!

cli/cmd/deploy.go
```diff
@@ -85,12 +89,16 @@ func buildUndeployServiceCommand(srv string) *cobra.Command {
 		Short: `Undeploys a ` + srv + ` service`,
 		Long:  `Undeploys a ` + srv + ` service, removing it from a running profile, identified by its name`,
 		Run: func(cmd *cobra.Command, args []string) {
-			serviceManager := compose.NewServiceManager()
```
adam-stokes, line 88: What's the reasoning for moving docker compose specific calls into the deploy namespace?

e2e/_suites/metricbeat/metricbeat_test.go
```diff
@@ -378,14 +378,14 @@ func (mts *MetricbeatTestSuite) runMetricbeatService() error {
 			return err
 		}
 
-		err = docker.LoadImage(imagePath)
+		err = deploy.LoadImage(imagePath)
```
adam-stokes, line 381: I think these should stay in the docker namespace, the deployer is an abstraction on top of docker/k8s and I think this ties deployer to much to docker. I think a better approach is to have a deploy.LoadImage that supports both docker and k8s way of importing images into a registry

e2e/_suites/metricbeat/metricbeat_test.go
```diff
@@ -378,14 +378,14 @@ func (mts *MetricbeatTestSuite) runMetricbeatService() error {
...
 		if err != nil {
 			return err
 		}
 
 		mts.Version = mts.Version + "-amd64"
 
-		err = docker.TagImage(
+		err = deploy.TagImage(
```
adam-stokes, line 388: Same here, abstract out tagging an image to support both docker/k8s and keep docker's version of TagImage in its own namespace

cli/cmd/deploy.go
mdelapenya: The reason was to avoid cyclic references in packages. I explained it in 9c9768dfc506b5b769dbc7316a8cbf16534d512d commit message:

>We will need another abstraction to represent the Docker client operations,
    as it's clear what is a deployment and what is an operation in the deployment.
    Maybe a Client struct for each provider will help out in differenciate it

e2e/_suites/metricbeat/metricbeat_test.go
mdelapenya: Yep, I totally agree with you about the docker namespace would be the right thing to do. But because of the ServiceRequest struct, which resides in the deploy namespace, and how `deploy` package calls the `docker` one, it was not possible to define the ServiceRequest struct inside the `deploy` package without a cyclic reference.

mdelapenya: I did that as a compromise to add progress. Maybe the Provider.Client struct would solve this issue, which we could discuss/face it in follow up PRs

internal/deploy/compose.go
mdelapenya: Yeah, I verified that the scale option scales the base image, not the current state of the service. OTOH, I did not include the k8s version to discuss it with you first

cli/cmd/deploy.go
adam-stokes: Ah good catch on the cyclic references!

e2e/_suites/metricbeat/metricbeat_test.go
adam-stokes: Yep makes sense now!

adam-stokes (approved):
I think we should pull this in as is, and I can rework my installer code to make use of the new abstractions for services. There is some code in here that will change based on my work but I think it's mostly within the installer and interacting with the mounted services

cli/cmd/deploy.go
mdelapenya: If we do that we will be introducing a bug in the CLI side, as we have coupled the Bootstrap method with the Fleet profile. Maybe we can fix that in a separate PR.

mdelapenya: Well, indeed the bug is there if we decide to migrate the metricbeat test suite 😄

cli/config/compose/profiles/fleet/docker-compose.yml
```diff
@@ -29,7 +29,7 @@ services:
       test: "curl -f http://localhost:5601/login | grep kbn-injected-metadata 2>&1 >/dev/null"
       retries: 600
       interval: 1s
-    image: "docker.elastic.co/${kibanaDockerNamespace:-beats}/kibana:${kibanaVersion:-8.0.0-SNAPSHOT}"
```
adam-stokes, line 32: I always wondered about this one :)

mdelapenya: Wrong copy&paste on my side for sure 🤦 

I think I'm gonna merge it after the CI passes

PR #1186, opened 2021-05-19
fix: run linux integration on CI

mdelapenya:
## What does this PR do?
It includes the linux integration scenarios in the CI job, as a new parallel branch

## Why is it important?
This suite was missing from CI execution

## Checklist


jalvz:
Shouldn't we do something so that every new scenario added is executed on CI without more ceremony? ie, have the option maybe to opt out in case it is useful, instead of having to opt in.

kuisathaverat (approved):

mdelapenya:
> Shouldn't we do something so that every new scenario added is executed on CI without more ceremony? ie, have the option maybe to opt out in case it is useful, instead of having to opt in.

Agree There is an open issue to standardise the tags used in feature files, which could simplify how they are spread across the different parallel executions. I can picture this:

1. list all feature files
2. change how the Make goals run the tests: instead of the tags use the file (`-f $THE_FEATURE_FILE`)
3. generate the descriptor file on the fly based on a template, using one feature file per parallel branch
4. update calls to make goals to not pass tags

If this makes sense to you, we can convert this comment into a new issue to start working on it, as I totally agree with you about the ceremonies :)

mdelapenya:
Another idea, related to the used tags, could be to group all integrations under the `@integration` tag, and run them all together: we'd have endpoint and linux integration under the same parallel execution on CI

jalvz:
> change how the Make goals run the tests: instead of the tags use the file (-f $THE_FEATURE_FILE)

could we be even a folder?

mdelapenya:
> > change how the Make goals run the tests: instead of the tags use the file (-f $THE_FEATURE_FILE)
> 
> could we be even a folder?

Yes, or a glob expression: *.feature. but on CI we want to create a separation of concerns power feature, to speed up the build

Issue #687, opened 2021-02-01
Instrument the test framework with the APM Go agent


mdelapenya:
We need to decide what cluster we are going to send the information at CI time and at development time, and with what credentials.

## Use cases

```gherkin
Given a developer without access to the credentials for the Jenkins ES cluster
When the developer runs the tests
Then the tests are executed
   But no APM traces are sent to the Jenkins ES cluster
```

```gherkin
Given a developer with access to the credentials for the Jenkins ES cluster
When the developer runs the tests
Then the tests are executed
   And APM traces are sent to the Jenkins ES cluster in the "development" environment
```

```gherkin
Given the CI has access to the credentials for the Jenkins ES cluster
When the CI the tests
Then the tests are executed
   And APM traces are sent to the Jenkins ES cluster in the "CI" environment
```

cc/ @elastic/observablt-robots 

## Related issues
- [x] Helm test suite
- [x] Metricbeat test suite
- [x] Fleet test suite
- [x] Kubernetes autodiscover test suite

PR #1234, opened 2021-06-02
feat: add APM traces for Fleet test suite

mdelapenya:
## What does this PR do?
This PR adds APM traces for the Fleet test suite, and for that we had to basically pass the Go context through all API methods. We are:

- starting a transaction before the test scenario
  - adding a span for each internal operation/method, passing the current context as parent
  - each operation defers the close of the span
- closing the transaction after the test scenario
- using Godog's `beforeStep/afterStep` life cycle hooks to automatically create/close spans per scenario step, which will serve as parent span for the internal operations within it. This life cycle hook will be used to capture step errors, sending APM errors to the Cloud.
- using Godog's `beforeScenario/afterScenario` life cycle hooks to start/flush the transaction for each scenario. The transaction will receive the `test.scenario` type.
- using Godog's `beforeSuite/afterSuite` to create specific transaction for initialisation/destruction of the runtime dependencies. The transaction will receive the `test.suite` type.

Another thing we are doing here is to remove the local APM environment, which provisioned a local APM Server and a Kibana (if/when needed). This local environment was intended to be used to capture test executions while developing the framework, but for Fleet we needed to create a lot of custom resources (Users, roles, etc) that made it a bit complex. Instead, we are going to send the APM traces directly to the Cloud, but using the `local` APM environment (see the `ELASTIC_APM_ENVIRONMENT` variable).

## Why is it important?
Fleet is the most active and most critical project, and we want to measure where we are wasting resources: in particular the integration with the Security App, which we see long delays while waiting for responses. For that reason, we want to share metrics with the Fleet team about that, so that we can count on numbers, not in guesses.

## Checklist

## Author's Checklist

- [ ]

It will use `local` as APM environment (guided by the `ELASTIC_APM_ENVIRONMENT` environment variable).

```shell
$ TAGS="linux_integration" TIMEOUT_FACTOR=3 LOG_LEVEL=TRACE DEVELOPER_MODE=true ELASTIC_APM_ACTIVE=true make -C e2e/_suites/fleet functional-test
```

## Related issues

- Completes #687


cachedout (approved):

adam-stokes (approved):

mdelapenya:
The failing error comes and vanishes from build to build. Merging and will inspect the flaky test analyser for it.

PR #1245, opened 2021-06-09
chore: build Kibana's ARM image in parallel

mdelapenya:
## What does this PR do?
It uses a parallel branch to build the ARM image on CI. Because the code building the image is exactly the same for ARM and AMD, we are encapsulating the groovy code to invoke the shared pipeline library steps into a function, that will be called in both branches of the parallel stage.

## Why is it important?
E2E are running on ARM since past week, but the kibana image used when testing a Kibana PR is not built for that platform, with this PR we'll cover that use case: a kibana PR being tested in both supported platforms (AMD and ARM).

## Checklist

## Author's Checklist

- [ ] Parallel branch and agents

## Related issues

- Closes https://github.com/elastic/apm-pipeline-library/issues/1146


kuisathaverat (approved):

.ci/e2eKibana.groovy
```diff
@@ -46,15 +46,57 @@ pipeline {
...
+        }
+        stage('ARM build') {
+          agent { label 'arm' }
+          steps {
+            buildKibanaPlatformImage('arm64')
+          }
+        }
+      }
+    }
+    stage('Push multiplatform manifest'){
```
v1v, line 80: NIT: you can create a meta-stage with one node allocation that will be used in this stage and the following one, in case you'd like to reuse the same worker and reduce the allocation time

v1v (approved):

.ci/e2eKibana.groovy
mdelapenya: @v1v something like in https://github.com/elastic/e2e-testing/pull/1245/commits/2f8a24c0c5861e05fc28d0a1624ddcb99f0a0552?

PR #1272, opened 2021-06-21
feat: support running elastic-agent alongside Beats

### LINKED ISSUE #330: [Elastic Agent] Add e2e test for Agent to be installed on top of Filebeat & Metricbeat docker containers
This comes from Jason / Ruflin who reported some problems seen somewhere and we want to have a small automated test in place to make sure it always works, at least on a basic level.

I'll add this ticket to the main e2e testing issue for tracking.

Output of ticket:
elastic/e2e-testing repo has an Ingest Manager / Agent oriented test that proves a Filebeat (and Metricbeat) container can have Fleet Agent installed on it and both sets of processes will stream data to ES.

mdelapenya:
## What does this PR do?
This PR reuse existing building blocks to create a new scenario in which the elastic-agent installer is installed on a vanilla systemd linux box (Centos and Debian) which has already installed certain Beats processes (filebeat, metricbeat).

For that reason, we are adding the ability to add background processes to the ServiceRequest struct, so that the installer code for a service is able to also install those processes before installing the agent. We are installing these beats processes using the same mechanism as the one used to install the agent, which means fetching the DEB, RPM files from the artifactory, and then using local tools (APT, YUM) to install them in the aforementioned vanilla linux container.

We are installing the beat in a linux container instead of using the official Beats one (i.e. docker.elastic.co/beats/metricbeat) because we need to install `systemd` in the beat image first. And this problem is resolved if we use our already baked systemd images for Debian and Centos. @ericdavisx, please let us know if not using the official docker images is an issue.

To run the tests, we are adding a new feature file with a new scenario, so that it's possible to filter the test runner to select these new tests. The scenario installs the agent on top of the desired beats process (at this moment filebeat and metricbeat, only), and then verifies the number of processes is two (defined by basic policy) plus the one related to the background process: 3 in total. Finally, the scenario checks the agent is online in Fleet. @ericdavisx, is there any other assertion we should perform here? We chose those ones to reuse the existing steps to the maximum.

To finish, this new feature file is added to the parallel execution on CI, for both ARM and AMD.

As a side note, we merged two methods that were always called in combination into only one: `FetchElasticArtifact` now calls `buildArtifactName` and `fetchBeatsBinary`. These two methods were converted into private (lowercase) to avoid using them separately.

## Why is it important?
It will cover certain use cases reported in https://github.com/elastic/e2e-testing/issues/330

## Checklist

## Author's Checklist

- [ ]

```shell
$ TAGS="running_on_beats && enroll" TIMEOUT_FACTOR=3 LOG_LEVEL=TRACE DEVELOPER_MODE=false ELASTIC_APM_ACTIVE=false make -C e2e/_suites/fleet functional-test
```

## Related issues

- Closes #330

### DIFF
--- a/.ci/.e2e-tests.yaml
+++ b/.ci/.e2e-tests.yaml
@@ -32,6 +32,9 @@ SUITES:
       - name: "Backend Processes"
         pullRequestFilter: " && ~debian"
         tags: "backend_processes"
+      - name: "Beats Background Processes"
+        pullRequestFilter: " && ~debian"
+        tags: "running_on_beats"
   - suite: "metricbeat"
     platforms:
       - "ubuntu-18.04"
--- a/e2e/_suites/fleet/features/running_on_top_of_beats.feature
+++ b/e2e/_suites/fleet/features/running_on_top_of_beats.feature
@@ -1,0 +1,18 @@
+@running_on_beats
+Feature: Running on top of Beats
+  Scenarios for the Elastic Agent verifying the elastic-agent process runs alongside an existing Beats process.
+
+@enroll
+Scenario Outline: Deploying the <os> Elastic-Agent with enroll and then run on top of <beats-process>
+  Given a "<os>" agent is deployed to Fleet on top of "<beats-process>"
+  When the "elastic-agent" process is in the "started" state on the host
+  Then there are "3" instances of the "<beats-process>" process in the "started" state
+    And the agent is listed in Fleet as "online"
+
+@beats
+Examples: Beats
+| os     | beats-process |
+| centos | filebeat      |
+| centos | metricbeat    |
+| debian | filebeat      |
+| debian | metricbeat    |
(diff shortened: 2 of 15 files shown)


adam-stokes (approved):

mdelapenya:
I'd say the failures are consistent with the current state of the master branch. Will mark this as ready for review.

@EricDavisX would you mind checking if the assertions are enough? We are simply checking that there are three instances of the backend processes (metricbeat and filebeat), without doing anything else. Maybe it's worth it to merge this one, and open an issue to create more comprehensive assertions.

mdelapenya:
As per @EricDavisX 's comment in https://github.com/elastic/e2e-testing/issues/330#issuecomment-867947470, we will track work for a deeper validation in https://github.com/elastic/e2e-testing/issues/1292. This is good to go!

Issue #1340, opened 2021-07-20
Remove systemd from the centos/debian images


mdelapenya:
After our daily, we suggested removing the systemd layer in the base images used for testing Fleet. We think this layer could be causing the flakiness described in #1271.

- remove the creation of the systemd-based images for Debian and Centos
- replace systemd-based images with the official ones.

cc/ @elastic/observablt-robots @EricDavisX

mdelapenya:
After replacing the Base Docker images with the official centos/debian images, we see that any execution of the "elastic-agent install" command fails with the following error:

Test error:
>ERRO[2021-07-21T08:18:02+02:00] Error executing command                       args="[exec -u root -i fleet_elastic-agent_1 /elastic-agent/elastic-agent install -e -v --force --insecure --enrollment-token=VnlhMngzb0JXWi0td0Nkb2c4MHY6TXBjaUpzYjRUbU9pRjR6SE5NS2UyZw== --url http://fleet-server:8220]" baseDir=. command=docker error="exit status 1"

Agent error:
>2021-07-21T06:18:02.105Z\tINFO\t[composable.providers.docker]\tdocker/docker.go:43\tDocker provider skipped, unable to connect: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
2021-07-21T06:18:02.206Z\tINFO\tcapabilities/capabilities.go:59\tcapabilities file not found in /elastic-agent/capabilities.yml
2021-07-21T06:18:02.694Z\tINFO\t[composable.providers.docker]\tdocker/docker.go:43\tDocker provider skipped, unable to connect: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
2021-07-21T06:18:02.795Z\tINFO\tcapabilities/capabilities.go:59\tcapabilities file not found in /elastic-agent/capabilities.yml\nError: failed to start service (elastic-agent): \"service\" failed: exec: \"service\": executable file not found in $PATH"

It seems that the agent needs the "service" command to be present in the target machine, which makes it difficult to run in a plain container without an init system (such as systemd). We'd like to know from @elastic/apm-agent-devs if `systemd` is a strong requirement for the agent to be present in the target machine (vm, container).

michalpristas:
when installing an agent we are creating a new service using `github.com/kardianos/service` 
if this fails installation fails. 
[...]

Issue #1271, opened 2021-06-19
Elastic-agent install command fails on Debian + ARM in an inconsistent manner


mdelapenya:
The elastic-agent `install` command fails in an inconsistence manner when running on a Debian + ARM Docker image. With inconsistence we mean that we use the `install` command for all scenarios, and it only fails for the `Deploying the debian agent`

You can find an example CI build [here](https://beats-ci.elastic.co/blue/organizations/jenkins/e2e-tests%2Fe2e-testing-mbp/detail/master/995/pipeline)

We do not see errors at the PR level because we selected centos as the only image to be run on PRs. For that reason we observe this error on the nightly builds for [Fleet tests](https://beats-ci.elastic.co/job/e2e-tests/job/e2e-testing-fleet-daily-mbp/) 

> We offer a **private** Docker image for Debian with multi-platform support (AMD and ARM): https://container-library.elastic.co/r/observability-ci/debian-systemd, so to reproduce this, you should be able to run the tests on an ARM box, or in the CI

#### Log Error
[...]

PR #1341, opened 2021-07-20
chore(ci): be more specific about the underlying ARM OS

mdelapenya:
## What does this PR do?
It forces `centos-8-arm` as the label used by Jenkins to grab the CI workers when the tests need to be run on ARM.

## Why is it important?
We are seeing certain flakiness in one single scenario for Debian (See https://github.com/elastic/e2e-testing/issues/1271) and we'd like to narrow down the issue: because we have multiple ARM CI workers under the `arm` label, we want to always use the same.

## Checklist

## Author's Checklist

- [ ] The label does exist on Infra's catalog

## Related issues

- Relates https://github.com/elastic/e2e-testing/issues/1271
- Relates https://github.com/elastic/e2e-testing/issues/1340

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -316,8 +316,8 @@ def generateFunctionalTestStep(Map args = [:]){
   def workerLabels = "${platform} && immutable"
   if (platform == "arm64") {
     goArch = "arm64"
-    // ARM is not ready as immutable workers
-    workerLabels = "arm"
+    // let's use centos8 ARM as base VM, trying to mitigate the flakiness described in https://github.com/elastic/e2e-testing/issues/1271
+    workerLabels = "centos-8-aarch64"
   }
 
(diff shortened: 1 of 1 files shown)


v1v (approved):
Do you want to run this a few times a day? In case you'd like to add the cron option in the toplevel options section

cachedout (approved):

mdelapenya:
> Do you want to run this a few times a day? In case you'd like to add the cron option in the toplevel options section

I'd say that with the current frequency it's enough, but I see your point: we could maybe keep a PR open simply running the offending test suite. 🤔 Does it make sense, right?

v1v:
I thought this PR will remain open, that was the reason for my question, since it's merged then it was not valid :)

mdelapenya:
We can send a new one simplifying the test layout, including the proposed cron

PR #1361, opened 2021-07-22
Add kubernetes autodiscover scenario with inputs

jsoriano:
## What does this PR do?

Add a common scenario for collection of kubernetes logs.

Added to the autodiscover suite because is an overlapping use case, though this doesn't use autodiscover features.

## Why is it important?

This is the default configuration recommended for filebeat in kubernetes and is a pretty common use case.

It adds coverage for `add_kubernetes_metadata`.

## Checklist

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/e2e/_suites/kubernetes-autodiscover/features/filebeat.feature
+++ b/e2e/_suites/kubernetes-autodiscover/features/filebeat.feature
@@ -35,3 +35,10 @@ Scenario: Logs collection from a pod with an ephemeral container
     And "filebeat" collects events with "kubernetes.container.name:container-in-pod"
    When an ephemeral container is started in "a pod"
    Then "filebeat" collects events with "kubernetes.container.name:ephemeral-container"
+
+# Not really autodiscover, but this is a frequent use case in autodiscover-like
+# scenarios for logs collection.
+Scenario: Enrichment of normal input using add_kubernetes_metadata
+  Given "filebeat" is running with "input and add_kubernetes_metadata"
+   When "a pod" is deployed
+   Then "filebeat" collects events with "kubernetes.pod.name:a-pod"
--- a/internal/kubernetes/kubernetes.go
+++ b/internal/kubernetes/kubernetes.go
@@ -11,6 +11,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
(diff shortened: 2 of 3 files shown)


internal/kubernetes/kubernetes.go
```diff
@@ -122,8 +123,15 @@ func (c Cluster) Kubectl() Control {
...
 func (c Cluster) isAvailable(ctx context.Context) error {
-	_, err := c.Kubectl().Run(ctx, "api-versions")
-	return err
+	out, err := c.Kubectl().Run(ctx, "api-versions")
+	if err != nil {
+		return err
+	}
+	if len(strings.TrimSpace(out)) == 0 {
+		return fmt.Errorf("no api versions?")
+	}
```
jsoriano, line 132: Not related to this change, but I found that `kubectl api-versions` succeeds with an empty output if there is no context configured. I can move this to a different PR if desired.

ChrsMark (approved):
Great!

adam-stokes (approved):

e2e/_suites/kubernetes-autodiscover/features/filebeat.feature
```diff
@@ -35,3 +35,10 @@ Scenario: Logs collection from a pod with an ephemeral container
     And "filebeat" collects events with "kubernetes.container.name:container-in-pod"
    When an ephemeral container is started in "a pod"
    Then "filebeat" collects events with "kubernetes.container.name:ephemeral-container"
+
+# Not really autodiscover, but this is a frequent use case in autodiscover-like
+# scenarios for logs collection.
+Scenario: Enrichment of normal input using add_kubernetes_metadata
+  Given "filebeat" is running with "input and add_kubernetes_metadata"
```
mdelapenya, line 42: Sorry to jump-in late: what would happen if `"filebeat" is running with "input"` or `"filebeat" is running with "add_kubernetes_metadata"` (only of them)?

jsoriano: I am including this specific case of having both an input + `add_kubernetes_metadata` because it is included as default in the reference manifests for filebeat ([here](https://github.com/elastic/beats/blob/877d8bcd176b2f5d4efd2a81846a481b94798b49/deploy/kubernetes/filebeat-kubernetes.yaml#L12-L20)), and is a pretty common use case in general.

If `"filebeat" is running with "input"`, without `add_kubernetes_metadata` and without any autodiscover, then this is not a kubernetes use case, and would be covered by other suites.

If `"filebeat" is running with "add_kubernetes_metadata"` but without any input, then this will be a noop, the processor won't do anything. We could add a case with `add_kubernetes_metadata`, without any input, but with autodiscover, but this is a case we don't recommend, because autodiscover already makes the same metadata enrichment that  `add_kubernetes_metadata` does, so more resources are used to do the same thing again.

mdelapenya: Ok, understood. Then if suggest creating the second scenario, as it seems a feasible use case. The scenario could explicitly explain what you describe

jsoriano: PR to add this scenario: https://github.com/elastic/e2e-testing/pull/1385

PR #1393, opened 2021-07-28
fix(ci): wait for downstream jobs on scheduled jobs

### LINKED ISSUE #1376: Propagate build result in scheduled builds
We are usually creating pipelines (or MBP) that trigger other builds and we usually do not propagate the build result.

We’d see it convenient to propagate the result of the downstream job to the parent (upstream or caller), with the sole goal of narrowing down the monitors we’d like to create around the e2e tests: because the master branch is a helper pipeline triggered by many, it sometimes could be red, therefore I’d like to store the result of the nightly builds using this propagate and my monitors will only read the result of the scheduled nightly builds.

mdelapenya:
## What does this PR do?
It forces the scheduled pipelines to wait for the downstream job.

## Why is it important?
Setting `propagate: true` is not enough, as it seems Jenkins is not able to update the state of a build once it has finished. Adding `wait: true` we expect to achieve what we want.

## Checklist

## Related issues

- Completes #1377
- Closes #1376 

## Screenshots
Screenshot 2021-07-28 at 10 09 48

### DIFF
--- a/.ci/e2eKibana.groovy
+++ b/.ci/e2eKibana.groovy
@@ -207,7 +207,7 @@ def runE2ETests(String suite) {
   build(job: "${e2eTestsPipeline}",
     parameters: parameters,
     propagate: true,
-    wait: false
+    wait: true
   )
 
 /*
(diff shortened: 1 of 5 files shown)


v1v (approved):

mdelapenya:
Not adding `v7.13.x` because k8s-autodiscover test suite is not present in that branch

PR #1406, opened 2021-07-29
fix: check if the snapshot version contains a commit

### LINKED ISSUE #1405: Check if the beat version snapshot contains a git commit
I think we need to improve the automated bump with removing the calculation of the BeatVersionBase, which comes including the hash:

```golang
v, err := utils.GetElasticArtifactVersion(BeatVersionBase)
if err != nil {
	log.WithFields(log.Fields{
		"error":   err,
		"version": BeatVersionBase,
	}).Fatal("Failed to get Beat base version, aborting")
}
BeatVersionBase = v
```
A possible change could be to carry the fallback version between checks, replacing the `log.Fatal` with `log.Warn`:
```golang
v, err := utils.GetElasticArtifactVersion(BeatVersionBase)
if err != nil {
	log.WithFields(log.Fields{
		"error":   err,
		"version": BeatVersionBase,
	}).Warn("Failed to get Beat base version, aborting")
} else {
	BeatVersionBase = v
}
```

@adam-stokes wdyt?

_Originally posted by @mdelapenya in https://github.com/elastic/e2e-testing/issues/1391#issuecomment-888032893_

mdelapenya:
## What does this PR do?
When checking the version using Elastic's versions API, this PR adds a verification for the version to have a commit sha between the semver version and the SNAPSHOT suffix. If it does contain it, then the API is not looked up. To achieve it we added a helper method as an utility, where we use a regex that checks for `X.Y.Z-sha-SNAPSHOT`. We added unit tests verifying the new behaviour.

On the same hand, we noticed that the existing `IsCommit` method would return true for this hashed snapshots, so when checking Kibana version we used that method to change the Docker namespace for the image. With the previous implementation the method would return true, changing the namespace to `observability-ci`. For that reason we improved the regex in there to check for commits from the first position in the checked string.

## Why is it important?
The automation to bump the versions replaces the current version with i.e. `8.0.0-abcdef-SNAPSHOT`. That version was used to look up the versions API, and because it does not exist, the tests fails with a log.Fatal: because the stack (ES, Kibana, Fleet-Server) is considered a runtime dependency, we cannot run the tests without it. With this change, we simply use the version as it is, without the versions API (which is used to resolve version aliases).

## Checklist

## Author's Checklist

- [ ] regex masters, can you review the new regex? :pray:

```shell
# run unit tests
$ make unit-test

# run e2e tests with a hashed version
$ TAGS="apm_server" TIMEOUT_FACTOR=3 LOG_LEVEL=TRACE BEAT_VERSION="8.0.0-ae1e1c3b-SNAPSHOT" DEVELOPER_MODE=true ELASTIC_APM_ACTIVE=false make -C e2e/_suites/fleet functional-test

# run e2e tests with a fixed version without commit
$ TAGS="apm_server" TIMEOUT_FACTOR=3 LOG_LEVEL=TRACE BEAT_VERSION="8.0.0-SNAPSHOT" STACK_VERSION=8.0.0-SNAPSHOT DEVELOPER_MODE=true ELASTIC_APM_ACTIVE=false make -C e2e/_suites/fleet functional-test
```

## Related issues

- Closes #1405
- Unblocks any automated bump PR: #1391 #1403

### DIFF
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -234,8 +234,8 @@ func getGCPBucketCoordinates(fileName string, artifact string, version string) (
 
 // GetElasticArtifactVersion returns the current version:
 // 1. Elastic's artifact repository, building the JSON path query based
-// If the version is a PR, then it will return the version without checking the artifacts API
-// i.e. GetElasticArtifactVersion("$VERSION")
+// If the version is a SNAPSHOT including a commit, then it will directly use the version without checking the artifacts API
+// i.e. GetElasticArtifactVersion("$VERSION-abcdef-SNAPSHOT")
 func GetElasticArtifactVersion(version string) (string, error) {
 	cacheKey := fmt.Sprintf("https://artifacts-api.elastic.co/v1/versions/%s/?x-elastic-no-kpi=true", version)
 
@@ -245,6 +245,11 @@ func GetElasticArtifactVersion(version string) (string, error) {
 			"version": val,
 		}).Debug("Retrieving version from local cache")
 		return val, nil
+	}
+
+	if SnapshotHasCommit(version) {
+		elasticVersionsCache[cacheKey] = version
+		return version, nil
 	}
 
 	exp := GetExponentialBackOff(time.Minute)
@@ -589,7 +594,7 @@ func getBucketSearchNextPageParam(jsonParsed *gabs.Container) string {
 
 // IsCommit returns true if the string matches commit format
 func IsCommit(s string) bool {
-	re := regexp.MustCompile(`\b[0-9a-f]{5,40}\b`)
+	re := regexp.MustCompile(`^\b[0-9a-f]{5,40}\b`)
 
 	return re.MatchString(s)
 }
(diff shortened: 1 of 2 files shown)


kuisathaverat (approved):

internal/utils/utils.go
```diff
@@ -589,7 +594,7 @@ func getBucketSearchNextPageParam(jsonParsed *gabs.Container) string {
 
 // IsCommit returns true if the string matches commit format
 func IsCommit(s string) bool {
-	re := regexp.MustCompile(`\b[0-9a-f]{5,40}\b`)
+	re := regexp.MustCompile(`^\b[0-9a-f]{5,40}\b`)
```
mdelapenya, line 597: Before this change, it captured the sha in whatever position of the string, so `8.0.0-abcdef-SNAPSHOT` was matched. Therefore this block: https://github.com/elastic/e2e-testing/blob/master/e2e/_suites/fleet/ingest_manager_test.go#L153-L157, was incorrectly evaluated to true.

cachedout (approved):

PR #1472, opened 2021-08-19
chore: increase builds to keep

### LINKED ISSUE #1449: Increase the number of retained builds
The project only keeps the last 20 builds, and we'd need to retain a bigger number: 100? Because the main pipeline is a used as a helper, used as downstream for multiple sources (Beats PR/merges, nightly builds), in one day we loose the entire history.

mdelapenya:
## What does this PR do?
This PR mimics what the APM Agent Ruby has in its JJBB configuration for the build.

## Why is it important?
We want to keep as many build executions as possible, as the branches in the main MBP are used as downstream by multiple projects: nightly builds, Beats, and the development of the project itself.

## Checklist

## Related issues

- Closes #1449

### DIFF
--- a/.ci/jobs/e2e-testing-mbp.yml
+++ b/.ci/jobs/e2e-testing-mbp.yml
@@ -5,7 +5,11 @@
     description: Jenkins pipeline for the e2e-testing project
     view: E2E
     project-type: multibranch
-    days-to-keep: 21
+    logrotate:
+      daysToKeep: 30
+      numToKeep: 300
+    number-to-keep: 100
+    days-to-keep: 30
     script-path: .ci/Jenkinsfile
     scm:
       - github:
--- a/.ci/jobs/e2e-testing-mbp.yml
+++ b/.ci/jobs/e2e-testing-mbp.yml
@@ -7,7 +7,7 @@
     project-type: multibranch
     logrotate:
       daysToKeep: 30
-      numToKeep: 300
+      numToKeep: 100
     number-to-keep: 100
     days-to-keep: 30
     script-path: .ci/Jenkinsfile


cachedout:
Given that each of these builds is generating ~45MB in artifacts, I think this should keep around 1.5GB of data around at any given time.

I'm sure that's probably fine but I think we should add @leathekd as a reviewer just to be sure that's fine from the CI side of things.

cachedout (approved):

.ci/jobs/e2e-testing-mbp.yml
```diff
@@ -5,7 +5,11 @@
     description: Jenkins pipeline for the e2e-testing project
     view: E2E
     project-type: multibranch
-    days-to-keep: 21
+    logrotate:
+      daysToKeep: 30
+      numToKeep: 300
+    number-to-keep: 100
```
leathekd, line 11: Did you want these to be the same?

leathekd (approved):
Left a comment about the actual number to keep (100 vs 300).  Other than that, this LGTM!

.ci/jobs/e2e-testing-mbp.yml
```diff
@@ -5,7 +5,11 @@
     description: Jenkins pipeline for the e2e-testing project
     view: E2E
     project-type: multibranch
-    days-to-keep: 21
+    logrotate:
+      daysToKeep: 30
+      numToKeep: 300
```
mdelapenya, line 10:
```suggestion
      numToKeep: 100
```

.ci/jobs/e2e-testing-mbp.yml
mdelapenya: Let's started with 100 first, thanks for taking a look!

v1v (approved):
+1

qq: If we ship the logs to Elasticsearch then this won't be needed right? 

The envision with the Jenkins Opentelemetry will be to support this, but it's still an experimental feature -> https://opentelemetry.io/status/#logging

mdelapenya:
> +1
> 
> qq: If we ship the logs to Elasticsearch then this won't be needed right?
> 
> The envision with the Jenkins Opentelemetry will be to support this, but it's still an experimental feature -> https://opentelemetry.io/status/#logging

At this moment we are not seeking only logs but test results and artifacts. Once everything is observed from the CI, and we have a nice UI to retrieve it, then it would be great to have them all in the Elasticsearch instance.

PR #1480, opened 2021-08-19
chore: grab stack logs

mdelapenya:
## What does this PR do?
It does three things for the tests running on Windows:

- wraps the execution of the remote stack (which runs on a Linux node) with a `filebeat` step, to be able to capture container logs and see them in Jenkins UI.
- removes the installatino of UNZIP, as we use powershell to extract the archive
- avoids hardcoding the test suite name when running the stack profile, which was set to Fleet.

## Why is it important?
We are seeing a consistent error for Fleet `install` command on Windows in the 7.13.x, and we'd need to collect the logs of fleet-server to identify why the command fails and hangs up without response, running out the default timeout for the operation (up to 10 mins).

## Checklist

## Related issues

- Relates https://github.com/elastic/beats/issues/27501


mdelapenya:
Merging!

PR #1502, opened 2021-08-25
feat: support elastic-package provider for Fleet test suite

### LINKED ISSUE #1520: Remove metricbeat test suite
We are willing to remove the metricbeat integrations test suite in the e2e-testing framework, and those are the main reasons:

- **Remove duplication**: the elastic-package tool already tests the integrations using the new format (as package integrations).
- **Legacy**: metricbeat integrations modules will be overlapped by the elastic-agent's integration packages, so using the new tool could be the new direction to take.
- **Completeness**: while the e2e-testing framework only supports a reduced set of modules, more specifically those having a `supported-versions.yml` file on Beats, the elastic-package runs unit and integration tests for all packages under the https://github.com/elastic/integrations repo.
- **Maintanability**: reduce the code we have in the e2e-testing repo, as no docker-compose provider will be needed at all. The Helm charts and the k8s autodiscover test suites leverage k8s and kind, and the fleet test suite will leverage the elastic-package (see https://github.com/elastic/e2e-testing/pull/1502).
- **Dog-fooding**: because of the above, leveraging the elastic-package tool will allow us to use a tool that is already used by the Data Collections team for testing their projects. Being an external user of the tool can help the maintainers covering different use cases.

cc/ @jsoriano @elastic/observablt-robots

mdelapenya:
## What does this PR do?
This PR comes with a lot of changes:

1. It adds a tools.go file and package to define external Go dependencies, pinning them as a Go module. Therefore, we will be able to run them with `go run github.com/elastic/DEPENDENCY`.  We are pinning elastic-package tool there. (adaa1e6c381087d9b161dc033841442ef2561bf5)
1. Enriched the existing methods in the service managers to pass a Go context as first argument. This enables creating APM traces in all missing methods: PreBootstrap, Logs, Remove, Start, Stop (9cc6ee6032346391f1c957f0588065f8ce81c827 > 7f16db0d7fca331a5e0a38537330821445961be0 > b05abfa3ea8b76b8f557458dc3d056bbcc02be61 > 4dcda56c5f27683dc2d402107bdfba88694fa620 > da530aa66e36f7c0c25b3a9409f7253cbb6dcf04)
1. Removed a dependency between common package and deploy. This is needed to not cause cyclic dependencies. (c2d623434aa872f0b7d80e19eeb7956d63656cbc)
1. Simplified how the container name is calculated for the elastic-agent Docker image. We are going to Inspect the container using the Docker engine instead of hardcoding the name (cd5a6b2732c372e96422833a42b2195c776eb70c and 323e3107779fa309150ffb8180fb460372feb593)
1. We are going to remove the elastic-agent container after each scenario. (5a67c3156917c490eb211c1c1fe4b5fa37546caf and 219214c58a6ba9ae7ccff3887d7e82e49ab3bd93).
1. [Boyscout rule] we are moving sync integrations code from the Cobra command to the config package (6616b30afdbbf3f46ff507e29a1b45c1c77ae2a7)
1. [Boyscout rule] Thanks to the above, we will always sync metricbeat's integration modules, executed in the BeforeSuite phase of the test execution. (35ecc02308dd8284338884bc75798416f7c1f392). This can be skipped with the SKIP_PULL=true variable.
1. Added an `elastic-package` provider (59a0079b09a717a33aa31ffc75e53955a4ea917f). This new provider, which is only available for Fleet (failing in other case), will:
	- bootstrap the run-time dependencies using the `stack up` command, passing all available services but the elastic-agent, which we want to control ourselves under each test scenario.
	- add elastic-agent services using a vanilla created with Testcontainers (other type of services will cause the method to fail). We will attach the container to the very same network as in the stack provisioned by the `elastic-package` tool, so that communications can be resolved via hostnames, and we will provide an unique container name per test suite, using an UUID.
	- destroy the run-time dependencies using the `stack down` command.
	- basic container operations (copy files to the service, execute commands inside, inspect, getting logs, remove, start and stop) will use same approach than in the Docker provider, as we are using a Generic container. **TO-DO**: this will need a refector to reuse code.
1. Modify the CI descriptor so that platforms are objects instead of array elements. This is needed because we would like to run different providers per platform: i.e. `fleet_mode_agent` parallel branch is going to be run on windows and on ubuntu. (c686c9d5c588a0d5dcb27d709c2de7e7d5a92b43)
1. Select which platforms are going to use the `elastic-package` provider. (e577cecf1630b3b300688e3fab9c38cac0a90d0b). At this moment, we want to try out the new provider in two parallel branches: `fleet_mode_agent` and `stand_alone_agent`, both on linux, not on windows.
1. Remove the metricbeat's integration modules test suite. As described in #1520, we are duplicating efforts for the integrations. Removing this test suite will allow us to remove a lot of code (in a follow-up) related to manage the runtime dependencies with docker-compose (will use the elastic-package provided by this PR). Besides that, we are removing the nightly builds for the metricbeat test suite. (084b0d739cdb106f04523d815e9e2a98233ca71f, 0d5771d0434f7d9b7db8a84af53b23a1b200d833 and 0a82115c8cba7ea7ff74717c16da8a09a8cfd16d)

## Why is it important?
We'd like to use a proven tool, which is maintained by the Integrations team, instead of maintaining our own codebase. We analysed both projects and found a lot of similarities in terms of provisioning run-time dependencies (basically the stack). For that, we want to keep one single source of truth when talking about dependencies.

## Checklist

## Author's Checklist

- [ ]

```shell
TAGS="fleet_mode_agent" TIMEOUT_FACTOR=3 LOG_LEVEL=TRACE DEVELOPER_MODE=true ELASTIC_APM_ACTIVE=false PROVIDER=elastic-package make -C e2e/_suites/fleet functional-test
TAGS="stand_alone_agent" TIMEOUT_FACTOR=3 LOG_LEVEL=TRACE DEVELOPER_MODE=true ELASTIC_APM_ACTIVE=false PROVIDER=elastic-package make -C e2e/_suites/fleet functional-test
```

## Related issues

 - Closes #1520 

## Follow-ups
- ~~Expand the usage of the `elastic-package` provider to the rest of Fleet test suites, once this PR passes.~~ Done in https://github.com/elastic/e2e-testing/pull/1502/commits/c185b7d675c9f9e0abd063ee5d1d8e87b1b75250
- Remove Metricbeat test suite in favour of the system tests for the Integrations repository, already ran with the elastic-package tool: https://github.com/elastic/e2e-testing/issues/1520


.go-version
```diff
@@ -1 +1 @@
-1.15.9
+1.16.3
```
mdelapenya, line 1: We found an ugly error while downloading a dependency of Go's cobra on CI, with Go 1.15: See https://github.com/spf13/viper/issues/1161

internal/deploy/docker.go
```diff
@@ -226,14 +249,26 @@ func (c *dockerDeploymentManifest) Remove(profile ServiceRequest, services []Ser
...
-func (c *dockerDeploymentManifest) Start(service ServiceRequest) error {
+func (c *dockerDeploymentManifest) Start(ctx context.Context, service ServiceRequest) error {
+	span, _ := apm.StartSpanOptions(ctx, "Starting service from compose deployment", "docker-compose.service.start", apm.SpanOptions{
+		Parent: apm.SpanFromContext(ctx).TraceContext(),
+	})
+	span.Context.SetLabel("service", service)
+	defer span.End()
+
 	manifest, _ := c.Inspect(context.Background(), service)
-	_, err := shell.Execute(c.Context, ".", "docker", "start", manifest.Name)
```
mdelapenya, line 231: @adam-stokes do you think we still need to store a context in the dockerDeploymentManifest struct?

internal/deploy/docker_client.go
```diff
@@ -311,6 +311,10 @@ func InspectContainer(service ServiceRequest) (*types.ContainerJSON, error) {
 		}).Fatal("Cannot list containers")
 	}
 
+	if len(containers) == 0 {
```
mdelapenya, line 314: We have a lot of usages for the `.Inspect` method where we simply discard the raised error. We need to think about changing that pattern

internal/deploy/elastic-package.go
```diff
@@ -0,0 +1,331 @@
...
+			"error":   err,
+			"service": service.Name,
+		}).Error("Could not retrieve Elastic Agent logs")
+
+		return err
+	}
+	return nil
+}
+
+// PreBootstrap sets up environment with docker compose
```
mdelapenya, line 211:
```suggestion
// PreBootstrap sets up environment with the elastic-package tool
```

internal/deploy/elastic-package.go
```diff
@@ -0,0 +1,331 @@
...
+	if srv.Flavour == "centos" || srv.Flavour == "debian" {
+		imageNamespace = fmt.Sprintf("elastic-agent-%s", srv.Flavour)
+		// use observability's systemd base images to install the elastic-agent on them
+		img = "docker.elastic.co/observability-ci/" + srv.Flavour + "-systemd:latest"
+		entrypoint = []string{"/sbin/init"}
+		bindMounts = map[string]string{
+			"/sys/fs/cgroup": "/sys/fs/cgroup",
+		}
+		privileged = true
+	} else if srv.Flavour == "cloud" {
```
mdelapenya, line 291: I'm adding cloud flavour, but I think we can remove it, including the descriptors in the config dir

cli/config/config.go
```diff
@@ -180,6 +182,43 @@ func PutServiceVariantEnvironment(env map[string]string, service string, service
...
+	// BeatsRepo default object representing Beats project
+	var BeatsRepo = git.ProjectBuilder.
+		WithBaseWorkspace(path.Join(workspace, "git")).
+		WithDomain("github.com").
+		WithName("beats").
+		WithRemote(remote).
+		Build()
+
+	if deleteRepository {
+		repoDir := path.Join(workspace, "git", BeatsRepo.Name)
```
adam-stokes, line 198: should this use `filepath.Join`?

cli/config/config.go
```diff
@@ -180,6 +182,43 @@ func PutServiceVariantEnvironment(env map[string]string, service string, service
...
+		}).Trace("Removing repository")
+		os.RemoveAll(repoDir)
+		log.WithFields(log.Fields{
+			"path": repoDir,
+		}).Trace("Repository removed")
+	}
+
+	git.Clone(BeatsRepo)
+
+	pattern := path.Join(
```
adam-stokes, line 211: This one as well for `filepath.Join`?

cli/config/config.go
```diff
@@ -180,6 +182,43 @@ func PutServiceVariantEnvironment(env map[string]string, service string, service
...
+		BeatsRepo.GetWorkspace(), "metricbeat", "module", "*", "_meta", "supported-versions.yml")
+	copyIntegrationsComposeFiles(BeatsRepo, pattern, workspace)
+
+	xPackPattern := path.Join(
```
adam-stokes, line 215: Same for `filepath.Join`?

cli/config/config.go
```diff
@@ -180,6 +182,43 @@ func PutServiceVariantEnvironment(env map[string]string, service string, service
...
+		BeatsRepo.GetWorkspace(), "x-pack", "metricbeat", "module", "*", "_meta", "supported-versions.yml")
+	copyIntegrationsComposeFiles(BeatsRepo, xPackPattern, workspace)
+
+	return nil
+}
+
```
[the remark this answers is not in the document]
adam-stokes: @mdelapenya I made some additional comments where I saw `path.Join`

adam-stokes (approved):

mdelapenya:
@adam-stokes I think this one is ready to go! If you agree, let's merge it and ...

- get rid of metricbeat test suite!
- create an issue to remove the compose-related provisioning code
- start using elastic-package, opening issues if/when needed

Issue #1445, opened 2021-08-05
make our process inspection cross platform


adam-stokes:
Currently we utilize pgrep to check process state, this is only avaiable on mac and linux. We need a portable way to check process state across *nix, macOS, Windows.

The current implementation is here: https://github.com/elastic/e2e-testing/blob/master/e2e/_suites/fleet/world.go#L85-L265

To make this work we need to move this implementation into the installer internal code so that when we attach an installer we can easily inspect processes based on the underlying operating system.

@IgorGuz2000 Is this something you want to try and tackle?

PR #1519, opened 2021-09-03
Move OS specific commands into actions

adam-stokes:
This adds support for what I am calling ServiceOperation. We are moving
specific executions into this namespace so that depending on the action we want
to run it will do the right thing on the underlying OS.

This is part 1 of the work to pave the way for supporting Windows with our first
action on checking process state.

## Why is it important?

Further expands our ability to easily integrate OS specific functionality into our testsuite

## Checklist

```
OP_LOG_LEVEL=DEBUG go test -timeout 60m -v --godog.tags='@fleet_mode_agent && @install'
```

## Related issues

- Relates #1445 

Once this PR is merged, a followup PR will be sent with the Windows specific action along with enabling of more tests that require this


internal/deploy/base.go
```diff
@@ -44,6 +44,11 @@ type ServiceOperator interface {
 	Uninstall(ctx context.Context) error
 }
 
+// ServiceOperatorAction represents an action that can be run within a ServiceOperator
+type ServiceOperatorAction interface {
```
mdelapenya, line 48: Wdyt naming it as `ServiceOperation`?

adam-stokes: That works for me! Naming things is hard ,😄

internal/action/base.go
```diff
@@ -0,0 +1,40 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package action
```
mdelapenya, line 5: I'm thinking about packages and names, and I'd like to discuss it a little bit further: I can see an action as an intrinsic part of the deployment operator, for that I'd see it with good eyes to include the ServiceOperation in the `deploy` package.

Besides that, because this is very similar to the Command pattern, I'd rename the `Attach` method to `Run()`, `Exec()` or `Execute()`, with no parameters. All the input parameters could be part of the state of the disposable ServiceOperation as attributes in the struct on creation.

internal/action/process.go
```diff
@@ -0,0 +1,237 @@
...
+func AttachActionWaitProcess(deploy deploy.Deployment, service deploy.ServiceRequest, actionOpts map[string]string) deploy.ServiceOperatorAction {
+	return &actionWaitProcess{
+		service: service,
+		deploy:  deploy,
+		opts:    actionOpts,
+	}
+}
+
+// Run executes the command
+func (a *actionWaitProcess) Run(ctx context.Context) (string, error) {
```
mdelapenya, line 38: Ah got the Command pattern here! :) Forget about this part in my comment above! 👏  

>Besides that, because this is very similar to the Command pattern, I'd rename the Attach method to Run(), Exec() or Execute(), with no parameters. All the input parameters could be part of the state of the disposable ServiceOperation as attributes in the struct on creation.

internal/action/process.go
```diff
@@ -0,0 +1,237 @@
...
+	mustBePresent := false
+	if a.opts["desiredState"] == "started" {
+		mustBePresent = true
+	}
+	retryCount := 1
+
+	processStatus := func() error {
+		occurrences, _ := strconv.Atoi(a.opts["occurrences"])
+		log.WithFields(log.Fields{
+			"desiredState": a.opts["desiredState"],
```
mdelapenya, line 51: I'd expose here a struct for the process options, instead of a map, so that the consumer is able to discover what to pass to the process, instead of a map of string|string.

internal/action/base.go
```diff
@@ -0,0 +1,43 @@
...
+	})
+	defer span.End()
+
+	log.WithFields(log.Fields{
+		"service": service,
+		"action":  action,
+	}).Trace("Attaching action for service")
+
+	switch action {
+	case "wait-for-process":
```
mdelapenya, line 29: Should we extract that string to (list of) a constant(s)?

internal/action/base.go
```diff
@@ -0,0 +1,43 @@
...
+		newActionOpts, ok := actionOpts.(ProcessAction)
+		if !ok {
+			log.Fatal("Unable to cast to action options to ProcessAction")
```
mdelapenya, line 32: Is it OK to fail here? If not, we still want to return a custom error

internal/action/base.go
```diff
@@ -0,0 +1,43 @@
...
+		}
+		if runtime.GOOS == "windows" {
+			attachAction := AttachActionWaitProcessWin(deploy, service, newActionOpts)
+			return attachAction, nil
+		}
+		attachAction := AttachActionWaitProcess(deploy, service, newActionOpts)
+		return attachAction, nil
+	}
+
```
mdelapenya, line 41: We could log.Warn/Info that NOOP is executed here for the `action` passed

mdelapenya (approved):
Apart from my comments, LGTM

internal/action/base.go
adam-stokes: I think we want to fail here, if the type assertion fails then we either don't know what the correct action should be or the caller is passing in an invalid set of options

PR #1638, opened 2021-10-04
Permission change on updating and adding integration

IgorGuz2000:
## What does this PR do?
This PR is testing permission change on adding or updating integrations

## Checklist

## Author's Checklist

- [ ]

## Related issues

-


e2e/_suites/fleet/features/permission_output_change.feature
```diff
@@ -0,0 +1,36 @@
...
+Scenarios for Permission Change
+
+@adding-integration-change-permission
+Scenario Outline: Adding the Linux Integration to an Agent ...
+  Given a "<os>" agent is deployed to Fleet with "tar" installer
+    And the agent is listed in Fleet as "online"
+    And the agent get Default Api Key
+  When the "Linux" integration is "added" in the policy
+    And a Linux data stream exists with some data
+  Then verify that Default Api Key is "different"
```
mdelapenya, line 12: I'd prefer this style
```suggestion
  Then the Default Api Key is "different"
```
OTOH, what does "different" mean? Is it a Fleet concept by itself?

e2e/_suites/fleet/fleet.go
```diff
@@ -1300,16 +1344,21 @@ func (fts *FleetTestSuite) thePolicyIsUpdatedToHaveSystemSet(name string, set st
 	if name != "log" {
 		condition = true
 	}
+	if name == "linux/metrics" {
```
mdelapenya, line 1347: I think you are looking for a `if/else if` case here instead of multiple `if` without `else`: 

```golang
	var condition = false
	var metrics = "system"
	var file = "/metrics.json"

	if name == "linux/metrics" {
		condition = true
		file = "/linux_metrics.json"
		metrics = "linux"
	} else if name != "system/metrics" && name != "logfile" && name != "log" {
		condition = true
	}
```
Is there any other possible value for name? If not, I'd remove the above else if, reverting the logic:

```golang
	var condition = true
	var metrics = "system"
	var file = "/metrics.json"

	if name == "linux/metrics" {
		file = "/linux_metrics.json"
		metrics = "linux"
	} else if name == "system/metrics" || name == "logfile" || name = "log" {
		condition = false
	}
```

e2e/_suites/fleet/fleet.go
```diff
@@ -1300,16 +1344,21 @@ func (fts *FleetTestSuite) thePolicyIsUpdatedToHaveSystemSet(name string, set st
...
+		condition = true
+		file = "/linux_metrics.json"
+		metrics = "linux"
+	}
 
 	if condition != true {
```
mdelapenya, line 1353: I'd move this evaluation to the top of the method, to fail fast. Therefore, The above condition variable would not be needed.

e2e/_suites/fleet/fleet.go
IgorGuz2000: They all different values and can be used separate

mdelapenya (changes requested):

mdelapenya:
@IgorGuz2000 I added https://github.com/elastic/e2e-testing/pull/1638/commits/3478c042e3bae5383e172b43e78c4ece366c7231, and will merge this once the CI passes.

Thanks for this fabulous contribution!!

PR #1693, opened 2021-10-20
fix: run heartbeat not as uid 1000

mdelapenya:
## What does this PR do?
In the k8s-autodiscover test suite, it sets an env var in heartbeat deployment to not hearbeat process as 1000, returning back to use root.

## Why is it important?
I think it's better explained here: https://github.com/elastic/beats/issues/28570

## Checklist

## Author's Checklist

- [ ] @jsoriano, what are the implication of running this pod as root again?

```shell
kind delete clusters --all && \
TAGS="heartbeat" TIMEOUT_FACTOR=3 LOG_LEVEL=TRACE DEVELOPER_MODE=true \
GITHUB_CHECK_SHA1=81c38fc4c009348d57c92ae85920aed35297a89e \
ELASTIC_APM_ACTIVE=false make -C e2e/_suites/kubernetes-autodiscover functional-test
```

## Related issues

- Relates https://github.com/elastic/beats/issues/28570

### DIFF
--- a/e2e/_suites/kubernetes-autodiscover/testdata/templates/heartbeat.yml.tmpl
+++ b/e2e/_suites/kubernetes-autodiscover/testdata/templates/heartbeat.yml.tmpl
@@ -70,6 +70,8 @@ spec:
           "-e",
         ]
         env:
+        - name: BEAT_SETUID_AS
+          value: ""
         - name: NODE_NAME
           valueFrom:
             fieldRef:


andrewvc:
This is fine as a short term fix, but since the root cause is that heartbeat/uid1000 doesn't have perms to read /etc/heartbeat, the better fix would be changing perms there.

PR #1743, opened 2021-10-29
fix: run local package registry with docker provider

mdelapenya:
## What does this PR do?
It adds the package-registry to the compose file used by the Docker provider, copying the configurations of the elastic-package profiles to current kibana configs.

## Why is it important?
The elastic-package provider is starting the registry locally, but we removed it from our own implementation in #707.

In the ARM flavour of the test suite, where we cannot use elastic-package, we need to add it back, to bypass the issues we found for the staging/snapshot/production package registry and kibana's recent breaking changes, where no packages are retrieved for 8.0.

For that reason, we need the local package registry in the Docker compose.

## Checklist

## Related issues

- Relates #1738

### DIFF
--- a/cli/config/compose/profiles/fleet/default/package-registry.config.yml
+++ b/cli/config/compose/profiles/fleet/default/package-registry.config.yml
@@ -1,0 +1,5 @@
+package_paths:
+  - /packages/development
+  - /packages/production
+  - /packages/staging
+  - /packages/snapshot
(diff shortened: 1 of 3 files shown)


cachedout (approved):

mdelapenya:
I've just noticed that on ARM, the package-registry does not have a Docker image, so the package-registry container that is started on the ARM test suite never starts because it downloads the AMD image, which fails to start on an ARM worker.

For that reason, any test suite using the local package registry (see kibana configuration) will fail

Issue #1737, opened 2021-10-28
Use elastic-package again


mdelapenya:
In #1736 we moved back to docker-compose, but we'd like to continue using the tool

mdelapenya:
Superseded by https://github.com/elastic/e2e-testing/issues/1985

PR #1753, opened 2021-11-02
fix: use staging environment for EPR

mdelapenya:
## What does this PR do?
It removes the package registry service from the startup list when using the elastic-package provider.

Finally, we are reverting #1750, as it added a local pacakge registry to the compose provider, which is wrong for the ARM flavour of the test suite: there is no ARM Docker image for the package registry.

## Why is it important?
Fixes the ARM test suite in 7.15, also adding consistency across branches about what package-registry to use.

## Checklist

## Related issues

- Relates #1737


mdelapenya:
Merging, as ARM test suite is green again

PR #1776, opened 2021-11-04
[mergify] fix redundancy with the backport labels

v1v:


v1v:
I merged directly since changes were harmless and already tested in other repository

PR #1914, opened 2021-12-16
fix: add resiliency when an integration cannot be added to a policy

mdelapenya:
## What does this PR do?
It does two things: 

1. changes log.Fatal to log.Error on failures when adding an integration to a policy
1. adds a retry with maxTimeout option to the addIntegrationToPolicy method, so that we are covered on errors in the staging environment of the package-registry

```
[2021-12-16T05:41:55.069Z] FATA[2021-12-16T05:41:53Z] Unable to add integration to policy           err="could not add package to policy; API status code = 500; response body = {\"statusCode\":500,\"error\":\"Internal Server Error\",\"message\":\"'502 Bad Gateway' error response from package registry at https://epr-staging.elastic.co/package/endpoint/1.4.0\"}" packageDS="{ endpoint-50fff084-9495-4633-b1a7-0a817b61d6bc Endpoint Security default cdd3d370-5e32-11ec-9ece-534be1f506d0 true  [] { endpoint Endpoint Security 1.4.0}}"

```

## Why is it important?
Resiliency in the pipeline execution

## Checklist

### DIFF
--- a/internal/kibana/integrations.go
+++ b/internal/kibana/integrations.go
@@ -60,6 +60,8 @@ func (c *Client) AddIntegrationToPolicy(ctx context.Context, packageDS PackageDa
 			}).Warn("Could not add package to policy. Retrying")
 
 			retryCount++
+
+			return err			
 		}
 
 		if statusCode != 200 {
--- a/internal/kibana/integrations.go
+++ b/internal/kibana/integrations.go
@@ -5,8 +5,11 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/Jeffail/gabs/v2"
+	"github.com/cenkalti/backoff/v4"
+	"github.com/elastic/e2e-testing/internal/utils"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"go.elastic.co/apm"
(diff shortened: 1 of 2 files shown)


internal/kibana/integrations.go
```diff
@@ -22,24 +25,65 @@ type IntegrationPackage struct {
...
+		reqBody, err := json.Marshal(packageDS)
+		if err != nil {
+			log.WithFields(log.Fields{
+				"elapsedTime": exp.GetElapsedTime(),
+				"err":         err,
+				"package":     packageDS,
+				"retry":       retryCount,
+			}).Warn("Could not convert policy-package (request) to JSON. Retrying")
+
+			retryCount++
```
cachedout, line 48: This is probably just me, but I don't understand how the retry logic is working here. Is this variable staying in scope and somehow being consumed by `backoff.Retry`? In looking at the documentation, I can't see any mention of this behavior so I wonder if perhaps https://pkg.go.dev/github.com/cenkalti/backoff/v4#WithMaxRetries should be used instead?

(Haven't had much coffee yet so possible that I'm missing something obvious.)

internal/kibana/integrations.go
```diff
@@ -22,24 +25,65 @@ type IntegrationPackage struct {
...
+		statusCode, respBody, err := c.post(ctx, fmt.Sprintf("%s/package_policies", FleetAPI), reqBody)
+		if err != nil {
+			log.WithFields(log.Fields{
+				"elapsedTime": exp.GetElapsedTime(),
+				"err":         err,
+				"package":     packageDS,
+				"retry":       retryCount,
+			}).Warn("Could not add package to policy. Retrying")
+
+			retryCount++
```
mdelapenya, line 62:
```suggestion
			retryCount++

			return err			
```

internal/kibana/integrations.go
mdelapenya: Mmm, this is interesting because I hadn't explored that part of the API yet... The retry variable here is more or less cosmetic, as it's used to provide feedback in logs about what's the current state of the retry function. Apart from that, we use an ExponentialBackoff, which is another type: https://pkg.go.dev/github.com/cenkalti/backoff/v4#ExponentialBackOff, adding a non-linear wait time between following calls.

cachedout: It's interesting to me that it works at all like that. I would have thought that the variable went out of scope after each invocation of the retry. There must be something fancy happening in the retry library to preserve it. Thanks for the explanation!

mdelapenya: Because it's declared outside of the retry function, its value is preserved as a "global-to-the-function"

adam-stokes (approved):

PR #1917, opened 2021-12-16
ci: use withAPMEnv

v1v:
## What does this PR do?

Use the APM specific endpoint rather than the one provided by the Jenkins OpenTelemetry plugin.

## Why is it important?

The Jenkins OTEL plugin uses the OpenTelemetry Collector, therefore it's OTEL specific, while in the past it was APM Server (which allows dual protocols: elastic apm specific and otel). 

Let's fallback to the APM specific approach.

## Follow ups

- If the instrumentation should be elastic specific or otel?

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -483,7 +483,7 @@ def generateFunctionalTestStep(Map args = [:]){
           // This step will help to send the APM traces to the
           // withOtelEnv is the one that uses the APM service defined by the Otel Jenkins plugin.
           // withAPMEnv uses Vault to prepare the context.
-					// IMPORTANT: withAPMEnv is now the one in used since withOtelEnv uses a specific Opentelemetry Collector at the moment.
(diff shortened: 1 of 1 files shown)


.ci/Jenkinsfile
```diff
@@ -481,9 +481,11 @@ def generateFunctionalTestStep(Map args = [:]){
         unstash 'sourceEnvModified'
         withEnv(envContext) {
           // This step will help to send the APM traces to the
-          // APM service defined by the Otel Jenkins plugin.
+          // withOtelEnv is the one that uses the APM service defined by the Otel Jenkins plugin.
+          // withAPMEnv uses Vault to prepare the context.
+					// IMPORTANT: withAPMEnv is now the one in used since withOtelEnv uses a specific Opentelemetry Collector at the moment.
```
v1v, line 486:
```suggestion
          // IMPORTANT: withAPMEnv is now the one in used since withOtelEnv uses a specific Opentelemetry Collector at the moment.
```

mdelapenya (approved):

PR #1807, opened 2021-11-16
chore: stabilise new test scenarios

mdelapenya:
## What does this PR do?
This PR removes the `Preconfigured policies` test suite from the CI descriptor that drives the parallel execution on Jenkins. Therefore, no parallel stage will be created to run the tests tagged with `@preconfigured_policies`. 

Besides that, I'm refactoring the code for the `verifyDefaultAPIKey` scenario method (`^the default API key has "([^"]*)"$`):
1. simplify if/else if logic
2. simplify logs
3. simplify error handling
4. reset the state of the DefaultAPIKey variable after test, to discard issues with incorrupt state across test scenarios.

## Why is it important?
We're skipping this test suite on CI until further investigation resolving why it fails on CI in a consistent manner, but passes locally

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have run the Unit tests (`make unit-test`), and they are passing locally
- [ ] I have run the End-2-End tests for the suite I'm working on, and they are passing locally
- [ ] I have noticed new Go dependencies (run `make notice` in the proper directory)

## Author's Checklist

- [ ] @nchaulet, we are still investigating. Please let us know if you had any other discovery

## Related issues

- Relates #1791

PR #1940, opened 2021-12-22
chore: restore permissions change tests

mdelapenya:
## What does this PR do?
It restores the permissions change scenarios, using the YAML file that drives the parallel execution at the CI level.

In the process:
- fixed the scenario name for the permissions_change feature file
- we are adding a log trace before and after every scenario, adding the name of the scenario
- we are adding a log trace before and after every scenario step, adding the name of the step
> This will allow us identify where in the scenario it failed

- rename the Inspect method of the Deployer struct to GetServiceManifest, to provide a better name
- add a log with the service manifest in all deployer structs
> we have detected that no logs are produced in this call:

```golang
func (fts *FleetTestSuite) getAgentDefaultAPIKey() (string, error) {
	agentService := deploy.NewServiceRequest(common.ElasticAgentServiceName)
	manifest, _ := fts.deployer.GetServiceManifest(fts.currentContext, agentService)
	agent, err := fts.kibanaClient.GetAgentByHostname(fts.currentContext, manifest.Hostname)
	if err != nil {
		return "", err
	}
	return agent.DefaultApiKey, nil
}
``` 

> I've observed that there test failure happens here, so it probably fails when retrieving the service manifest (the Kibana call does have logs, so it must be before that)

## Why is it important?
We skipped those tests while running on containers because they demonstrated certain flakiness, and we want to verify if running on real VMs restores the expected behaviours.

## Checklist

## Related issues

- Relates #1807


v1v (approved):

mdelapenya:
It's very weird:

Screenshot 2021-12-22 at 18 02 25

```
2021-12-22T16:11:19.308Z] TRAC[2021-12-22T16:11:19Z] Before Fleet scenario step: the default API key has "not changed" 
[2021-12-22T16:11:19.308Z] TRAC[2021-12-22T16:11:19Z] After Fleet scenario: Updating the Integration on an Agent not changing Default API key 
```

The beforeStep is logged, and right after that the afterScenario is logged without any inner logs, nor afterStep logs.

In this PR I've been adding log traces in the inner calls, but they are simply not called. I even thought that the qording in the feature file was wrong, but no, that's not the case.

mdelapenya:
Superseded by https://github.com/elastic/e2e-testing/pull/2872

PR #1943, opened 2021-12-22
feat: export Fetch&Download methods in the /pkg directory

### LINKED ISSUE #1942: Expose the functionality to download Beats binaries in a public package
Suggested by @narph, who is working in something similar, and @v1v pointed her to this repo, where we already download binaries from:
- a local path for Beats repo
- a snapshot produced by Beats CI, using Beats hash commit as directory in the CI snapshots bucket
- a released version of beats (even official snapshots), using the Elastic's artifact service.

mdelapenya:
- chore: copy internal's version files into the exported pkg public package
- chore: run unit tests from top-level goal
- chore: use public package instead of the internal one
- chore: remove old internal files for habdling downloads

## What does this PR do?
This PR exposes the methods that were available in the old `internal/versions.go` file, using the well-known `/pkg` public (see https://github.com/golang-standards/project-layout/tree/master/pkg for standard layouts).

The Makefile running the unit tests has been updated to include the new `pkg` package.

## Why is it important?
@narph needs to use similar code to download Beats binaries (from releases, CI snapshots, local paths), and we decided to expose this functionality from the E2E so that the team does not need to reinvent the wheel.

## Checklist

## Author's Checklist

- [x] Following standard layout for Go projects

## Related issues

- Closes #1942


pkg/downloads/versions.go
```diff
@@ -2,7 +2,7 @@
 // or more contributor license agreements. Licensed under the Elastic License;
 // you may not use this file except in compliance with the Elastic License.
 
-package internal
+package downloads
```
mdelapenya, line 5: @adam-stokes I could not find a better name for this package. Maybe you have something better in mind

mdelapenya:
@narph let's us know if this would be enough for you to consume this.

narph (approved):

adam-stokes (approved):

PR #1984, opened 2022-01-10
Update `fetchBeatsBinary` to be reused in elastic-agent-poc

narph:
- made method public to avoid any binary name construction
- added download path to download artifacts in the specified location


internal/utils/utils.go
```diff
@@ -54,7 +55,11 @@ func DownloadFile(url string) (string, error) {
 	}
 	defer tempFile.Close()
 
-	filepathFull := tempFile.Name()
+	if downloadPath != "" {
+		filepathFull = downloadPath
+	} else {
```
mdelapenya, line 60: Could you move https://github.com/elastic/e2e-testing/pull/1984/files#diff-a7cadd8b7616e655fd258138320b33eb96f7cce8e8663ec92651f49ce07df000R45-R46 right here, covered by the else branch? We will be creating the temp file if and only if the downloadPath is not empty

pkg/downloads/versions.go
```diff
@@ -83,7 +83,7 @@ func CheckPRVersion(version string, fallbackVersion string) string {
 // FetchElasticArtifact fetches an artifact from the right repository, returning binary name, path and error
 func FetchElasticArtifact(ctx context.Context, artifact string, version string, os string, arch string, extension string, isDocker bool, xpack bool) (string, string, error) {
 	binaryName := buildArtifactName(artifact, version, os, arch, extension, isDocker)
-	binaryPath, err := fetchBeatsBinary(ctx, binaryName, artifact, version, utils.TimeoutFactor, xpack)
+	binaryPath, err := FetchBeatsBinary(ctx, binaryName, artifact, version, utils.TimeoutFactor, xpack,"")
```
mdelapenya, line 86: I think we need to whitespace the arguments
```suggestion
	binaryPath, err := FetchBeatsBinary(ctx, binaryName, artifact, version, utils.TimeoutFactor, xpack, "")
```

pkg/downloads/versions_test.go
```diff
@@ -398,7 +398,7 @@ func TestFetchBeatsBinaryFromLocalPath(t *testing.T) {
 		artifactName := versionPrefix + "-x86_64.rpm"
 		expectedFilePath := path.Join(distributionsDir, artifactName)
 
-		downloadedFilePath, err := fetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true)
+		downloadedFilePath, err := FetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true,"")
```
mdelapenya, line 401:
```suggestion
		downloadedFilePath, err := FetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true, "")
```

pkg/downloads/versions_test.go
```diff
@@ -409,7 +409,7 @@ func TestFetchBeatsBinaryFromLocalPath(t *testing.T) {
 		artifactName := versionPrefix + "-aarch64.rpm"
 		expectedFilePath := path.Join(distributionsDir, artifactName)
 
-		downloadedFilePath, err := fetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true)
+		downloadedFilePath, err := FetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true,"")
```
mdelapenya, line 412:
```suggestion
		downloadedFilePath, err := FetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true, "")
```

pkg/downloads/versions_test.go
```diff
@@ -421,7 +421,7 @@ func TestFetchBeatsBinaryFromLocalPath(t *testing.T) {
 		artifactName := versionPrefix + "-amd64.deb"
 		expectedFilePath := path.Join(distributionsDir, artifactName)
 
-		downloadedFilePath, err := fetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true)
+		downloadedFilePath, err := FetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true,"")
```
mdelapenya, line 424:
```suggestion
		downloadedFilePath, err := FetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true, "")
```

pkg/downloads/versions_test.go
```diff
@@ -432,7 +432,7 @@ func TestFetchBeatsBinaryFromLocalPath(t *testing.T) {
 		artifactName := versionPrefix + "-arm64.deb"
 		expectedFilePath := path.Join(distributionsDir, artifactName)
 
-		downloadedFilePath, err := fetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true)
+		downloadedFilePath, err := FetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true,"")
```
mdelapenya, line 435:
```suggestion
		downloadedFilePath, err := FetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true, "")
```

pkg/downloads/versions_test.go
```diff
@@ -444,7 +444,7 @@ func TestFetchBeatsBinaryFromLocalPath(t *testing.T) {
 		artifactName := versionPrefix + "-linux-amd64.tar.gz"
 		expectedFilePath := path.Join(distributionsDir, artifactName)
 
-		downloadedFilePath, err := fetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true)
+		downloadedFilePath, err := FetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true,"")
```
mdelapenya, line 447:
```suggestion
		downloadedFilePath, err := FetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true, "")
```

pkg/downloads/versions_test.go
```diff
@@ -512,7 +512,7 @@ func TestFetchBeatsBinaryFromLocalPath(t *testing.T) {
 		artifactName := ubi8VersionPrefix + "-linux-arm64.docker.tar.gz"
 		expectedFilePath := path.Join(distributionsDir, artifactName)
 
-		downloadedFilePath, err := fetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true)
+		downloadedFilePath, err := FetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true,"")
```
mdelapenya, line 515:
```suggestion
		downloadedFilePath, err := FetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true, "")
```

pkg/downloads/versions_test.go
```diff
@@ -501,7 +501,7 @@ func TestFetchBeatsBinaryFromLocalPath(t *testing.T) {
 		artifactName := ubi8VersionPrefix + "-linux-amd64.docker.tar.gz"
 		expectedFilePath := path.Join(distributionsDir, artifactName)
 
-		downloadedFilePath, err := fetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true)
+		downloadedFilePath, err := FetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true,"")
```
mdelapenya, line 504:
```suggestion
		downloadedFilePath, err := FetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true, "")
```

pkg/downloads/versions_test.go
```diff
@@ -489,7 +489,7 @@ func TestFetchBeatsBinaryFromLocalPath(t *testing.T) {
 		artifactName := versionPrefix + "-linux-arm64.docker.tar.gz"
 		expectedFilePath := path.Join(distributionsDir, artifactName)
 
-		downloadedFilePath, err := fetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true)
+		downloadedFilePath, err := FetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true,"")
```
mdelapenya, line 492:
```suggestion
		downloadedFilePath, err := FetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true, "")
```

pkg/downloads/versions_test.go
```diff
@@ -478,7 +478,7 @@ func TestFetchBeatsBinaryFromLocalPath(t *testing.T) {
 		artifactName := versionPrefix + "-linux-amd64.docker.tar.gz"
 		expectedFilePath := path.Join(distributionsDir, artifactName)
 
-		downloadedFilePath, err := fetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true)
+		downloadedFilePath, err := FetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true,"")
```
mdelapenya, line 481:
```suggestion
		downloadedFilePath, err := FetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true, "")
```

pkg/downloads/versions_test.go
```diff
@@ -466,7 +466,7 @@ func TestFetchBeatsBinaryFromLocalPath(t *testing.T) {
 		artifactName := versionPrefix + "-linux-arm64.tar.gz"
 		expectedFilePath := path.Join(distributionsDir, artifactName)
 
-		downloadedFilePath, err := fetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true)
+		downloadedFilePath, err := FetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true,"")
```
mdelapenya, line 469:
```suggestion
		downloadedFilePath, err := FetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true, "")
```

internal/utils/utils.go
```diff
@@ -40,7 +40,8 @@ func GetArchitecture() string {
 // DownloadFile will download a url and store it in a temporary path.
 // It writes to the destination file as it downloads it, without
 // loading the entire file into memory.
-func DownloadFile(url string) (string, error) {
+func DownloadFile(url string, downloadPath string) (string, error) {
```
mdelapenya, line 43: I'm looking at the implementation and it's not clear to me if we need to return the path again, as it's passed in the download path. I can picture a new struct (i.e. `DownloadRequest`) with two attributes: url and downloadPath. The DownloadFile method would receive a pointer to an instance of this struct. If the downloadPath is empty, then create a temporary file (as it does now), populating the download path, otherwise use the value in the request. The method would still return the request's download path.

Wdyt?

adam-stokes: Yea if we're already passing in the downloadPath then a struct containing that information could be easily reused/referenced elsewhere.  I also tend to go with defined structs as it improves readability and intent

pkg/downloads/versions.go
adam-stokes: We may want to add a pre-commit check to ensure gofmt doesn't return anything other than 0

pkg/downloads/versions_test.go
```diff
@@ -387,7 +387,7 @@ func TestFetchBeatsBinaryFromLocalPath(t *testing.T) {
 		defer func() { BeatsLocalPath = "" }()
 		BeatsLocalPath = beatsDir
 
-		_, err := fetchBeatsBinary(ctx, "foo_fileName", artifact, version, utils.TimeoutFactor, true)
+		_, err := FetchBeatsBinary(ctx, "foo_fileName", artifact, version, utils.TimeoutFactor, true,"")
```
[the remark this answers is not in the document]
adam-stokes: Ditto here, have pre-commit check for formatting changes

internal/utils/utils.go
```diff
@@ -40,36 +46,40 @@ func GetArchitecture() string {
...
 // loading the entire file into memory.
-func DownloadFile(url string) (string, error) {
-	tempParentDir := filepath.Join(os.TempDir(), uuid.NewString())
-	internalio.MkdirAll(tempParentDir)
+func DownloadFile(downloadRequest *DownloadRequest) error {
+	var filePath string
+	if downloadRequest.DownloadPath == "" {
+		tempParentDir := filepath.Join(os.TempDir(), uuid.NewString())
+		internalio.MkdirAll(tempParentDir)
+		filePath = filepath.Join(tempParentDir, uuid.NewString())
```
mdelapenya, line 54: Should we update the downloadPath to the temp file? I wonder if consumers would like to use it after downloading
```suggestion
		filePath = filepath.Join(tempParentDir, uuid.NewString())
		downloadRequest.DownloadPath = filePath
```

pkg/downloads/versions.go
```diff
@@ -109,7 +109,7 @@ func GetCommitVersion(version string) string {
 // i.e. GetElasticArtifactURL("elastic-agent-$VERSION-$ARCH.deb", "elastic-agent", "$VERSION")
 // i.e. GetElasticArtifactURL("elastic-agent-$VERSION-x86_64.rpm", "elastic-agent","$VERSION")
 // i.e. GetElasticArtifactURL("elastic-agent-$VERSION-linux-$ARCH.tar.gz", "elastic-agent","$VERSION")
-func GetElasticArtifactURL(artifactName string, artifact string, version string) (string, error) {
+func GetElasticArtifactURL(artifactName string, artifact string, version string) (string, string, error) {
```
mdelapenya, line 112: I think we need to update the method description to:

```golang
// GetElasticArtifactURL returns the URL of a released artifact, which its full name is defined in the first argument,
// from Elastic's artifact repository, building the JSON path query based on the full name.
// It also returns the URL of the sha512 file of the released artifact.
```

pkg/downloads/versions.go
```diff
@@ -347,12 +348,12 @@ func buildArtifactName(artifact string, artifactVersion string, OS string, arch
...
 }
 
-// fetchBeatsBinary it downloads the binary and returns the location of the downloaded file
+// FetchBeatsBinary it downloads the binary and returns the location of the downloaded file
 // If the environment variable BEATS_LOCAL_PATH is set, then the artifact
 // to be used will be defined by the local snapshot produced by the local build.
 // Else, if the environment variable GITHUB_CHECK_SHA1 is set, then the artifact
 // to be downloaded will be defined by the snapshot produced by the Beats CI for that commit.
-func fetchBeatsBinary(ctx context.Context, artifactName string, artifact string, version string, timeoutFactor int, xpack bool) (string, error) {
+func FetchBeatsBinary(ctx context.Context, artifactName string, artifact string, version string, timeoutFactor int, xpack bool, downloadPath string, downloadSHAFIle bool) (string, error) {
```
mdelapenya, line 356: Typo
```suggestion
func FetchBeatsBinary(ctx context.Context, artifactName string, artifact string, version string, timeoutFactor int, xpack bool, downloadPath string, downloadSHAFile bool) (string, error) {
```

pkg/downloads/versions.go
```diff
@@ -422,24 +431,41 @@ func fetchBeatsBinary(ctx context.Context, artifactName string, artifact string,
...
+		bucket, prefix, object := getGCPBucketCoordinates(artifactName, artifact)
+
 		downloadURL, err = getObjectURLFromBucket(bucket, prefix, object, maxTimeout)
 		if err != nil {
 			return "", err
 		}
+		downloadLocation, err := handleDownload(downloadURL)
 
+		// check if sha file should be downloaded, else return
+		if downloadSHAFIle == false {
```
mdelapenya, line 445:
```suggestion
		if downloadSHAFile == false {
```

mdelapenya:
@narph how would you see delegating the download of the sha512 file to the handleDownload file? If it receives the new DownloadRequest struct instead of the URL, then it can internally handle the download of both files, the binary and its sha512. Then, we would be able to remove the duplication of the calls to the handleDownload in both scenarios: the artifactory and the bucket.

PR #2034, opened 2022-01-20
feat: add CentOS 8 support

mdelapenya:
- chore: rename platforms to include OS
- fix: include sles15 in Beats builds
- chore: expand coverage for SLES15
- chore: run k8s-autodiscover on Debian VMs
- chore: run helm test suite on Debian VMs
- chore: remove unused platforms
- feat: add centos8 support, running fleet test suite

## What does this PR do?
This PR adds CentOS 8 as a backend to run the tests, using the AMIs provided by the official Centos website: https://www.centos.org/download/aws-images/

We have also refactored the platforms YAML file to identify the platform with the OS name: i.e. `fleet_amd64 > Debian_amd64`, also removing duplicated entries.

Finally, we are expanding the feature coverage for SLES15, running all fleet test suites on those boxes.

## Why is it important?
Expand OS/Arch coverage

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have run the Unit tests (`make unit-test`), and they are passing locally
- [ ] I have run the End-2-End tests for the suite I'm working on, and they are passing locally
- [ ] I have noticed new Go dependencies (run `make notice` in the proper directory)

## Related issues

- Closes #2033

PR #2075, opened 2022-01-30
chore: always install tools in runners

mdelapenya:
## What does this PR do?
~~It removes the when clause for the installation of the dependant roles when provisioning the runners representing the target VMs~~

Update: it fixes a typo in the when expression, where we were using the `nodeLabel` variable instead of `suite`, also updating the value used for the when condition:

```diff
- kubernetes_autodiscover
+ kubernetes-autodiscover
```

## Why is it important?
We want to have reproducible builds, and having all tooling installed for each specific test suite is key, that¡s why we are installing the tools with the right when conditions.

## Checklist

### DIFF
--- a/.ci/ansible/playbook.yml
+++ b/.ci/ansible/playbook.yml
@@ -102,13 +102,9 @@
     - role: geerlingguy.docker
       docker_daemon_options:
         default-ulimit: ["nofile=1024000:1024000"]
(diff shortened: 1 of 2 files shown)


v1v (commented):
LGT, can you help to understand how the always install tools is affected by renaming the variable in the when condition, is that a typo?

mdelapenya:
> LGT, can you help to understand how the always install tools is affected by renaming the variable in the when condition, is that a typo?

Indeed. I detected a typo caused by #2034. I'm updating the description to represent the changes after sending follow-up commits

v1v (approved):

PR #2107, opened 2022-02-08
fix: update APT cache for both Debian and Ubuntu

mdelapenya:
## What does this PR do?
This PR makes a wider evaluation of variables to decide when to update APT's cache: instead of simply checking for Ansible Distribution == Debian, we are checking that the package manager is APT or the distribution is Debian OR Ubuntu

## Why is it important?
We noticed that there were situations in which the AWS machine failed to receive the Docker role because of this failure:

```
[2022-02-08T08:49:41.951Z] TASK [geerlingguy.docker : Ensure additional dependencies are installed (on Ubuntu < 20.04 and any other systems).] ***
[2022-02-08T08:49:42.901Z] fatal: [3.14.83.35]: FAILED! => {"changed": false, "msg": "No package matching 'gnupg2' is available"}
```

There is a closed issue in the Ansible role for that: https://github.com/geerlingguy/ansible-role-docker/issues/263

## Checklist

## Related issues

- Relates https://github.com/geerlingguy/ansible-role-docker/issues/263

### DIFF
--- a/.ci/ansible/playbook.yml
+++ b/.ci/ansible/playbook.yml
@@ -30,7 +30,7 @@
   pre_tasks:
     - name: Update apt cache.
       apt: update_cache=yes cache_valid_time=600
-      when: ansible_pkg_mgr == 'apt' or ansible_distribution in ["Debian", "Ubuntu"]
+      when: ansible_pkg_mgr == 'apt'
   roles:
     - role: geerlingguy.docker
       docker_daemon_options:
@@ -103,7 +103,7 @@
   pre_tasks:
     - name: Update apt cache.
       apt: update_cache=yes cache_valid_time=600
-      when: ansible_pkg_mgr == 'apt' or ansible_distribution in ["Debian", "Ubuntu"]
+      when: ansible_pkg_mgr == 'apt'
   roles:
     - role: geerlingguy.docker
       docker_daemon_options:
--- a/.ci/ansible/playbook.yml
+++ b/.ci/ansible/playbook.yml
@@ -30,7 +30,7 @@
   pre_tasks:
     - name: Update apt cache.
       apt: update_cache=yes cache_valid_time=600
-      when: ansible_os_family == 'Debian'
+      when: ansible_pkg_mgr == 'apt' or ansible_distribution in ["Debian", "Ubuntu"]
   roles:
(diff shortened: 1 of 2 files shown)


.ci/ansible/playbook.yml
```diff
@@ -30,7 +30,7 @@
   pre_tasks:
     - name: Update apt cache.
       apt: update_cache=yes cache_valid_time=600
-      when: ansible_os_family == 'Debian'
+      when: ansible_pkg_mgr == 'apt' or ansible_distribution in ["Debian", "Ubuntu"]
```
v1v, line 33: IIUC, `apt` is the only package manager for `Debian` specific (including Ubuntu). So what's the reason for being so explicit?

mdelapenya: The error comes from https://beats-ci.elastic.co/blue/organizations/jenkins/e2e-tests%2Fe2e-testing-mbp/detail/PR-2103/3/pipeline, where a Debian machine (as declared in https://github.com/elastic/e2e-testing/blob/main/.ci/.e2e-tests.yaml#L4) hypothetically with Debian 10:
Screenshot 2022-02-08 at 10 28 14
 
does not find the gnugp package because the pre_task to update cache is skipped

Screenshot 2022-02-08 at 10 30 25

On the contrary, I've seen builds where the pre_task to update cache is in fact called:

Screenshot 2022-02-08 at 10 31 33
 
So my guess is the previous Ansible code is not evaluated

v1v: Gotcha, I understand the issue, but I don't understand the condition yet, maybe `when: ansible_pkg_mgr == 'apt'` should be sufficient?

mdelapenya: I'm pretty sure it is. I simply copied the very same when condition we have in https://github.com/elastic/e2e-testing/blob/main/.ci/ansible/tasks/install_deps.yml#L14

We could simplify all in this PR though

mdelapenya: Done in https://github.com/elastic/e2e-testing/pull/2107/commits/26037f0a593922f3d2f7defffbffa77fcb0ce052

v1v (approved):
:100 
Thanks Manu for helping me to understand this issue

PR #2203, opened 2022-03-03
Separate ES workload from Agent spec and make it req for the scenarios

ChrsMark:
## What does this PR do?

Separate ES workload from Agent spec and make it a requirement for the scenarios so as to ensure that deadlines because of download time are not hidden inside the scenario execution.

### DIFF
--- a/e2e/_suites/kubernetes-autodiscover/autodiscover_test.go
+++ b/e2e/_suites/kubernetes-autodiscover/autodiscover_test.go
@@ -116,8 +116,10 @@ func (m *podsManager) executeTemplateFor(podName string, writer io.Writer, optio
 }
 
 func (m *podsManager) configureDockerImage(podName string) error {
+	namespace := "beats"
+
 	if podName != "filebeat" && podName != "heartbeat" && podName != "metricbeat" && podName != "elastic-agent" && podName != "elasticsearch"{
-		log.Debugf("Not processing custom binaries for pod: %s. Only [filebeat, heartbeat, metricbeat, elastic-agent] will be processed", podName)
+		log.Debugf("Not processing custom binaries for pod: %s. Only [elasticsearch, filebeat, heartbeat, metricbeat, elastic-agent] will be processed", podName)
 		return nil
 	}
 
@@ -158,26 +160,16 @@ func (m *podsManager) configureDockerImage(podName string) error {
 			return err
 		}
 
-		if podName == "elastricsearch" {
-			// tag the image with the proper docker tag, including platform
-			err = deploy.TagImage(
-				"docker.elastic.co/elasticsearch/"+podName+":"+downloads.GetSnapshotVersion(common.BeatVersionBase),
-				"docker.elastic.co/observability-ci/"+podName+":"+beatVersion,
-			)
-			if err != nil {
-				return err
-			}
-		} else {
-			// tag the image with the proper docker tag, including platform
-			err = deploy.TagImage(
-				"docker.elastic.co/beats/"+podName+":"+downloads.GetSnapshotVersion(common.BeatVersionBase),
-				"docker.elastic.co/observability-ci/"+podName+":"+beatVersion,
-			)
-			if err != nil {
-				return err
-			}
+		if podName == "elasticsearch" {
+			namespace = "elasticsearch"
 		}
-
+		err = deploy.TagImage(
+			"docker.elastic.co/"+namespace+"/"+podName+":"+downloads.GetSnapshotVersion(common.BeatVersionBase),
+			"docker.elastic.co/observability-ci/"+podName+":"+beatVersion,
+		)
+		if err != nil {
+			return err
+		}
 		// load PR image into kind
 		err = cluster.LoadImage(m.ctx, "docker.elastic.co/observability-ci/"+podName+":"+beatVersion)
 		if err != nil {
--- a/e2e/_suites/kubernetes-autodiscover/autodiscover_test.go
+++ b/e2e/_suites/kubernetes-autodiscover/autodiscover_test.go
@@ -118,7 +118,7 @@ func (m *podsManager) executeTemplateFor(podName string, writer io.Writer, optio
 func (m *podsManager) configureDockerImage(podName string) error {
 	namespace := "beats"
 
-	if podName != "filebeat" && podName != "heartbeat" && podName != "metricbeat" && podName != "elastic-agent" && podName != "elasticsearch"{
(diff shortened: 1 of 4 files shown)


mdelapenya:
The errors are still there. If you think you need to wait before executing certain code, the tests allow you to wait for a certain condition, using a retry with backoff strategy, or more hard, using a Sleep.

ChrsMark:
@mdelapenya I didn't have the time to deal with it today but the thing is that running the suite locally it doesn't fail specially with the change applied on this PR. I believe that the issue in general is that we don't properly wait until ES is started properly. Since it seems that the init container does not really help here, do we have any specific alternative leveraging the framework in order to ensure that ES is up and running before starting the scenario? Sth like:

```
Scenario: Logs collection from running pod
  Given "elastic-agent" is running with "logs generic"
    And "elasticsearch" is running
   When "a pod" is deployed
   Then "elastic-agent" collects events with "kubernetes.pod.name:a-pod"
```

mdelapenya:
> do we have any specific alternative leveraging the framework in order to ensure that ES is up and running before starting the scenario? Sth like:
> 
> ```
> Scenario: Logs collection from running pod
>   Given "elastic-agent" is running with "logs generic"
>     And "elasticsearch" is running
>    When "a pod" is deployed
>    Then "elastic-agent" collects events with "kubernetes.pod.name:a-pod"
> ```

Let me think about it 🤔 I'll fetch this branch and try locally

mdelapenya:
I think you are right, we could decouple the elasticsearch k8s resources (configMap, deployment and service) and start them **before** the elastic agent is started.


I have two ideas:

1. rephrase your scenario like this (elasticsearch first):
```gherkin
Scenario: Logs collection from running pod
  Given "elasticsearch" is running
    And "elastic-agent" is running with "logs generic"
   When "a pod" is deployed
   Then "elastic-agent" collects events with "kubernetes.pod.name:a-pod"
```

2. set elasticsearch as a required resource, and install it in the `ctx.BeforeSuite` methods, right after initialising the kind cluster. I like this approach the most, as it reflects a real scenario where the elasticsearch instance does exist before any agent connected to it. OTOH, this corner case represents a good question for the agent team: what happens to the agent if the elasticsearch is not there on time?

ChrsMark:
Thanks for looking into this!

I think I had tried already the option 1 but with no success. I will try to give it another shot.
Regarding option 2: Will this mean that ES will be installed once in the startup of the suite? If this is true then the various scenarios will share the same ES and this might mean that the data will also be shared along the runs which is something that we don't want since we might find fields to be present from previous scenarios will the current scenario is actually failing. Is this concern valid?

mdelapenya:
We may need to rewrite the scenario and leverage the different life cycle hooks:
- move the elasticsearch installation to beforeSuite, so that it's started once
- make sure the elasticsearch indices of interest are clean in the afterScenario, so that we are not adding dirty data affecting other tests 
- make sure the data is namespaced. Is it possible to rename the indices where the data is stored so that they not collide in future, possible parallel tests?

ChrsMark:
After checking all options I will go with the basic one for now which seems simpler. Changing the suite implementation seems to me a bit overkill since we are gonna to re-use the same utils that we already have and can just work well with the current implementation of the scenario. I still believe that we can reach deadlines if ES image takes too long to get downloaded but not sure how we can handle it better.

e2e/_suites/kubernetes-autodiscover/autodiscover_test.go
```diff
@@ -158,13 +158,24 @@ func (m *podsManager) configureDockerImage(podName string) error {
...
 		}
 
-		// tag the image with the proper docker tag, including platform
-		err = deploy.TagImage(
-			"docker.elastic.co/beats/"+podName+":"+downloads.GetSnapshotVersion(common.BeatVersionBase),
-			"docker.elastic.co/observability-ci/"+podName+":"+beatVersion,
-		)
-		if err != nil {
-			return err
+		if podName == "elastricsearch" {
```
mdelapenya, line 161: typo:
```suggestion
		if podName == "elasticsearch" {
```

e2e/_suites/kubernetes-autodiscover/autodiscover_test.go
```diff
@@ -116,7 +116,7 @@ func (m *podsManager) executeTemplateFor(podName string, writer io.Writer, optio
 }
 
 func (m *podsManager) configureDockerImage(podName string) error {
-	if podName != "filebeat" && podName != "heartbeat" && podName != "metricbeat" && podName != "elastic-agent" {
+	if podName != "filebeat" && podName != "heartbeat" && podName != "metricbeat" && podName != "elastic-agent" && podName != "elasticsearch"{
 		log.Debugf("Not processing custom binaries for pod: %s. Only [filebeat, heartbeat, metricbeat, elastic-agent] will be processed", podName)
```
mdelapenya, line 120:
```suggestion
		log.Debugf("Not processing custom binaries for pod: %s. Only [elasticsearch, filebeat, heartbeat, metricbeat, elastic-agent] will be processed", podName)
```

e2e/_suites/kubernetes-autodiscover/autodiscover_test.go
mdelapenya: I'd define a `namespace := "beats"` variable, assigning it to `namespace = "elasticsearch"` if `podName == "elasticsearch"`. Then the below logic would not be duplicated, using the image's fully qualified name the namespace as part of the string.

e2e/_suites/kubernetes-autodiscover/features/elastic_agent_standalone.feature
```diff
@@ -4,30 +4,35 @@ Feature: elastic-agent standalone
   Use Kubernetes autodiscover features in elastic-agent standalone to collect logs and metrics
 
 Scenario: Logs collection from running pod
-  Given "elastic-agent" is running with "logs generic"
+  Given "elasticsearch" is running
+    And "elastic-agent" is running with "logs generic"
```
mdelapenya, line 8: We could have a single `Given` clause with:

```gherkin
Given "elastic-agent" is running with "logs generic"
```

and define a `Background` clause at the beginning of the feature file:

```gherkin
Background: Setting up elasticsearch
  Given "elasticsearch" is running
```
which will be run before every scenario. Therefore you can simplify each scenario removing the initial "Given elasticsearch" step

mdelapenya (approved):

mdelapenya:
@mergifyio

PR #2263, opened 2022-03-22
ci: publish cucumber reports as GitHub checks

v1v:
## What does this PR do?

Publish GitHub checks with the cucumber report

## Why is it important?

Developers wont' need to go anywhere else but look at the GitHub checks.

It uses https://github.com/elastic/apm-pipeline-library/blob/main/vars/githubCheck.groovy

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -603,7 +603,7 @@ def generateFunctionalTestStep(Map args = [:]){
           artifacts: "outputs/${testRunner.ip}/TEST-*${runId}*.xml, outputs/${testRunner.ip}/TEST-*${runId}*.json, ${cucumberFile}"
         // Create a GitHub check with the report
         if (fileExists(cucumberFile)) {
-          githubCheckNotify(name, 'neutral', cucumberFile)
+          githubCheckNotify(githubCheckName: "${name}-${runId}-cucumber", status: 'neutral', cucumberFile: cucumberFile)
         }
       }
     }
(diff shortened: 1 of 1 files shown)


.ci/Jenkinsfile
```diff
@@ -598,9 +598,38 @@ def generateFunctionalTestStep(Map args = [:]){
...
     }
   }
 }
+
+/**
+ Notify the GitHub check for the cucumber report
+ **/
+def githubCheckNotify(name, status, cucumberFile) {
+  def description = "${name} ${status.toLowerCase()}"
+  def context = name
```
mdelapenya, line 618:
```suggestion
  def checkName = name
```

.ci/Jenkinsfile
```diff
@@ -598,9 +598,38 @@ def generateFunctionalTestStep(Map args = [:]){
...
+  def url = "${env.RUN_DISPLAY_URL}"
+  def repo = env.REPO
+  def sha = env.GIT_BASE_COMMIT
+  if (params.GITHUB_CHECK_NAME?.trim() && params.GITHUB_CHECK_REPO?.trim() && params.GITHUB_CHECK_SHA1?.trim()) {
+    context = "${params.GITHUB_CHECK_NAME}-cucumber"
```
mdelapenya, line 623:
```suggestion
    checkName = "${params.GITHUB_CHECK_NAME}-cucumber"
```

.ci/Jenkinsfile
```diff
@@ -598,9 +598,38 @@ def generateFunctionalTestStep(Map args = [:]){
...
+    description = "${params.GITHUB_CHECK_NAME} ${status.toLowerCase()}"
+    sha = params.GITHUB_CHECK_SHA1
+    repo = params.GITHUB_CHECK_REPO
+  }
+  githubCheck(name: context,
```
mdelapenya, line 628:
```suggestion
  githubCheck(name: checkName,
```

mdelapenya (approved):
Great! Only a few comments about variable names. Not blockers at all

v1v:
I wonder if we need to ask for the cucumber plugin then this particular implementation will be much simpler?

mdelapenya:
It works! Although the html+css is not properly rendered, but it's good progress!

mdelapenya:
> I wonder if we need to ask for the cucumber plugin then this particular implementation will be much simpler?

Reading their docs, not sure if that would work: https://plugins.jenkins.io/cucumber-reports/#plugin-content-use

> You must use a Freestyle project type in jenkins.

v1v:
It supports pipelines too -> https://plugins.jenkins.io/cucumber-reports/#plugin-content-pipeline-usage

If that's the case and the URL predictable, we could create a GitHub status commit with a link to the Cucumber report

PR #2245, opened 2022-03-17
chore: bump Godog to v0.12.4

mdelapenya:
- chore: bump godog to v.0.12.4
- chore: update BeforeScenario/AfterScenario to Before/After
- chore: update BeforeStep/AfterStep to StepContext().Before/After
- chore: log before/after steps
- chore: consistent variable name
- chore: log step status after execution
- chore: support multiple formatters
- chore: store cucumber reports in Jenkins UI

## What does this PR do?
It bumps Godog from 0.11 to 0.12.4, which includes multi-formatter support.

For that, we have updated the Go dependencies, and had to update the usages of the lifecycle hooks defined by Godog: BeforeScenario/BeforeStep and AfterScenario/AfterStep are now deprecated.

Finally, we updated the scripts that generate the test report to include 3 formatters: pretty, cucumber and JUnit.

The cucumber format will be stored as a JSON file, downloaded by the Jenkins pipeline, and stored in Jenkins UI using the cucumber-reports plugin (See https://plugins.jenkins.io/cucumber-reports/)

## Why is it important?
Let's keep dependencies up-to-date, getting the new features it provides, in this case, the multi-formatter support, which will allow us to share nicer reports about the cucumber scenarios.

## Checklist

- [x] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have run the Unit tests (`make unit-test`), and they are passing locally
- [ ] I have run the End-2-End tests for the suite I'm working on, and they are passing locally
- [x] I have noticed new Go dependencies (run `make notice` in the proper directory)

## Related issues

- Closes #1488
- Relates #120

PR #2273, opened 2022-03-28
chore: include suite, platform and tags in test reports name

mdelapenya:
## What does this PR do?
It adds a prefix to the test report names so that it's easier to look them up in the Jenkins UI. In the past, the report name was created calling `uname` in the remote VM, now it uses the metadata coming from the platform descriptors

## Why is it important?
Consistency about file names, and adding discoverability when reading reports in the Jenkins UI

## Checklist

## Related issues

- Completes #2245

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -528,6 +528,7 @@ def generateFunctionalTestStep(Map args = [:]){
   def envContext = []
   envContext.add("PROVIDER=${provider}")
   envContext.add("SUITE=${suite}")
+  envContext.add("REPORT_PREFIX=${suite}_${platform}_${tags}")
   envContext.add("ELASTIC_APM_GLOBAL_LABELS=branch_name=${BRANCH_NAME},build_pr=${isPR()},build_id=${env.BUILD_ID},go_arch=${goArch},beat_version=${env.BEAT_VERSION},elastic_agent_version=${env.ELASTIC_AGENT_VERSION},stack_version=${env.STACK_VERSION}")
 
   def stackRunner = getNodeIp("${env.WORKSPACE}/${env.BASE_DIR}", 'stack')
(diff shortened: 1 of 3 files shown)


.ci/Jenkinsfile
```diff
@@ -528,6 +528,7 @@ def generateFunctionalTestStep(Map args = [:]){
   def envContext = []
   envContext.add("PROVIDER=${provider}")
   envContext.add("SUITE=${suite}")
+  envContext.add("REPORT_PREFIX=${suite}_${platform}_${tags}")
```
mdelapenya, line 531: Same as the parallel stage name

.ci/scripts/functional-test.sh
```diff
@@ -25,13 +25,14 @@ STACK_VERSION=${3:-"${BASE_VERSION}"}
 BEAT_VERSION=${4:-"${BASE_VERSION}"}
 ELASTIC_AGENT_VERSION=${5:-"${BASE_VERSION}"}
 GOARCH=${GOARCH:-"amd64"}
+REPORT_PREFIX="${SUITE}_${GOARCH}_${TAGS}"
```
v1v, line 28: What do you think to allow overriding the value only if REPORT_PREFIX is not set? IIUC, 

https://github.com/elastic/e2e-testing/pull/2273/files#diff-ef8b2f5e4d21d81c0eba4754dea3a3dfc2c67d26284f195165b650cf4aff7beaR564 uses `.ci/scripts/functional-test.sh` with only one parameter, so `REPORT_PREFIX="${SUITE}_${GOARCH}_${TAGS}"` = `REPORT_PREFIX=${tags}_ amd64_" but `REPORT_PREFIX` is already in place as it's created on the fly in the jenkinsfile

mdelapenya: That could be fine too. My only concern is that the script that is run on CI is generated on the fly in the Ansible code, and this is different to the committed script. I'd try to align them as much as possible

kuisathaverat (approved):

.ci/scripts/functional-test.sh
mdelapenya, line 28:
```suggestion
REPORT_PREFIX=${REPORT_PREFIX:-"${SUITE}_${GOARCH}_${TAGS}"}
```

mdelapenya: @v1v this is what you mean? I can enclose everything in quotes too, but I'm following same approach that in the above variables

PR #2293, opened 2022-03-30
chore: start sending OTEL traces for the e2e

mdelapenya:
## What does this PR do?
It uses the `junit2otel` apm-library step to collect OpenTelemetry traces and metrics for the project.

## Why is it important?
We'd like to start collecting traces, first to dogfood the test reports integration, second to collect useful information regarding SCM metrics from this project.

## Checklist

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -22,6 +22,7 @@ pipeline {
     AWS_PROVISIONER_SECRET = 'secret/observability-team/ci/elastic-observability-aws-account-auth'
     TEST_MATRIX_FILE = "${params.testMatrixFile}"
     RUN_ID = UUID.randomUUID().toString()
+    JUNIT_2_OTLP = "true"
   }
   options {
     timeout(time: 90, unit: 'MINUTES')
@@ -150,7 +151,11 @@ pipeline {
               }
               post {
                 always {
-                  junit(allowEmptyResults: true, keepLongStdio: true, testResults: "${BASE_DIR}/outputs/TEST-unit-*.xml")
+                  withEnv([
+                      "JUNIT_OTEL_TRACE_NAME=junit-unit-tests"
+                  ]){
+                    junit2otel(allowEmptyResults: true, keepLongStdio: true, testResults: "${BASE_DIR}/outputs/TEST-unit-*.xml")
+                  }
                   archiveArtifacts allowEmptyArchive: true, artifacts: "${BASE_DIR}/outputs/TEST-unit-*.xml"
                 }
               }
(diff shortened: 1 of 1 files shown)


v1v (approved):

.ci/Jenkinsfile
```diff
@@ -150,7 +151,11 @@ pipeline {
               }
               post {
                 always {
-                  junit(allowEmptyResults: true, keepLongStdio: true, testResults: "${BASE_DIR}/outputs/TEST-unit-*.xml")
+                  withEnv([
+                      "JUNIT_OTEL_TRACE_NAME=junit-unit-tests"
+                  ]){
+                    junit2otel(allowEmptyResults: true, keepLongStdio: true, testResults: "${BASE_DIR}/outputs/TEST-unit-*.xml")
```
v1v, line 157: I'm just wondering if the signature for this step could be changed to use arguments instead env variables? Probably we have discussed this in the past, sorry if so. No blocking but just an idea

mdelapenya: Yes we did :) https://github.com/elastic/apm-pipeline-library/pull/1514/commits. In particular,  https://github.com/elastic/apm-pipeline-library/pull/1514/commits/ed989e9ffd6a6508cf1e5d8ae12fc6fad93f9fbc we decided to go with env vars

mdelapenya: Screenshot 2022-03-30 at 11 08 58

v1v: Gotcha, so the signature is the main issue, I now remember what about using a map argument and then remove it when calling the junit (args)

```
call(def args) {
   otelArgs = args.otelArgs
   def service = otelArgs.service
   ...
   ...   

   return junit(args.delete(otelArgs))
```

Sorry for asking something not relevant to this particular PR, but somehow I'm thinking about the usability and whether it makes sense before rolling out to different places the new step

mdelapenya: indeed! for me the simplest usage would be:

```
junit2otel(traceName: "junit-unit-tests", allowEmptyResults: true, keepLongStdio: true, testResults: "${BASE_DIR}/outputs/TEST-unit-*.xml")
```

Which I think we can achieve it now, because we started overriding `junit` but we renamed the step to `junit2otel`. So I guess we can use params instead.

wdyt if we put this in the back burner and rewrite the junit2otel step to accept parameters?

v1v: > wdyt if we put this in the back burner and rewrite the junit2otel step to accept parameters?

Totally agree :)

mdelapenya: I submitted the changes in https://github.com/elastic/apm-pipeline-library/pull/1629

kuisathaverat (approved):

.ci/Jenkinsfile
mdelapenya: I changed the signature to use the step properly

Issue #1643, opened 2021-10-05
Scenario for the upgrade agent need to be fixed


IgorGuz2000:
https://github.com/elastic/e2e-testing/pull/1642
This scenario:

```gherkin
@upgrade-agent
@nightly
@skip:windows
Scenario Outline: Upgrading the installed <os> agent
  Given a "<os>" agent "stale" is deployed to Fleet with "tar" installer
    And agent is in version "stale"
    And certs are installed
    And the "elastic-agent" process is "restarted" on the host
  When agent is upgraded to version "latest"
  Then agent is in version "latest"
```
this need to be fixed because is it not working as designed first stem never installed stale version.

IgorGuz2000:
Use the branch that I created to run the scenario, and you will see that stale version never installed

mdelapenya:
Revisiting this issue.

After talking to @jlind23 we agreed in defining an upgrade matrix of **released** artifacts as stale agents, so that the code will install a released version of the elastic-agent, trying to upgrade it to the current branch's version.

From:
```gherkin
Scenario Outline: Upgrading the installed agent
  Given an agent "stale" is deployed to Fleet with "tar" installer
    And certs are installed
    And the "elastic-agent" process is "restarted" on the host
  When agent is upgraded to version "latest"
  Then agent is in version "latest"
```

We should move to (for 8.2 branch), in example:
```gherkin
Scenario Outline: Upgrading the installed agent
  Given an agent in "<stale-version>" is deployed to Fleet with "tar" installer
    And certs are installed
    And the "elastic-agent" process is "restarted" on the host
  When agent is upgraded to version "latest"
  Then agent is in version "latest"
Examples:
| stale-version  |
| 7.17.1         |
| 8.1.1          |
| 8.1.2          |
```

@jlind23 could you please fill in this upgrade matrix?

| branch | upgrade versions |
|--------|-------------------|
| main (8.3) | |
| 8.2 | |
| 8.1 | |
| 7.17 | |

I have another question: why we need to `restart` the agent process before the upgrade happens, to see the certificates? Is this something we can do before the installation so that we can skip the restart?

jlind23:
@mdelapenya The matrix should look like as follow
| branch | upgrade versions |
|--------|-------------------|
| main (8.3) | main (Self upgrade) |
| 8.2 | 8.2, main|
| 8.1 | 8.1, 8.2, main|
| 7.17 | 7.17, 8.1, 8.2, main|

[...]

PR #2344, opened 2022-04-07
fix: bump stale agent to 7.17.2

### LINKED ISSUE #2326: Flaky Test [Initializing / End-To-End Tests / fleet_centos8_arm64_fleet_mode_agent / Upgrading the installed agent – Fleet Mode Agent]
## Flaky Test

* **Test Name:** `Initializing / End-To-End Tests / fleet_centos8_arm64_fleet_mode_agent / Upgrading the installed agent – Fleet Mode Agent`
* **Artifact Link:** https://beats-ci.elastic.co/blue/organizations/jenkins/e2e-tests%2Fe2e-testing-mbp%2F7.17/detail/7.17/464/
* **PR:** None
* **Commit:** 6534fb8609444934bb9799c0dd4039e59ac1aa21

### Error details
```
Step certs are installed
```

mdelapenya:
## What does this PR do?
It bumps the stale agent version from 7.17.1 to 7.17.2

## Why is it important?
The 7.17.1 version disappeared from artifacts-api

I believe this upgrade agent scenario should be rewritten in a way that is not that sensible to the snapshot version, and instead uses a release version. For 7.17, it currently uses the branch's version minus 1 patch as a snapshot (see current changes), but this moving target is retired from the artifacts-api with certain frequency.

@v1v and I have mentioned a few times to detect the move and bump the stale agent version in an automated way, but I think it would be hiding the real problem of the sensitivity of this scenario.

@jlind23 @ph wdyt?

## Checklist

## Related issues

- Closes #2326 #2325 #2324 
- Relates https://github.com/elastic/e2e-testing/issues/1643

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -50,7 +50,7 @@ pipeline {
     string(name: 'ELASTIC_AGENT_DOWNLOAD_URL', defaultValue: '', description: 'If present, it will override the download URL for the Elastic agent artifact. (I.e. https://snapshots.elastic.co/7.12.0-069dfaa4/downloads/beats/elastic-agent/elastic-agent-7.12.0-SNAPSHOT-linux-x86_64.tar.gz')
     string(name: 'ELASTIC_AGENT_VERSION', defaultValue: '7.17.3-60b577ee-SNAPSHOT', description: 'SemVer version of the Elastic Agent to be used for the tests. You can use here the tag of your PR to test your changes')
     string(name: 'BEAT_VERSION', defaultValue: '7.17.3-60b577ee-SNAPSHOT', description: 'SemVer version of the Beat to be used for the tests. You can use here the tag of your PR to test your changes')
-    string(name: 'ELASTIC_AGENT_STALE_VERSION', defaultValue: '7.17.1-SNAPSHOT', description: 'SemVer version of the stale stand-alone elastic-agent to be used for Fleet upgrade tests.')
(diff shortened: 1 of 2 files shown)


jlind23:
@mdelapenya my question is, shouldn't we update those tests at the same time as the 7.17.2 is created? If we go too slow then artifacts will be deleted. Bumping the version in an automated way as we do for other repositories should fix this particular behaviour.

cachedout:
I don't understand why this is statically defined to begin with. Couldn't you just find the latest 7.x by querying [this API endpoint](https://artifacts-api.elastic.co/v1/versions/)?

mdelapenya:
> @mdelapenya I don't understand why this is statically defined to begin with. Couldn't you just find the latest 7.x by querying [this API endpoint](https://artifacts-api.elastic.co/v1/versions/)?

Me neither. This test was designed by @michalpristas before he left, and we do maintain. As explained in the issue:

>I believe this upgrade agent scenario should be rewritten in a way that is not that sensible to the snapshot version, and instead uses a release version. For 7.17, it currently uses the branch's version minus 1 patch as a snapshot (see current changes), but this moving target is retired from the artifacts-api with certain frequency.

A refactor of the scenario, having the team directions should be needed

michalpristas:
if i remember correctly there was something in framework which prevented me to go with dynamic versions.
if this is no longer the case refactor should take place rather than bumping up versions for sure

kuisathaverat (approved):

mdelapenya:
@cachedout as described in https://github.com/elastic/e2e-testing/issues/1643#issuecomment-1091549380, the idea is to refactor the steps according to that upgrade matrix using releases. Therefore, we would have to maintain the upgradeable versions for each branch.

PR #2468, opened 2022-05-03
feat: run tests on windows 2019

### LINKED ISSUE #2338: Windows Support
This is the main tracking issue for getting Windows fully supported in e2e-testing. The idea is that we will use the ogc provisioning tool to accomplish this. 

ref: https://github.com/elastic/e2e-testing/issues/1682

mdelapenya:
## What does this PR do?

In order to support Windows we have done the following:
- the Ansible code created an `.env` file to define environment variables that must have been sourced into the environment. Instead, we are using the convenient [`godotenv` library](https://github.com/joho/godotenv) that reads from a file and uses it to populate the current environment. So instead of sourcing within the Ansible scripts, we simply put the `.env` file in a well-known location (the `.op` directory under user's home) and let the library to automatically source it. A consequence of this is that the .env file does not need to be platform specific: no need of `export` or `set`. We simply need a file with `KEY=VALUE`.
- we are adjusting the usage of `become` in Ansible to represent the operations that need to be run as `root`. In particular, the installation of the Agent under Linux systems must be done using `root`, that's why we are keeping the invocation of the shell scripts/Make goals behind `sudo`. For the sake of simplicity, the Ansible scripts are copying the aforementioned `.env` file from default user in the machine (i.e. `admin` on Debian) to root, so that the `godotenv` library will read current user's home (root's) when loading the env file.
- we are adding the default user for each platform to the Docker group, which will allow users to SSH into the machines and check the Docker state.
- We are using Ansible vars and group_vars, to define variables that are specific to each supported platform (basically Linux Vs Windows). These variables files will hold references to the way Ansible communicates with the remote (`ansible_connection`), the shell to be used (`ansible_shell_type`: `sh` for Linux, `cmd` for Windows), and the Ansible Python interpreter (`ansible_python_interpreter`, `auto` for Linux and `C:\\Python310\\python.exe` for Windows). Finally, we define there paths that are platform-specific, such as the root directory of the project, and the current user's home dir.
- Separated concerns in different Ansible playbooks, so that it's simpler to identify what to update:
  - there is a `playbook.yml` for the main operation
  - there is a `run-tests.yml` for running the tests
  - there is a `fleet-server.yml` for recreating fleet-server
  - these playbooks will be run from the existing Make build, not the CI (there is a PR that will align what the CI does with the build. See #2490
- Added (basically duplicated) those Ansible tasks that modified files (`lineinfile`, `copy`) to be compliant with Windows. **I expected Ansible to handle this under the hood, but it seems that there is a `community.windows.win_lineinfile` and a `ansible.windows.win_copy` that need to be used** in a separate tasks YAML file. Therefore we are loading platform-specific tasks depending on what the Ansible facts retrieve, leveraging `ansible_distribution`, `ansible_system` and `ansible_os_family` to check for Windows (and any other Linux flavour). Another consequence of this is that using the `shell` module is not supported on Windows, needing `ansible.windows.win_shell` instead.
- Installed windows-specific tools with chocolatey: make. They could be installed with Packer (@adam-stokes?)
- Created Windows-specific Powershell scripts (*.ps1) to:
  - load the public SSH keys for the authorised users
  - run the tests
- Uses the AMI created by @adam-stokes that includes Windows and all the packages needed to run tests: ex, SSH.
- Replaced the powershell calls in Go to extract the ZIP file with the usage of the more convenient `archive/zip` Go package. Therefore, the zip file representing the agent is extracted in a native manner.
- Changed how the process state was obtained from the host: still using the Go library, but it does not work on Windows, so I simply check if the process is running instead of checking the process state.

## Why is it important?
We want to add back Windows support and start running tests on that platform, but to make it work we needed a lot of maintenance in the Ansible part.

## Checklist

## How to test this

0. create a VM for the stack
```shell 
make -C .ci create-stack
```

1. refresh fleet-server: for some reason, fleet-server dies because Kibana is not ready yet, so we need to recreate it
```shell
make -C .ci recreate-fleet-server
```

2. list the platforms to identify which node type to source into the build
```shell
make -C .ci list-platforms
```

3. create a windows node to run the tests. We are calling the `set-env-$PLATFORM` Make goal in order to generate the variables for that node type.
```shell
make -C .ci set-env-windows2019
source .ci/.node-windows2019-env
make -C .ci provision-node
```
It's very possibly that the SSH access to the Windows node takes too much (3-5 minutes?) Please wait for it before continuing to configure the node

4. configure the windows node
```shell
make -C .ci setup-node
```

5. run the tests
```shell
make -C .ci run-tests-win TAGS="fleet_mode && install"
```

## Related issues

- Closes #2338


.ci/.e2e-platforms.yaml
```diff
@@ -35,6 +35,11 @@ PLATFORMS:
     image: "ami-0f7cb53c916a75006"
     instance_type: "c5.4xlarge"
     username: ec2-user
+  windows2019:
+    description: "Windows 2019 x86_64 (HVM)"
+    image: "ami-0587bd602f1da2f1d"
+    instance_type: "c5.2xlarge"
+    username: ogc
```
mdelapenya, line 42: As documented in https://adam-stokes.github.io/ogc/user-guide/windows/

v1v (approved):

cachedout (approved):

kuisathaverat (approved):

.ci/ansible/tasks/setup_test_script.yml
```diff
@@ -115,46 +115,56 @@
...
 - name: Create Fleet test script file
-  become: no
   copy:
     owner: "{{ansible_user}}"
     mode: '0777'
-    dest: "/home/{{ansible_user}}/e2e-testing/.ci/scripts/functional-test.sh"
+    dest: "{{ e2e_base_dir }}.ci/scripts/functional-test.sh"
     content: |
        #!/usr/bin/env bash
        set -euxo pipefail
```
mdelapenya, line 154:
```suggestion

       BASE_DIR="/home/{{ansible_user}}/e2e-testing"
```

v1v (approved):
LGTM, with a minor comment

.ci/ansible/tasks/setup_test_script_windows.yml
```diff
@@ -124,6 +129,8 @@
 
        Add-Content C:\Users\ogc\.op\.env  "${REPORT}"
 
-       $env:TAGS="{{ lookup('env', 'TAGS') }}"; $env:FORMAT="pretty,cucumber:${REPORT}.json,junit:${REPORT}.xml"; make --no-print-directory -C "{{ e2e_base_dir }}e2e/_suites/${SUITE}" functional-test-win
+       $env:Path = [System.Environment]::GetEnvironmentVariable("Path","Machine")
+       cd "{{ e2e_base_dir }}\e2e\_suites\${SUITE}"
+       go test -timeout 60m -v --godog.format="pretty,cucumber:${REPORT}.json,junit:${REPORT}.xml" --godog.tags="{{ lookup('env', 'TAGS') }}"
```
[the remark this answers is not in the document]
mdelapenya: IIRC for some reason I was not able to propagate the environment so that running the Makefile caused the execution to fail because `go` was not found. With this, because the `env:Path` is set, it fairly works.

mdelapenya:
@jlind23 this PR is ready, so windows 2019 support is done.

We are running the fleet_mode scenarios (https://github.com/elastic/e2e-testing/blob/main/e2e/_suites/fleet/features/fleet_mode.feature). Are you interested in running any other ones?

Issue #2500, opened 2022-05-09
Do not use aliases when checking versions


mdelapenya:
The e2e started originally with an implementation based on version aliases, ex. 8.2-SNAPSHOT. But after moving to hashed snapshots (a long time ago) we did not update the checks for aliases.

It seems that, in the middle of a bump to a patch version (from 8.2.0-SNAPSHOT to 8.2.1-SNAPSHOT) we have detected an chicken-egg issue regarding the artifacts API is not finding the 8.2.1 artifacts, because Beats CI was not able to produce the artifacts because the e2e failed 😱 

@v1v is triggering a 8.2 build using the 8.2-SNAPSHOT alias for BEAT_VERSION (https://beats-ci.elastic.co/blue/organizations/jenkins/e2e-tests%2Fe2e-testing-mbp/detail/8.2/148/pipeline) which should pass as expected.

We are marking as top priority the elimination of any check for aliases in the e2e given we will always use hashed snapshots or released versions.

v1v:
PR:
- https://github.com/elastic/e2e-testing/pull/2501

PR #2501, opened 2022-05-09
avoid aliases if no major.minor format

v1v:
## What does this PR do?

Keep backward compatibility to query the artifacts-api if a version is an alias

## Why is it important?

Cornercase when bumping the patch version, since the artifacts-api is not yet available. So the alias detection should not be needed if no passing an alias.

## Checklist

Closes https://github.com/elastic/e2e-testing/issues/2500

### DIFF
--- a/internal/common/defaults.go
+++ b/internal/common/defaults.go
@@ -5,6 +5,8 @@
 package common
 
 import (
+	"regexp"
+
 	"github.com/elastic/e2e-testing/internal/shell"
 	"github.com/elastic/e2e-testing/pkg/downloads"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


mdelapenya:
Let's wait for the second build, using 8.2 parameters, so we can merge

mdelapenya:
Before merging, let me check the k8s-autodiscover scenarios: they are all failing

mdelapenya:
Build seems correct so far. Once finished, will merge and backport this one

Screenshot 2022-05-10 at 11 29 02

mdelapenya:
Test error is not related and marked as flaky

mdelapenya:
@mergify refresh

PR #2531, opened 2022-05-16
chore: use a nightly profile

mdelapenya:
- chore: copy default profile for nightly
- chore: remove upgrade tests from defaults
- chore: use nightly profile in the nightly build
- chore: simplify pipeline removing nightly code

## What does this PR do?
It creates a new profile YAML file that includes the upgrade scenarios. Instead of passing a boolean check to determine if the nightly scenarios should be included, we will simply pass the descriptor that defines waht scenarios to include:

- the new nightly descriptor is a copy of the current one, including the upgrade scenarios. It will be used from the nightly builds
- the default descriptor will be put in front of PRs. We are removing the upgrade scenarios from there.
- we are deprecating the `NIGHTLY_SCENARIOS` Jenkins parameter, in favour of the `testMatrixFile` descriptor (used by Beats and Elastic-Agent)
  - Agent: we should update https://github.com/elastic/elastic-agent/blob/752c699a10e9706d9489a490fd45afc571f8b840/.ci/Jenkinsfile#L256 to use the agent profile
  - Beats: https://github.com/elastic/beats/blob/main/.ci/packaging.groovy#L439
- we are removing targets in Makefile that are outdated, all of them using the nightly feature
- the pipeline has been simplified, as there are less logic to evaluate: it should simply use the profile without evaluation

>Note: we need to update the shared library here: https://github.com/elastic/apm-pipeline-library/blob/main/vars/runE2E.groovy#L65

## Why is it important?
The pipeline was creating a worker for the upgrade scenarios, which created an AWS instance to run no scenarios. With this code, we are simply not adding them to the pipeline, saving 3 Jenkins workers and 3 AWS instances.

## Checklist

## Author's Checklist

- [ ]


.ci/.e2e-tests-nightly.yaml
```diff
@@ -0,0 +1,61 @@
...
+        platforms: ["debian_arm64", "debian_amd64", "sles15"]
+      # - name: "Permissions Output change"
+      #   tags: "permission_change"
+      #   platforms: ["debian_arm64", "debian_amd64", "sles15"]
+      - name: "System Integration"
+        tags: "system_integration"
+        platforms: ["debian_arm64", "debian_amd64", "sles15"]
+      # - name: "Stand-alone"
+      #   tags: "stand_alone_mode"
+      #   platforms: ["fleet_elastic_pkg"]
```
mdelapenya, line 37: Because it's a new file, I think we can remove comments:
```suggestion
      - name: "System Integration"
        tags: "system_integration"
        platforms: ["debian_arm64", "debian_amd64", "sles15"]
```

pazone (approved):

kuisathaverat (approved):

.ci/e2eTestingFleetDaily.groovy
```diff
@@ -39,6 +39,7 @@ pipeline {
       steps {
         runE2E(jobName: "${env.JOB_BASE_NAME}",
                nightlyScenarios: true,
+               testMatrixFile: '.ci/.e2e-tests-nightly.yaml',
```
v1v, line 42: I'd say daily instead to match with the pipeline name?

```suggestion
               testMatrixFile: '.ci/.e2e-tests-daily.yaml',
```

mdelapenya: I like the names to be consistent, although people use to talk about nightly builds. But I'm OK with the change in any case

v1v (approved):

.ci/e2eTestingFleetDaily.groovy
v1v: I somehow missed to approve the PR, it was a minor comment, so no blocking this at all

mdelapenya: If you agree, we can rename the daily descriptor pipeline to nightly, so it's consistent everywhere

mdelapenya: After checking that any other "daily" file (JJB, groovy) is using the "daily" suffix, I think we should rename this PR's profile to "daily too. Will update it now

Thanks @v1v from bringing this concern!

v1v: Thanks for working on it 💯

PR #2563, opened 2022-05-26
chore: bump maintenance versions in mergify

mdelapenya:
8.3 is here, let's remove the least minor

## What does this PR do?
It adds 8.3 to mergify, removing 8.1 (prior lowest minor in the same major), and keeping 7.17 as the lowest minor

## Why is it important?
Keep the maintenance branches in sync with the upstream projects

## Checklist


.mergify.yml
```diff
@@ -18,41 +18,41 @@ pull_request_rules:
...
     actions:
       backport:
         assignees:
           - "{{ author }}"
         branches:
-          - "8.1"
+          - "8.2"
         title: "[{{ destination_branch }}](backport #{{ number }}) {{ title }}"
-  - name: backport patches to 7.17 branch
+  - name: backport patches to 8.1 branch
```
v1v, line 45: so `7.17` will not have any default backport, is that the expected behaviour?

mdelapenya: We always move the branch window one level. Although, thinking it twice, we may need to move the 8.minor pointer, not the 7 one. Therefore, we should keep:

7.17 > 8.2 > 8.3 > main

Is this correct?

v1v: I think so, `7.17` is the last `7` minor to be supported until `9` as per https://www.elastic.co/support/eol

mdelapenya: I kept 7.17, removing 8.1 instead. Thanks for the clarification!

kuisathaverat (approved):

PR #2637, opened 2022-06-13
chore: simplify extracting info from the supported platforms

mdelapenya:
- chore: support extracting env from the platforms file
- chore: support listing supported platforms

## What does this PR do?
This PR adds three build commands:

- list-platforms: uses YQ to extract the top-level keys in the YAML file representing the available platforms
- set-env-PLATFORM: uses YQ to generate a .env file for the platform, so that it's possible to source it and avoid passing it to each build command
- set-env-stack: same as the previous one, but for the stack

## Why is it important?
It will simplify how the build scripts interact with the node: instead of passing the env vars again and again, it will simply generate the env file once, and print a help message to source it.

Therefore, when running the tests against multiple OS targets (Centos 8, Debian 11, Ubuntu 22.04...), the end user will simply need to source the right env file just once.

Besides that, the list command will allow end-users to discover what platforms are supported.

## Checklist

List available platforms:
```shell
make -C .ci list-platforms
```

Set env for a specific platform:
```shell
make -C .ci set-env-centos8_amd64
make -C .ci set-env-centos8_arm64
make -C .ci set-env-debian_10_amd64
make -C .ci set-env-debian_10_arm64
make -C .ci set-env-debian_11_amd64
make -C .ci set-env-debian_11_arm64
make -C .ci set-env-oracle_linux8
make -C .ci set-env-sles15
make -C .ci set-env-ubuntu_22_04_amd64
```

## Related issues

-

### DIFF
--- a/.ci/scripts/yq.sh
+++ b/.ci/scripts/yq.sh
@@ -31,4 +31,7 @@ echo "export ${ENV_PREFIX}_LABEL=${PLATFORM}" >> "${FILE_ENV}"
 echo "export ${ENV_PREFIX}_SHELL_TYPE="${SHELL_TYPE} >> "${FILE_ENV}"
 echo "export ${ENV_PREFIX}_USER="$(docker run --rm -i  -w "/workdir" -v "${PWD}":/workdir ${YQ_IMAGE} ".${PLATFORM}.username" < yml.tmp) >> "${FILE_ENV}"
 
+echo ">>> Please source the generated file in order to use it while running the tests:"
+echo "source ${PWD}/${FILE_ENV}"
+
 rm -f ${PWD}/yml.tmp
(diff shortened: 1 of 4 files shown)


.ci/scripts/yq.sh
```diff
@@ -0,0 +1,37 @@
...
+
+if [ "${PLATFORM}" == "stack" ]; then
+  FILE_ENV=".stack-env"
+  ENV_PREFIX="STACK"
+fi
+
+docker run --rm  -w "/workdir" -v "${PWD}":/workdir ${YQ_IMAGE} ".PLATFORMS | with_entries( select( .key | (. == \"${PLATFORM}\") ) )" .e2e-platforms.yaml > ${PWD}/yml.tmp
+
+SHELL_TYPE=$(docker run --rm -i  -w "/workdir" -v "${PWD}":/workdir ${YQ_IMAGE} ".${PLATFORM}.shell_type" < yml.tmp)
+if [ "${SHELL_TYPE}" == "null" ]; then
```
mdelapenya, line 24: Condition needed by platforms not having shell_type (Windows, in #2468)

PR #2692, opened 2022-06-23
Add error checks to string typecast

fearful-symmetry:
## What does this PR do?

Hello from the elastic-agent team. I recently had an issue with `FetchElasticArtifact` where it tired to fetch an artifact that didn't exist, and rather than erroring out, it threw a really difficult to trace typecast panic. This appears to be an issue with how the API itself behaves when you ask for a non-existent package:
```
curl https://artifacts-api.elastic.co/v1/search/8.4.0-SNAPSHOT/elastic-agent-shipper\?x-elastic-no-kpi\=true
{"packages":[],"manifests":{"last-update-time":"Thu, 23 Jun 2022 17:38:20 UTC","seconds-since-last-update":134}}
```
We don't get an error, just an empty list. When we try to cast to string, we get an error. Not sure if there's a better way to do this, but I figured it would be at least nice to have.

## Why is it important?

This produces a very difficult to debug error. particularly when `e2e-testing` is imported from elsewhere.

## Checklist

### DIFF
--- a/pkg/downloads/releases.go
+++ b/pkg/downloads/releases.go
@@ -141,8 +141,14 @@ func (r *ArtifactURLResolver) Resolve() (string, string, error) {
 		return "", "", fmt.Errorf("object not found in Artifact API")
 	}
 
-	downloadURL := downloadObject.Path("url").Data().(string)
-	downloadshaURL := downloadObject.Path("sha_url").Data().(string)
+	downloadURL, ok := downloadObject.Path("url").Data().(string)
+	if !ok {
+		return "", "", fmt.Errorf("key 'url' does not exist for artifact %s", artifact)
+	}
+	downloadshaURL, ok := downloadObject.Path("sha_url").Data().(string)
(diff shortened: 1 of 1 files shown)


fearful-symmetry:
Not sure if a backport is needed here, since this is a small fix?

mdelapenya:
Thanks for your PR! We never faced the issue because the artifacts "are always there" but I see a lot of value in your contribution.

I'm retriggering the build here https://beats-ci.elastic.co/blue/organizations/jenkins/e2e-tests%2Fe2e-testing-mbp/detail/PR-2692/2/pipeline, as I don't know why it did not trigger all the tests to check if the downloaded artifacts are affected. Will mark this as ready to merge once the CI finishes

Again, thanks for your time in fixing it!

mdelapenya (approved):
I've reviewed the test errors and they seems not related: the artifacts are properly downloaded, so this is a LGTM. Thanks!

👍

PR #2706, opened 2022-06-27
ci: notify test failures for MacOS nightly builds

v1v:
## What does this PR do?

Notify if test failures to the `elastic-agent` team so they can track any MacOS test failures easily.

## Why is it important?

As requested by the team


.ci/e2eTestingMacosDaily.groovy
```diff
@@ -86,11 +92,22 @@ pipeline {
       destroyCluster()
     }
     cleanup {
-      notifyBuildResult()
+      notifyBuildResult(prComment: true,
+                        slackHeader: "*Test Suite*: ${env.TAGS}",
```
mdelapenya, line 96: Should we add that the build is related to MacOS here?
```suggestion
                        slackHeader: "*Test Suite (MacOS)*: ${env.TAGS}",
```

PR #2707, opened 2022-06-27
refactor: pipeline for MacOS to use a Makefile instead

v1v:
## What does this PR do?

Simplify the CI pipeline by using a Make goal to run the tests with the different environment variables

## Why is it important?

Help with the configuration, so anyone can add changes in the Makefile without touching the CI Pipeline and test them locally easily.

### DIFF
--- a/.ci/Makefile
+++ b/.ci/Makefile
@@ -225,5 +225,6 @@ run-tests-win: validate-aws
 .PHONY: run-fleet-tests-macos
 run-fleet-tests-macos:
 	FORMAT="pretty,cucumber:fleet_mode.json,junit:fleet_mode.xml" \
+	PROVIDER='remote' \
 	TAGS="fleet_mode && install" \
 	$(MAKE) -C "../e2e/_suites/fleet" functional-test
--- a/.ci/Makefile
+++ b/.ci/Makefile
@@ -42,8 +42,15 @@ __check_defined = \
     $(if $(value $1),, \
       $(error Undefined $1$(if $2, ($2))))
 
-@:$(call check_defined, AWS_ACCESS_KEY_ID, You need to export AWS_ACCESS_KEY_ID to create AWS resources under that account)
-@:$(call check_defined, AWS_SECRET_ACCESS_KEY, You need to export AWS_SECRET_ACCESS_KEY to create AWS resources under that account)
+validate-aws:
+ifndef AWS_ACCESS_KEY_ID
+	@echo "You need to export AWS_ACCESS_KEY_ID to create AWS resources under that account"
+	@exit 1
+endif
+ifndef AWS_SECRET_ACCESS_KEY
+	@echo "You need to export AWS_SECRET_ACCESS_KEY to create AWS resources under that account"
+	@exit 1
+endif
 
 # Create the run id if it does not exist
 .runID:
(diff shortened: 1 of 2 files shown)


.ci/Makefile
```diff
@@ -213,3 +220,10 @@ run-tests-win:
 		-t run-tests \
 		-i $(NODE_IP_ADDRESS),
 	ssh -i $(SSH_KEY) $(NODE_USER)@$(NODE_IP_ADDRESS) 'powershell "C:\Users\$(NODE_USER)\e2e-testing\.ci\scripts\functional-test.ps1"'
+
+# Make goal to run the tests for the MacOS platform. Used by the CI.
+.PHONY: run-fleet-tests-macos
+run-fleet-tests-macos:
```
mdelapenya, line 226: I wonder if this goal will be misunderstood by the users, as it's CI-only, what do you think of prepending the `ci-` prefix to it? `ci-run-fleet-tests-macos`

.ci/Makefile
```diff
@@ -146,10 +153,10 @@ setup-node: .runID
 		-i $(NODE_IP_ADDRESS),
 
 .PHONY: create-node
-create-node: provision-node setup-node
+create-node: provision-node setup-node validate-aws
```
mdelapenya, line 156: Is this needed? The two downstream goals already check for the validation

v1v: let me remove it

.ci/Makefile
v1v: it's not only `ci-` specific, but wanted to clarify that it's consumed by the CI, ideally anyone could run the same goal locally to reproduce the tests

mdelapenya: if it's possible to locally invoke the ORKA provider, then it's OK to me, although not sure if that's currently possible, only done at the CI

kuisathaverat (approved):

.ci/Makefile
v1v:
```suggestion
create-node: provision-node setup-node
```

.ci/Makefile
```diff
@@ -213,3 +220,10 @@ run-tests-win:
 		-t run-tests \
 		-i $(NODE_IP_ADDRESS),
 	ssh -i $(SSH_KEY) $(NODE_USER)@$(NODE_IP_ADDRESS) 'powershell "C:\Users\$(NODE_USER)\e2e-testing\.ci\scripts\functional-test.ps1"'
+
+# Make goal to run the tests for the MacOS platform. Used by the CI.
+.PHONY: run-fleet-tests-macos
+run-fleet-tests-macos:
+	FORMAT="pretty,cucumber:fleet_mode.json,junit:fleet_mode.xml" \
```
v1v, line 227:
```suggestion
	PROVIDER='remote' \
	FORMAT="pretty,cucumber:fleet_mode.json,junit:fleet_mode.xml" \
```

.ci/Makefile
v1v: it depends on how to setup the environment, if ESS then:

```bash
### spin up cluster
export CLUSTER_NAME=my-test-cluster
.ci/scripts/deployment.sh 'create'
# set env variables stored in Vault
export ELASTICSEARCH_URL=$(vault read ..../${CLUSTER_NAME})
export KIBANA_URL=$(vault read ....)
...
# run the tests for MacOS
PROVIDER='remote' make -C .ci run-fleet-tests-macos

# destroy cluster
.ci/scripts/deployment.sh 'destroy'
```

mdelapenya: Ok, now I think I get it: this goal is to be used when running the tests on Mac, no matter if it's an Orka machine (CI time) or a local Mac, right?

mdelapenya: I'd like to have a make goal that retrieves the variables in the same manner the set-env-* does.

Apart from that, in a follow-up, we can entirely replace the contents of the `create-stack` goal with the creation of the cluster, and this would represent a great step forward to test in Cloud!!

PR #2811, opened 2022-07-19
feat(ci): support passing a commit SHA to run the MacOS tests

mdelapenya:
## What does this PR do?
It adds two input parameters to the Mac pipeline:

- GITHUB_CHECK_REPO: name of the repository where to look up the SHA. Default is the elastic-agent, and this should not be changed unless used by an upstream caller like Beats
- GITHUB_CHECK_SHA1: SHA of the commit in the above repo. The artifacts for the SHA must have been produced by the packaging job, therefore they live in a GCP bucket

If present, the e2e code will use them to download the artifacts for that commit, using them in the execution

## Why is it important?
To allow debug a Mac build

## Checklist

### DIFF
--- a/.ci/e2eTestingMacosDaily.groovy
+++ b/.ci/e2eTestingMacosDaily.groovy
@@ -37,7 +37,7 @@ pipeline {
     string(name: 'ELASTIC_AGENT_VERSION', defaultValue: '8.4.0-SNAPSHOT', description: 'SemVer version of the Elastic Agent to be used for the tests. You can use here the tag of your PR to test your changes')
     string(name: 'ELASTIC_STACK_VERSION', defaultValue: '8.4.0-SNAPSHOT', description: 'SemVer version of the stack to be used for the tests.')
     string(name: 'BEAT_VERSION', defaultValue: '8.4.0-SNAPSHOT', description: 'SemVer version of the Beat to be used for the tests. You can use here the tag of your PR to test your changes')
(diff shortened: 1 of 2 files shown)


.ci/e2eTestingMacosDaily.groovy
```diff
@@ -35,6 +37,8 @@ pipeline {
     string(name: 'ELASTIC_AGENT_VERSION', defaultValue: '8.4.0-SNAPSHOT', description: 'SemVer version of the Elastic Agent to be used for the tests. You can use here the tag of your PR to test your changes')
     string(name: 'ELASTIC_STACK_VERSION', defaultValue: '8.4.0-SNAPSHOT', description: 'SemVer version of the stack to be used for the tests.')
     string(name: 'BEAT_VERSION', defaultValue: '8.4.0-SNAPSHOT', description: 'SemVer version of the Beat to be used for the tests. You can use here the tag of your PR to test your changes')
+    string(name: 'GITHUB_CHECK_REPO', defaultValue: 'elastic-agent', description: 'Name of the GitHub repo to be updated. Only modified if this build is triggered from another parent stream (i.e. Beats).')
```
v1v, line 40: What's the reason for using a default value?

```suggestion
    string(name: 'GITHUB_CHECK_REPO', defaultValue: '', description: 'Name of the GitHub repo to be updated. Only modified if this build is triggered from another parent stream (i.e. Beats).')
```

mdelapenya: It could be empty too, to be honest. We use the agent repo by default to be meaningful to the agent repo, but it would be totally OK as an empty string

v1v: It might mislead, so I'd say to keep it empty

kuisathaverat (approved):

PR #2842, opened 2022-07-25
chore: deprecate the usage of BEATS_LOCAL_PATH

mdelapenya:
## What does this PR do?
This PR deprecates the usage of `BEATS_LOCAL_PATH`, removing it from the docs, removing the code branches that evaluated it, updating the unit tests that consumed it, and raising an error if used in the tests.

## Why is it important?
When we added support for running the tests on the AWS instances, we unconsciously dropped this support: we are not uploading the local binaries from that local path to the AWS instance in charge of running the tests.

## Checklist

## Author's Checklist

- [ ] we could keep this feature simply adding a `build/distributions` dir in the project that would upload the local binaries in the case they exist. But this needs a manual intervention from the user to copy the artifacts to that location first, in a well-known structure. This way we could keep the behaviour. But to be honest, I'm pretty sure nobody is using it :shrugh:


kuisathaverat (approved):

pkg/downloads/versions_test.go
```diff
@@ -396,125 +394,114 @@ func TestFetchBeatsBinaryFromLocalPath(t *testing.T) {
 		BeatsLocalPath = beatsDir
 
 		artifactName := versionPrefix + "-x86_64.rpm"
-		expectedFilePath := path.Join(distributionsDir, artifactName)
 
 		downloadedFilePath, err := FetchBeatsBinary(ctx, artifactName, artifact, version, utils.TimeoutFactor, true, "", false)
-		assert.Nil(t, err)
-		assert.Equal(t, downloadedFilePath, expectedFilePath)
+		assert.NotNil(t, err)
```
mdelapenya, line 399: Let's remove the tests once the variable is entirely removed

PR #2850, opened 2022-07-26
chore: decouple responsibilities for fleet test suite code

mdelapenya:
## What does this PR do?
This PR tries to decouple the logic that existed under the fleet.go file into smaller files, each one with its own resposibility.

For that, we have created Go files related to each feature file, so the methods that existed in the old and huge fleet.go file, are now spread on different, single-responsibilty Go files, specialized by the feature file that uses it.

We are following these rules:
- test files (i.e. fleet_test.go) are those that actually execute the tests, therefore any method related to the life cycle of the test execution should live there (i.e. befores/afters), including the mandatory TestMain method, the entrypoint of the test execution.
- non test files (i.e. fleet.go, elastic_agent_checks.go, upgrades.go, etc) contains logic for its specialized tasks. They still can be consumed by other files (they live in the same package)

During this big refactor, we are:

- removing the old IngestManagerTestSuite struct, which was an useless indirection layer to access the FleetTestSuite struct. All occurrences and methods have been updated to use the Fleet struct
- simplifying a few methods in the fleet struct, as they simply called another method. We have joined them in those cases.
- fixing a cosmetical issue with the name of the test suite at Godog's APIs: it was hardcoded to `godogs` and now they match with the real test suite

## Why is it important?
Simplicity, readability, maintainability... all the `ilities` that you can imagine when talking about maintaining a software project. Small files means small processing code chunks, faster search of specific functionality and less code to remember when working on the project.

Hopefully this PR helps in that direction

## Checklist

## Author's Checklist

- [ ]


mdelapenya:
cc/ @dikshachauhan-qasource

PR #2858, opened 2022-07-27
chore: simplify process action package

mdelapenya:
- chore: move action package into the process package
- chore: do not expose action struct and methods

## What does this PR do?
It moves everything into the `internal/action` package to the the `internal/process`, converting all its exposed methods, structs and contants into private.

## Why is it important?
Simplify the code, and hide the implementation details of checking a process state, exposing only one public method for that.

## Checklist


cachedout:
@pazone Perhaps it would be good to review this in a screen sharing session with Manu as a part of helping with the transition?

internal/process/processes.go
```diff
@@ -120,7 +120,7 @@ func CheckState(ctx context.Context, deployer deploy.Deployment, service deploy.
 }
 
 // Run executes the command
-func (a *actionWaitProcess) Run(ctx context.Context) (string, error) {
+func (a *actionWait) Run(ctx context.Context) (string, error) {
```
pazone, line 123: I'm not an expert at go but we could also make `run` function private :)

internal/process/processes.go
```diff
@@ -59,3 +118,288 @@ func CheckState(ctx context.Context, deployer deploy.Deployment, service deploy.
 
 	return nil
 }
+
+// Run executes the command
+func (a *actionWait) Run(ctx context.Context) (string, error) {
```
pazone, line 123:
```suggestion
func (a *actionWait) run(ctx context.Context) (string, error) {
```

pazone (approved):

internal/process/processes.go
mdelapenya: I'm checking the code, and there is an interface in the deploy package, `ServiceOperation`, that exposes a `Run` method. Therefore, this method should be kept public to match the interface.

Issue #541, opened 2020-12-15
[Fleet] Add a test for passing in api token to Docker container


EricDavisX:
this is apparently not covered in what we have, as it has been broken in 7.10 line for some time.  

More details here:
https://github.com/elastic/beats/pull/23155

I don't fully understand the use case, but we can fill out the ticket here and then implement the test for it.

ruflin:
My use case here is that I want to run the Elastic Agent Docker container but Fleet and everything is already setup. To make sure the Elastic Agent enrolls into Fleet on startup, environment variables need to be passed in.

mdelapenya:
> My use case here is that I want to run the Elastic Agent Docker container but Fleet and everything is already setup. To make sure the Elastic Agent enrolls into Fleet on startup, environment variables need to be passed in.

If I understand correctly this use case, we are starting ES, Kibana and Package registry before the agent container is even created. Then we initialise Fleet using an API call, and then we create the container. It always worked this way. Is that what you expect @ruflin?

ruflin:
@mdelapenya Almost. There is a small difference here. My guess is that the Fleet API so far was called twice, once through the separate API call and once when the Elastic Agent was started by the Container. This is the reason it still worked. But if you pass in only the enrollment key to the container, things broke (should be fixed now). The reason I assume CI does not test it that way exactly is because it seems it didn't break CI ;-)

mdelapenya:
When we deploy an agent to fleet, we:
1. Create a Fleet token for the agent, using an API call
1. PreInstall: for centos and debian installers, we do not execute code. For Tar installer, we extract the tar file
1. Install: for centos and debian, we use the package manager (yum and apt) to local-install the packages. For Tar installer, we execute the enroll command for the agent using its enrollment token
1. PostInstall: for centos and debian installers, we call systemd to enable and start the linux service, and then execute the `enroll` command for the agent using its enrollment token. For Tar installer, we do not execute any code else.

ruflin:
@mdelapenya Not sure we talk about the same thing. This issue is only related to Docker but your list above seems to be for the parts outside docker?

mdelapenya:
@ruflin @EricDavisX  sorry for the late response. We triaged this issue today, and we discovered we would need more information to make progress on it. IIUC, what we want here is:

- enroll a stand-alone agent (using official Docker image for the agent)
- run "A_COMMAND_WITH_FLAGS" in the agent
- check the agent is still enrolled in Fleet.

[...]

PR #2879, opened 2022-07-31
chore: bring back stand-alone agent

mdelapenya:
- chore: apply functional opts to agent deployment
- chore: rename deployment method
- chore: simplify stand-alone deployment removing flavour
- fix: revoke token scenario was outdated
- chore: simplify deploy stand-alone agent method
- chore: skip scenario for bootstrap fleet server agent in stand-alone
- chore: include stand-alone agent in CI

## What does this PR do?
It adds the stand-alone agent scenarios to the CI including a few changes:

- the bootstrap fleet-server scenario is skipped, because it uses docker-compose to start the fleet-server in local, and the agent is using the already deployed stack
- uses a new functional approach (more Go idiomatic) when deploying an agent to Fleet. It uses optional functions instead of input parameters to configure the deployment (installer type, extra flags, etc)
- updates the revoke-token scenario, as it was based on the old docker-compose style. Now it installs the agent, revokes the token, uninstalls it, and tries to install an agent with the same token again
- detected that Ansible Docker ARM role was not applied to the ARM instances, causing a failure for having Docker not installed on the instance. Docker is needed because the stand-alone mode uses the Docker image of the elastic-agent.

## Why is it important?
Add scenarios that were already working in the past, also improving the readability of the deploment of the agent code

## Checklist

## Related issues

- Relates #541

### DIFF
--- a/e2e/_suites/fleet/elastic_agent_deploy.go
+++ b/e2e/_suites/fleet/elastic_agent_deploy.go
@@ -24,24 +24,24 @@ func (fts *FleetTestSuite) anAgentIsDeployedToFleet(image string) error {
 		installerType = "deb"
 	}
 
-	return fts.anAgentIsDeployedToFleetWithInstallerAndFleetServer(InstallerType(installerType))
+	return fts.deployAgentToFleet(InstallerType(installerType))
 }
 
 func (fts *FleetTestSuite) anAgentIsDeployedToFleetOnTopOfBeat(beatsProcess string) error {
-	return fts.anAgentIsDeployedToFleetWithInstallerAndFleetServer(InstallerType("tar"), BeatsProcess(beatsProcess))
+	return fts.deployAgentToFleet(InstallerType("tar"), BeatsProcess(beatsProcess))
 }
 
 // supported installers: tar, rpm, deb
 func (fts *FleetTestSuite) anAgentIsDeployedToFleetWithInstaller(installerType string) error {
-	return fts.anAgentIsDeployedToFleetWithInstallerAndFleetServer(InstallerType(installerType))
+	return fts.deployAgentToFleet(InstallerType(installerType))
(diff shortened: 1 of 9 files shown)


cachedout (approved):

mdelapenya:
Mmm the Ansible role to install Docker does not install Docker Compose on ARM. Will take a look, as it's an external dependency

e2e/_suites/fleet/elastic_agent_deploy.go
```diff
@@ -23,64 +23,47 @@ func (fts *FleetTestSuite) anAgentIsDeployedToFleet(image string) error {
...
 func (fts *FleetTestSuite) anAgentIsDeployedToFleetWithInstallerAndTags(installerType string, flags string) error {
-	fts.BeatsProcess = ""
+	return fts.deployAgentToFleet(InstallerType(installerType), Flags(flags))
+}
 
-	// FIXME: We need to cleanup the steps to support different operating systems
-	// for now we will force the zip installer type when the agent is running on windows
-	if runtime.GOOS == "windows" && common.Provider == "remote" {
-		installerType = "zip"
+func (fts *FleetTestSuite) deployAgentToFleet(opts ...DeploymentOpt) error {
```
pazone, line 44: It seems like `opts` are not used later in the function. Only logged.

mdelapenya: It's used here: https://github.com/elastic/e2e-testing/pull/2879/files#diff-826ef9ac59e049ffe371ffe0262603320d60f544c3263af228aaded6598b4bfdR53

it loops through the opts using range, and for each passed opt, it applies it. You can find an example for functional opts here: https://github.com/tmrts/go-patterns/blob/master/idiom/functional-options.md

PR #2888, opened 2022-08-01
docs: add Python version

### LINKED ISSUE #2883: INclude Python +3.9 as a dependency
Include that the venv must match the python version to avoid Ansible errors from python deps

mdelapenya:
## What does this PR do?
It documents the Python version needed to interact with the build system.

## Why is it important?
The QA Source folks have faces errors in the build even when there have not been changes in the build system, which at first sight seemd weird. After debugging the issue, we identified the controller machines (the ones running Ansible) where using an outdated version og Python (3.7.x), which led to errors interacting with boto_core and the ec2 module.

With these changes we expect it's more clear what versions to use.

## Checklist

## Related issues

- Closes #2883 

## Follow-ups
Dockerising the build and running all the Make commands in an ephemeral docker run command using Python 3.9 will be super. We should create a build-tools machine for that

### DIFF
--- a/.ci/README.md
+++ b/.ci/README.md
@@ -46,6 +46,10 @@ You could be insterested in adding a new operative system in a specific architec
 
 ### Prereqs
 
+In order to interact with the build system, i.e. running Ansible commands to provision the AWS instances, you need to install Python 3.9 or above (See https://www.python.org/downloads/).
+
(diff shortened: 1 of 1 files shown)


mdelapenya:
cc/ @dikshachauhan-qasource

cachedout (approved):

PR #2906, opened 2022-08-03
fix: retire 8.2-SNAPSHOT from upgrade tests

mdelapenya:
## What does this PR do?
It removes the 8.2-SNAPSHOT example from the Scenario outline representing the upgrade agent tests.

## Why is it important?
In the light of the appearance of 8.5-SNAPSHOT, 8.2-SNAPSHOT dissapeared from the artifacts API. Therefore any usage of that version needs to be removed from any scenario.

And because the upgrade scenarios for each maintenance branch uses the previous version to perform the upgrades, we removed the version here in the 8.3 maintenance branch. 

## Checklist


ph:
thanks @mdelapenya

PR #2976, opened 2022-08-30
jjbb: support for feature flags

v1v:
## What does this PR do?

Enable feature branches with the naming `feature-.*`, such as `feature-arch-v2`

## Why is it important?

Consumers are using this approach:

> We need the Beats [feature-arch-v2](https://github.com/elastic/beats/tree/feature-arch-v2) branch paired with the Elastic agent [feature-arch-v2](https://github.com/elastic/elastic-agent/tree/feature-arch-v2) branch for E2E test runs. This should be the same situation as the release branches in both repositories, just with a feature branch instead.

## Issues

Supersedes https://github.com/elastic/apm-pipeline-library/pull/1850

### DIFF
--- a/.ci/jobs/e2e-testing-fleet-daily-mbp.yml
+++ b/.ci/jobs/e2e-testing-fleet-daily-mbp.yml
@@ -9,7 +9,7 @@
     scm:
       - github:
           branch-discovery: no-pr
-          head-filter-regex: '(main|8\.\d|7\.17)'
+          head-filter-regex: '(main|8\.\d|7\.17|feature-.*)'
           discover-pr-forks-strategy: merge-current
           discover-pr-forks-trust: permission
           discover-pr-origin: merge-current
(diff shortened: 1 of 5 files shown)


pazone (approved):

.ci/jobs/e2e-testing-fleet-daily-mbp.yml
```diff
@@ -9,7 +9,7 @@
     scm:
       - github:
           branch-discovery: no-pr
-          head-filter-regex: '(main|8\.\d|7\.17)'
+          head-filter-regex: '(main|8\.\d|7\.17|feature-.*)'
```
cmacknz, line 12: What effect is this having? It seems like it is just including feature branches in a search?

v1v: Will enable those branches to be available in the CI pipelines. https://beats-ci.elastic.co/job/e2e-tests/job/e2e-testing-mbp/ and similar multibranch pipelines

In other words, those regex will be the ones scanned and shown in Jenkins

PR #3137, opened 2022-10-19
ci: reprovision worker if it's flaky

v1v:
## What does this PR do?

Re-provision if the worker is flaky and retry again in a new CI worker.

## Why is it important?

If the worker is unstable, then the stage will fail and hence the build. 

It assumes the worker is unstable if it cannot reach the dir in the `finally` section. 

## Issues

A similar approach was planed in https://github.com/elastic/beats/pull/22662 but it never got merged.


pazone (approved):

.ci/Jenkinsfile
```diff
@@ -604,3 +622,50 @@ def generateFunctionalTestStep(Map args = [:]){
...
+  workerStatus[stageName].retries = current + 1
+}
+
+def setFlakyWorker(stageName) {
+  // set the status to be failed for now
+  workersStatus[stageName].status = false
+}
+
+def unsetFlakyWorker(stageName) {
+  // set the status to be failed for now
```
v1v, line 669:
```suggestion
```

.ci/Jenkinsfile
```diff
@@ -604,3 +622,50 @@ def generateFunctionalTestStep(Map args = [:]){
     }
   }
 }
+
+def retryWithNode(Map args = [:], Closure body) {
+  try {
+    // Increment the retries
```
v1v, line 628:
```suggestion
```

kuisathaverat (approved):

PR #3167, opened 2022-10-26
Update references of epr-staging to production

mrodm:
## What does this PR do?

This PR updates the references to epr-staging.elastic.co and its docker image, to use the production ones: epr.elastic.co and its docker image docker.elastic.co/package-registry/distribution:production.

## Why is it important?

After the switch to package storage V2, staging endpoint will be pointing to production too. 
And due to the same reason, the docker image of the staging distribution (docker.elastic.co/package-registry/distribution:staging) will not get more updates. The reference to use after the switch to V2 should be production URL and docker image.

## Checklist

- [ ] ~~My code follows the style guidelines of this project~~
- [ ] ~~I have commented my code, particularly in hard-to-understand areas~~
- [ ] ~~I have made corresponding changes to the documentation~~
- [ ] ~~I have made corresponding change to the default configuration files~~
- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~
- [ ] ~~I have run the Unit tests (`make unit-test`), and they are passing locally~~
- [ ] ~~I have run the End-2-End tests for the suite I'm working on, and they are passing locally~~
- [ ] ~~I have noticed new Go dependencies (run `make notice` in the proper directory)~~

## Author's Checklist

- [ ] To be merged after the switch to storage V2.


internal/config/kubernetes/base/package-registry/deployment.yaml
```diff
@@ -16,7 +16,7 @@ spec:
     spec:
       containers:
       - name: package-registry
-        image: docker.elastic.co/package-registry/distribution:staging
+        image: docker.elastic.co/package-registry/distribution:production
```
jsoriano, line 19: Maybe we should use the lite image here, or a vanilla one with a proxy to production. But this can be left for another PR.

jsoriano (approved):

PR #3375, opened 2023-01-16
fix: typo in the ansible playbook

kuisathaverat:
there is an extra `:`


dikshachauhan-qasource:
Hi @kuisathaverat 

Thanks for fixing the windows node creation script. We are now successfully able to create windows node vm and run the tests.

PR #3596, opened 2023-07-27
Migrate k8s autodiscovery pipeline

reakaleek:
## What does this PR do?

Migrate k8s autodiscovery pipeline to GitHub Actions workflow.

## Why is it important?

Jenkins Migration

## How to test?

Workflow run: https://github.com/elastic/e2e-testing/actions/runs/5689630212/job/15421501777?pr=3596

## Author's Checklist

- [x] Create workflow that runs k8s-autodiscover test
- [x] Remove deprecated Jenkins configs


.github/workflows/k8s-autodiscovery-test.yml
```diff
@@ -0,0 +1,42 @@
...
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ matrix.branch }}
+      - id: get-go-version
+        name: Get Go version
+        run: echo "version=$(cat .go-version)" >> "$GITHUB_OUTPUT"
+      - uses: actions/setup-go@v4
+        with:
+          go-version: ${{ steps.get-go-version.outputs.version }}
```
kuisathaverat, line 35: you can use go-version-file instead 

https://github.com/actions/setup-go/blob/main/action.yml#L7

.github/workflows/k8s-autodiscovery-test.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Create k8s Kind Cluster
+        uses: helm/kind-action@v1.8.0
```
kuisathaverat, line 37: wow, this actions is really nice

kuisathaverat:
@cachedout I think we should start removing all the stuff is not used in this repo to avoid confusion in the future. Of course in a follow up PR.

kuisathaverat (approved):

.github/workflows/k8s-autodiscovery-test.yml
reakaleek: I was considering this.

But the version defined in https://github.com/elastic/e2e-testing/blob/6a560356cdeb0befde19147cbe1bcb75d88c407b/go.mod#L3

differs from the version defined in https://github.com/elastic/e2e-testing/blob/main/.go-version

and according to the docs it only accepts paths to `go.mod` or `go.work` files. Hence, I was not able to use `.go-version` file.

PR #3834, opened 2024-05-07
Consider any HTTP response code > 399 as an error

ycombinator:
## What does this PR do?

When downloading files, consider any HTTP response codes > 399, not just 404 response codes, as errors.

## Why is it important?

To be more resilient to HTTP failures.


blakerouse (approved):
Looks good.

